PR #6, opened 2017-07-27
Fix wrong function call, caught by assert on Windows.

code-disaster:


floooh:
Wow, the OSX GL implementation totally doesn't catch that in glGetError(). On the other hand... not really surprising ;) Thanks!

PR #8, opened 2017-07-28
add missing defaults to switch statements

kevinw:
When compiling with fips and vs2015 or vs2017, warnings are errors and switch statements without a default case are warnings.

This change adds a `SOKOL_UNREACHABLE` macro which right now is just defined to `SOKOL_ASSERT(false)`.


floooh:
Thanks! I've been working on clang only so far, and this doesn't complain in those switch statements (if the switch contains all values in the enum). VS seems to have a bit different behaviour. My goal is of course to have all compilers warning free.

Issue #29, opened 2018-01-09
calling sg_reset_state_cache produces opengl error 0x502


pplux:
Every frame I'm rendering I mix sokol with other opengl related libraries, not a good Idea I know, but sokol is supposed to handle that with sg_reset_state_cache, doesn't it?

the order is:
1) render with sokol
2) render with another_lib (nanovg)
3) frame
4) call sg_reset_state_cache
--> something inside sg_reset_state_cache produces OpenGL error 0x502
5) sokol refuses to render due to openGL error

If I clear openglErrors between 4 and 5, it works... but something is not right.

(Thanks for an awesome library, btw 👍  )

PD: The proper way is to implement a nanovg backend on top of soko_gfx, I know! Just need time.

floooh:
Strange indeed... mixing with native API calls is supported of course, but is so far only tested in the 'inject' samples, so may be I overlooked something.

Is your repo public so that I could try to reproduce the problem? I'd like to find out where exactly in sg_reset_state_cache() the problem happens.

The following info would also be about your config would also be helpful: operating system, GPU model and maybe driver version (or if you're on Windows you can run "dxdiag" , save the information to a text file and attach it here, this is for D3D, but the hardware/driver info is right anyway).

PR #30, opened 2018-01-09
FIX:calling sg_reset_state_cache produces opengl error 0x502

pplux:
fixes floooh/sokol#29

restoring the state should also restore the VAO that gl backend uses.


floooh:
hah, good catch! thanks for the fix :)

PR #51, opened 2018-06-14
sokol-app: WinMain cmdline args using __argc & __argv

code-disaster:


floooh:
WHAAAT? How?!? Why didn't anybody tell me about this in the last 20 years?

floooh:
Thanks btw :)

PR #52, opened 2018-06-14
sokol-app: Win32 preprocessor

code-disaster:
minor cleanup; in my test, I'm including `glcorearb.h` (via gl3w.h) which already defines `WIN32_LEAN_AND_MEAN`


floooh:
Thanks! Win32 is still quite WIP btw, I'll try to add HighDPI support in the next few days.

floooh:
...ah and I'm also planning to add an optional minimal GL loader, only the stuff thats used by sokol_gfx.h, so that an extra GL header isn't needed on Windows. But there will be a define which switches this off.

PR #55, opened 2018-06-27
Conditional including of libs

albertofustinoni:
Since sokol already uses Direct3D11 as a back-end and it's not calling desktop only APIs, it's already UWP compliant; only one issue - UWP apps are meant to only link against `WindowsApp.lib` and nothing else.
Source [here](https://docs.microsoft.com/en-us/uwp/win32-and-com/win32-and-com-for-uwp-apps):

> ....Win32 APIs that are part of the UWP and that are implemented by all Windows 10 devices.
> .For convenience, an umbrella library named WindowsApp.lib is provided in the Microsoft Windows Software Development Kit (SDK), which provides the exports for this set of Win32 APIs.
> Link your app with WindowsApp.lib (and no other libraries) to access these APIs.

P.S. Love the idea behind your work and am trying to use it myself. Hence the small fix to make it fully compatible :)

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -4074,13 +4074,19 @@ _SOKOL_PRIVATE void _sg_update_image(_sg_image* img, const sg_image_content* dat
 #include <windows.h>
 #include <d3d11.h>
 
+#if (defined(WINAPI_FAMILY_PARTITION) && !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP))
+#pragma comment (lib, "WindowsApp.lib")
+#else
 #pragma comment (lib, "user32.lib")
 #pragma comment (lib, "dxgi.lib")
 #pragma comment (lib, "d3d11.lib")
 #pragma comment (lib, "dxguid.lib")
+#endif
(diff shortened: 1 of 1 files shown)


floooh:
Cool stuff! I'll most likely merge this on Friday (29-Jun) when I have access to both Win10 and Win7 machines. Thanks!

albertofustinoni:
Thank you so much :)

One question if I can: does Sokol support rendering to SwapChains (like the ones provided by SwapChainPanel on Windows, MTKView on Mac/iOS and GLSurfaceView on Android)?

I am interested in creating a cross platform app that renders using Sookol in a region of the UI and uses the native toolkits to create the rest of the UI.

floooh:
Re your swap chain question: I don't know, but I think it should work unless there are some 'oversight-bugs'. In sokol_app I'm using MTKView for Metal both for OSX and iOS, but haven't tried anything different than the MTKView covering the whole window client area.

albertofustinoni:
Well, I can try spinning up a project and see how it works.

Another question: is it possible to use Sokol to render to texture? Is it a supported use cas? What about viewport setting? I seem to recall it being set to the whole frame buffer in the source.

PR #56, opened 2018-06-28
Geometry shaders

albertofustinoni:
Creating a separate PR since this is a bigger change and there may be reasons not to do this I don't understand.

Since Sokol doesn't seem to use geometry shaders, and to cope with cases where the app may be using native APIs as well, modifying d3d state change and reset code to explicitly unset geometry shaders and related resources.

This prevents valid pipelines from failing because some other call left a geometry shader set in the device.


floooh:
Staaahp :) IMHO sokol-gfx should only reset state which it modified itself, and since there won't be GS support it also shouldn't clear them.

If you want to mix your own D3D code with sokol, you should do the required 'clear' operations yourself before calling sg_reset_state_cache(). Can you undo the latest commit? I'd still want to merge the UWP libs change.

albertofustinoni:
This is why Ii created a separate PR for this one. Closing it then.

PR #57, opened 2018-07-01
fixed raw mousebutton 0 on some mouse events

pplux:
Not sure if this is the proper solution, but on windows it didn't compile, I guess there are some mouse events that do not carry mouse button, so I added MOUSEBUTTON_NONE for those cases.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -238,9 +238,10 @@ typedef struct {
 } sapp_touchpoint;
 
 typedef enum {
-    SAPP_MOUSEBUTTON_LEFT = 0,
-    SAPP_MOUSEBUTTON_RIGHT = 1,
-    SAPP_MOUSEBUTTON_MIDDLE = 2,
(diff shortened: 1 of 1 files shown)


floooh:
Yes I think that's a good fix. What type of compile errors did you get (I guess type-conversion-stuff when compiling from C++ code)? I have tested with VS2017 and didn't get errors there.

I'll try to merge this tomorrow. I think I need to add SAPP_MOUSEBUTTON_NONE on the same events for other platforms as well.

Thanks!

floooh:
ah found a problem... I'll move the invalid mouse button enum value to something else than 0 (probably -1), so that the valid mouse buttons are in the range 0..N, otherwise dealing with loops over the mouse buttons, or sizing arrays will be cumbersome (for instance the fix breaks the ImGui samples). I'm working on a fix.

floooh:
ok, see f89f9d5d43a181b3b7e8e9e84c57ced69b0bb9a9 for the changes.

PR #73, opened 2018-09-26
error compilation fix for MSVC Win32 target

nesbox:
error compilation fix for MSVC Win32 target


floooh:
Hmm interesting, I didn't get that error on VS2017, but I guess it doesn't hurt either ;)

floooh:
(PS: maybe it's a 32-bit thing? I'm only testing 64-bit targets on Windows).

nesbox:
yes, with 64-bit target all is ok
thank you

PR #77, opened 2018-10-11
Only allow instancing in GLESv2 when the extensions exist

eloraiby:
This should fix compilation issues on ubuntu while linking with mesa/nvidia GLESv2.

P.S. Adding `GLES2/gle2ext.h` will default to Angle extension since it's defined first

Disclaimer:
I didn't test with angle (while extensions are on).
I didn't check with emscripten either.


floooh:
Thanks!

PR #79, opened 2018-10-17
Add SOKOL_API definition for public functions

septag:
decided to do a pull request to improve sokol APIs
Added SOKOL_API to all public functions that is overriable, instead of "extern", so the user can override public function specifiers. For example, one can ```#define SOKOL_API static``` to make functions private for wrappers/internal systems, or ```#define SOKOL_API __declspec(dllexport)``` to export them from dll ...

Also, a while back we were discussing a bug on sokol_app and d3d11 + cpp compiler, and you removed extern block from implementation .. I guess this brings linker _symbol not found_ when sokol headers are implemented inside cpp unit. So I put the new SOKOL_API before every function implementation too.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -9,7 +9,10 @@
 
     Optionally provide the following defines with your own implementations:
 
+    SOKOL_API           - public api function specifier (default: extern)
     SOKOL_ASSERT(c)     - your own assert macro (default: assert(c))
+    SOKOL_CALLOC(n, s)  - your own allocate and zero initialize array function (default: calloc(n,s))
+    SOKOL_FREE(p)       - your own free function (default: free(p))
     SOKOL_LOG(msg)      - your own logging function (default: puts(msg))
     SOKOL_UNREACHABLE() - a guard macro for unreachable code (default: assert(false))
     SOKOL_ABORT()       - called after an unrecoverable error (default: abort())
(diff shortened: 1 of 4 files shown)


floooh:
Hi, in general that's a good idea, but I have questions :)

```
 #ifndef SOKOL_API
    #ifdef __cplusplus
        #define SOKOL_API extern "C"
    #else
        #define SOKOL_API
    #endif
#endif
```
Up in the __cplusplus part, I don't quite understand why each function must be annotated with ```extern "C"```, since the entire declaration block is already inside an ```extern "C" { }```, and AFAIK annotating the functions in the implementation parts isn't necessary (early versions of the sokol headers had the ```extern "C"``` both around the declaration and implementation parts, butt his wasn't necessary).

The ```extern``` that's currently used might indeed be redundant, AFAIK this is implicitly added for function declarations, at least in C (not sure about C++).

Did you see any linker errors when using sokol headers from C++? (I should have that case covered with the Dear ImGui samples, which are implemented in C++, and the sokol implementation in a separate file).

I think I'll implement two separate macros ```SOKOL_DECL``` and ```SOKOL_IMPL```, the SOKOL_DECL would go before functions up in the declaration part, and SOKOL_IMPL would go before each public function implementation.

floooh:
Ok, I have added SOKOL_API_DECL and SOKOL_API_IMPL macros to all headers, however I kept the prefixes as they were before (default for SOKOL_API_DECL is "extern" and SOKOL_API_IMPL is empty).

Issue #84, opened 2018-10-23
sokol_gfx: update a stream buffer and draw it multiple times during the same pass


Alzathar:
First of all, thanks a lot for the development of sokol!

I successfully used sokol with the proposed samples and I am now doing complementary tests with OpenGL 3.3.

One of this test is to update a stream buffer and draw it several times during the same pass. However, it does not work because one assertion ([line 9009](https://github.com/floooh/sokol/blob/master/sokol_gfx.h#L9009)) is not valid in the function `sg_update_buffer`. If I comment this assertion, it works without problems.

Why is it not allowed to update a stream buffer multiple times during the same pass?

floooh:
I've added this restriction to have a simple and consistent way to avoid unintended blocking where the CPU needs to wait for the GPU because the GPU is currently drawing from the buffer (or image).

In the GL and Metal backend, dynamic buffers and images have multiple internal buffers which sokol rotates through to avoid blocking (in D3D11, LOCK_DISCARD is used, which does the same internally).

If you would update the same buffer multiple times per frame, it would rotate through the internal buffers all in the same frame, and the CPU would need to wait for the GPU. 

When you remove the assert, it appears to work in the GL backend, but the CPU might stall waiting for the GPU (depending on whether or not the GL driver does any magic under the hood, but that's not reliable, and definitely doesn't work on WebGL). On Metal it might also appear to work, but you may get random rendering artefacts because the CPU scribbled over buffers that are currently in use by the GPU.

Alzathar:
Thank you for these explanations. I better understand the reason why you use the global buffers `vertices` and `indices` in the sokol-samples imgui GLFW sample. 

What do you think about the possibility to remove these global buffers? In the OpenGL3 example furnished with ImGui 1.66WIP ([imgui_impl_opengl3.cpp:210](https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_opengl3.cpp#L210)), the GPU buffer is updated for each ImGui command list and then several draw commands are called based on the listed ImGui command buffers.

So instead of the proposed function [void imgui_draw_cb(ImDrawData* draw_data)](https://github.com/floooh/sokol-samples/blob/master/glfw/imgui-glfw.cc#L238), do you think it would be possible (with some modification in sokol_gfx.h) to do something like the following snippet? In this case the GPU buffer is updated for each command list.

```c++
// This sample might not compile. It is only to show the difference with the existing code
void imgui_draw_cb(ImDrawData* draw_data) {
[...]

PR #86, opened 2018-10-27
sokol-gfx: sg_append_buffer() work

floooh:
This implements the new sg_append_buffer() call in sokol_gfx.h discussed here:

https://github.com/floooh/sokol/issues/84

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -2592,6 +2592,7 @@ typedef struct {
     sg_rasterizer_state rast;
     bool polygon_offset_enabled;
     _sg_gl_cache_attr attrs[SG_MAX_VERTEX_ATTRIBUTES];
+    GLuint cur_gl_vb;
     GLuint cur_gl_ib;
     int cur_ib_offset;
     GLenum cur_primitive_type;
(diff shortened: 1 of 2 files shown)


sokol_gfx.h
```diff
@@ -146,14 +146,63 @@
         beginning a pass will reset the viewport to the size of the framebuffer used
         in the new pass,
 
-    --- to update the content of buffer and image resources, call:
+    --- to overwrite the content of buffer and image resources, call:
```
Alzathar: To keep the consistency between the API and the documentation, would you to rollback to the word `update` and put in parenthesis the world `overwrite`? 

```
--- to update (overwrite) the content of buffer and image resources, call:
```

sokol_gfx.h
```diff
@@ -146,14 +146,63 @@
...
             sg_update_buffer(sg_buffer buf, const void* ptr, int num_bytes)
             sg_update_image(sg_image img, const sg_image_content* content)
 
         buffers and images to be updated must have been created with
         SG_USAGE_DYNAMIC or SG_USAGE_STREAM
 
+        only one update per frame and resource is allowed per frame, as a
+        simple countermeasure to avoid the CPU scribbling over data the
+        GPU is currently using, or the CPU having to wait for the GPU
+
```
Alzathar: I propose to rewrite the previous paragraph as following to clarify what is updated and the rational behind it.

```
         only one update per frame is allowed for buffer and image resources. The rational is to have a
         simple countermeasure to avoid the CPU scribbling over data the
         GPU is currently using, or the CPU having to wait for the GPU
```

sokol_gfx.h
```diff
@@ -993,9 +1042,9 @@ typedef struct {
     uint32_t _start_canary;
     sg_pipeline pipeline;
     sg_buffer vertex_buffers[SG_MAX_SHADERSTAGE_BUFFERS];
-    uint32_t vertex_buffer_offsets[SG_MAX_SHADERSTAGE_BUFFERS];
+    int vertex_buffer_offsets[SG_MAX_SHADERSTAGE_BUFFERS];
```
Alzathar: @floooh Out of curiosity, why do you change the type of the offsets from `uint32_t` to `int`?

sokol_gfx.h
floooh: yes good idea

sokol_gfx.h
floooh: To make the integer types consistent across the API (e.g. the size attribute in sg_update_buffer and when creating resources are int).

Also I'm not yet sure what the return value of sg_append_buffer() should be when the buffer goes into "overflow" state (when the application tries to write more data then fits into the buffer). I *might* return a negative error value, but I'm immediately happy with that idea yet.

PR #89, opened 2018-11-07
sokol_app: fix use of SOKOL_WIN32_FORCE_MAIN

code-disaster:


floooh:
Whoops how embarrassing... thanks! I'll convert one of the D3D11 examples to SOKOL_WIN32_FORCE_MAIN so that this is covered properly.

PR #94, opened 2018-11-29
Added names to public structs so they can be forward delcared

iboB:
Currently there is no way of using these structs in headers without also including the relevant `sokol_` headers. By giving the public structs and enums names, this is now possible

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Sokol
 
-**Sokol (Сокол)**: Russian for Falcon, a smaller and more nimble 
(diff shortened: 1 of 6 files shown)


floooh:
this is a good change (I've been stumbling over the same problem before but it wasn't critical enough for me to change yet), I'll give it a test on OSX and Windows before merging, and will use this 'convention' for my other headers as well...

floooh:
Alright merged, thanks a lot for the PR! (and that other one too) :)

PR #99, opened 2018-12-03
added explicit casts so impl can be from c++

iboB:
Without these explicit casts there are compilation errors if the implementation file is a C++ unit (.cpp file)


floooh:
looks good, thanks!

PR #106, opened 2019-01-22
sokol_app.h emscripten: added explicit cast so impl can be from C++

iboB:
Without this explicit cast there is a compilation error if the implementation file is a C++ compilation unit (.cpp file)

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1818,7 +1818,7 @@ _SOKOL_PRIVATE EM_BOOL _sapp_emsc_mouse_cb(int emsc_type, const EmscriptenMouseE
                     case 0: _sapp.event.mouse_button = SAPP_MOUSEBUTTON_LEFT; break;
                     case 1: _sapp.event.mouse_button = SAPP_MOUSEBUTTON_MIDDLE; break;
                     case 2: _sapp.event.mouse_button = SAPP_MOUSEBUTTON_RIGHT; break;
-                    default: _sapp.event.mouse_button = emsc_event->button; break;
+                    default: _sapp.event.mouse_button = (sapp_mousebutton)emsc_event->button; break;
                 }
             }
             else {


iboB:
Oops... sorry :) I forgot that this it must conform to a C syntax. Force-pushed the correct fix.

floooh:
Thanks for the fix! That was an oversight on my part. I was assuming compiling from C++ is already tested because of the Dear ImGui sokol-samples, but in case of the sokol_app.h samples, the sokol implementations live in a separate .c file.

I'll add a general CI test for the Linux, OSX and Win32 backends which should catch most of those problems. Unfortunately running emscripten builds in CI is a bit trickier because of the timeout limits... I guess I need to figure out something later involving docker containers...

iboB:
Well, I guess you can shortcut it. I once implemented 5 or so emscripten functions which we used for a desktop platfor. We were using the SDL2 port for most stuff so the code between emscripten and the desktop platforms had more common parts than sokol_app. Anyway, the point was to have a faster way to check whether the emscripten target builds, without having to deal with the entire toolchain.

floooh:
That's a good idea!

floooh:
Ok, more work than expected, but everything should be clean now, and there's a new test which compiles all headers in C++ mode: 

https://github.com/floooh/sokol-samples/blob/master/tests/cpp-compile-test.cc

(there's also another one for OSX+Metal which tests against Objective-C++ instead).

The CI that runs on pushes to the sokol-samples repo only runs for Linux/OSX/Windows, but the tests are also compiled when building the sokol-samples project (so for emscripten everytime I'm building the webpage). I think that's good enough for now to catch most C++ problems.

I also found a couple of things in sokol-gfx and sokol-audio, with the new sokol-gfx dummy backend and more automatic testing hopefully such oversights will be caught much sooner in the future.

Thanks for providing the final nudge for this ;)

Issue #100, opened 2018-12-03
Missing WebGL2 features


ilar81:
According to fact that base functionality for Sokol is WebGL I found out some missing features that presented in WebGL2. 
- occlusion query
- transform feedback
- copy texture image
- copy buffer
As far as I know all these features are supported by all other Sokol's backend 3D APIs. Are these features planned for the future?

floooh:
Hi, one of the reasons is that the "base API" for sokol-gfx is "WebGL/GLES2 with common extensions", and it doesn't make much sense to require WebGL2 for the base functionality until Safari supports it IMHO.

But each of the points has different reasons (I also need to do some more research on some of the points):

- **Transform feedback**: I already had that working in Oryol for the GL/WebGL2 backend a while ago but decided to remove this again since it's hard to wrap under a common API for the different backends. The main problem was the shader side. The vertex shaders would look totally different between GL and Metal for instance, and AFAIK those differences can't be 'emulated' in shader cross compilers like SPIRVCross.
- **Occlusion query**: I probably wouldn't put this into the main API but into a separate extension header, since it's a fairly exotic feature and also not supported in Metal on all iOS devices (see https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf)
- **copy images**: makes sense as long as no format conversions or scaling is involved (which might limit its usability, a render pass which reads the source texture and renders into a target texture might be the more flexible solution)
- **copy buffers**: doesn't seem to be supported by WebGL or WebGL extensions (only WebGL2?)

amerkoleci:
@floooh any plans to add more pixel formats? maybe would be cool to have kind of feature check for formats?

amerkoleci:
I think all pixel formats, we can take look at metal pixel formats or vulkan and just check in backend if supported, for example I would remove rgb pixel format as only supported on opengl, I can submit some PR

amerkoleci:
Or maybe bgfx pixel formats for example

amerkoleci:
I think sokol can gain compute shader support as there are WebGL 2.0 compute specs (https://www.khronos.org/registry/webgl/specs/latest/2.0-compute/), I will work for D3D11 on my game framework if you want.

floooh:
[...]

PR #111, opened 2019-02-06
Remove SG_PIXELFORMAT_RGB8 pixel format

amerkoleci:
Remove SG_PIXELFORMAT_RGB8 as only supported under OpenGL


floooh:
Hmm, I'd like to keep it in for now and deal with this as part of a general cleanup of the pixel formats as you suggested in #100 (I'll open a separate issue for this now).

amerkoleci:
I would love to contribute, nvm tho :)
About pixel formats I'm working on modern gpu rendering library (https://github.com/amerkoleci/vgpu)

PR #119, opened 2019-02-20
sokol_gfx.h: fix msvc release build warnings when compiled with /W4

code-disaster:
Just some missing `_SOKOL_UNUSED()` statements.


floooh:
Thanks! Surprisingly few changes :)

code-disaster:
Yes. In between I tuned back to /W2 for my sokol implementation unit(s), but right now it builds fine with /Wall.

floooh:
Ok, on Windows I'm using /W3 for the sokol-samples. Last I tested /W4 (on other stuff) I had the impression that there's quite a lot of noise that doesn't really deserve a warning. But for library code, being /W4 clean probably makes sense.

PR #122, opened 2019-02-25
WIP: Bring back glcore33 support in sapp for macOS

gustavolsson:
Not all Macs support Metal (mine is from 2010 and is stuck on macOS High Sierra 10.13.6) so I brought back GLCORE33 support in sapp like we talked about over Twitter.

What do you think of a check for `__MAC_OS_X_VERSION_MAX_ALLOWED` (https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/cross_development/Using/using.html) to only allow compiling with OpenGL on macOS versions that do not support Metal?

I haven't looked into what all the Cocoa calls do yet (I used the old commit shas you gave me as a reference) so I will not mark this PR as ready until I have had the time to do some research. The branch works on my Mac now anyway :)

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -684,8 +684,8 @@ SOKOL_API_DECL int sapp_run(const sapp_desc* desc);
         #endif
     #else
         /* MacOS */
-        #if !defined(SOKOL_METAL)
-        #error("sokol_app.h: unknown 3D API selected for MacOS, must be SOKOL_METAL")
+        #if !defined(SOKOL_METAL) && !defined(SOKOL_GLCORE33)
+        #error("sokol_app.h: unknown 3D API selected for MacOS, must be SOKOL_METAL or SOKOL_GLCORE33")
         #endif
     #endif
 #elif defined(__EMSCRIPTEN__)
@@ -928,13 +928,20 @@ _SOKOL_PRIVATE void _sapp_frame(void) {
 
 /*== MacOS ===================================================================*/
 #if defined(TARGET_OS_IPHONE) && !TARGET_OS_IPHONE
+
+#if defined(SOKOL_METAL)
 #import <Metal/Metal.h>
 #import <MetalKit/MetalKit.h>
+#elif defined(SOKOL_GLCORE33)
+#include <Cocoa/Cocoa.h>
+#include <OpenGL/gl3.h>
+#endif
 
 @interface _sapp_macos_app_delegate : NSObject<NSApplicationDelegate>
 @end
 @interface _sapp_macos_window_delegate : NSObject<NSWindowDelegate>
 @end
+#if defined(SOKOL_METAL)
 @interface _sapp_macos_mtk_view_dlg : NSObject<MTKViewDelegate>
 @end
 @interface _sapp_macos_view : MTKView
@@ -942,13 +949,28 @@ _SOKOL_PRIVATE void _sapp_frame(void) {
     NSTrackingArea* trackingArea;
 }
 @end
+#elif defined(SOKOL_GLCORE33)
+@interface _sapp_macos_view : NSOpenGLView
+{
+    NSTrackingArea* trackingArea;
+}
+- (void)timerFired:(id)sender;
+- (void)prepareOpenGL;
+- (void)drawRect:(NSRect)bounds;
+@end
+#endif
 
 static NSWindow* _sapp_macos_window_obj;
 static _sapp_macos_window_delegate* _sapp_macos_win_dlg_obj;
 static _sapp_macos_app_delegate* _sapp_macos_app_dlg_obj;
 static _sapp_macos_view* _sapp_view_obj;
+#if defined(SOKOL_METAL)
 static _sapp_macos_mtk_view_dlg* _sapp_macos_mtk_view_dlg_obj;
 static id<MTLDevice> _sapp_mtl_device_obj;
+#elif defined(SOKOL_GLCORE33)
+static NSOpenGLPixelFormat* _sapp_macos_glpixelformat_obj;
+static NSTimer* _sapp_macos_timer_obj;
+#endif
 static uint32_t _sapp_macos_flags_changed_store;
 
 _SOKOL_PRIVATE void _sapp_macos_init_keytable(void) {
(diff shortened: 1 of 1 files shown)


floooh:
Nice, surprisingly few changes :)

I think we should allow compiling the GL backend on all macOS versions, but we should add a not in the header about the known problems. 

I've been keeping an eye on recent GLFW versions, and test those against the latest macOS betas as soon as they come out, but the vsync throttling is still not working (the initial "black screen" problem had been fixed with one of the beta updates).

Apparently SDL2 has a complete fix, but I haven't tested this yet.

floooh:
hmm, I did a quick check with SDL2 2.0.9 (which is the current stable version), and unless I did something wrong, vsync doesn't work.

This is what I did (source code below):

```sh
> brew install sdl2
...
...copy sokol_gfx.h here...
...
> clang sdl2test.c -g -o sdl2test -l SDL2 -framework OpenGL
> ./sdl2test
```

With this program:

```c
#include <OpenGL/gl3.h>
#include <SDL2/SDL.h>
#define SOKOL_IMPL
#define SOKOL_GLCORE33
#include "sokol_gfx.h"

int main() {
    SDL_Init(SDL_INIT_VIDEO);
    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 2);
    SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
    SDL_GL_SetSwapInterval(1);
    SDL_Window* window = SDL_CreateWindow("Hello SDL2",
        SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, 640, 480, SDL_WINDOW_RESIZABLE | SDL_WINDOW_OPENGL);
    SDL_GLContext* context = SDL_GL_CreateContext(window);

    sg_setup(&(sg_desc){0});
    bool running = true;
    float r = 0.0f;
    while (running) {
        SDL_Event event;
        while (SDL_PollEvent(&event)) {
            if (event.type == SDL_QUIT) {
                running = false;
            }
        }
        int w, h;
        SDL_GL_GetDrawableSize(window, &w, &h);
        sg_begin_default_pass(&(sg_pass_action){
            .colors[0] = {
                .action = SG_ACTION_CLEAR,
                .val = { r, 0.5f, 0.0f, 1.0f }
            }
        }, w, h);
        sg_end_pass();
        sg_commit();
        SDL_GL_SwapWindow(window);
        
        r += 1.0f / 60.0f;
        if (r > 1.0f) {
            r = 0.0f;
        }
    }

    sg_shutdown();
    SDL_GL_DeleteContext(context);
    SDL_DestroyWindow(window);
    SDL_Quit();
    return 0;
}
```

If vsync works, this should cycle through colors with a 1-second interval, but on my machine it just flickers madly.

floooh:
Hmm, apparently this is the SDL2 fix (maybe this isn't in stable 2.0.9 yet):

https://hg.libsdl.org/SDL/rev/73f3ca85ac0e

And here's a thread about GLFW which says a fix should land soon-ish:

https://github.com/glfw/glfw/pull/1417

But also see this comment:

https://github.com/glfw/glfw/pull/1417#issuecomment-465849303

Doesn't look as simple as I was hoping :)

gustavolsson:
I don't really think it's that important to support GL on Metal capable systems, is there a particular reason you want to? Not forcing users of sokol to write multiple shaders?

Even if someone finds a workaround, I suspect that Apple may "accidentally" break it in the future. I think it'll be difficult to go against that. Of course I'm not against doing it, it's just that I don't want to spend time on something I think will break in the future :)

floooh:
We shouldn't worry about the vsync problem yet and instead merge your PR without any Mojave fixes, just let me know when its ready :)

I just wanted to check what the current state is in GLFW and SDL2, to get an idea how much work it would be to fix the GL backend for Mojave. I can look into this later, after your PR is merged.

I still wouldn't disable the GL backend on Metal-capable Macs. The current GL backend (without Mojave fixes) would still work fine for people which have a Metal-capable GPU in their Mac, but haven't upgraded to Mojave.

sokol_app.h
```diff
@@ -1086,9 +1108,15 @@ int main(int argc, char* argv[]) {
...
-    _sapp.framebuffer_width = fb_size.width;
-    _sapp.framebuffer_height = fb_size.height;
+    #if defined(SOKOL_METAL)
+        const CGSize fb_size = [_sapp_view_obj drawableSize];
+        _sapp.framebuffer_width = fb_size.width;
+        _sapp.framebuffer_height = fb_size.height;
+    #elif defined(SOKOL_GLCORE33)
+        const NSRect fb_rect = [_sapp_view_obj convertRectToBacking:[_sapp_view_obj frame]];
+        _sapp.framebuffer_width = fb_rect.size.width;
+        _sapp.framebuffer_height = fb_rect.size.height;
```
gustavolsson: According to [this document](https://developer.apple.com/library/archive/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/CapturingScreenContents/CapturingScreenContents.html), `[_sapp_view_obj bounds]` denotes the number of pixels in the framebuffer. Should frame and bounds be switched for this logic? I don't have a Mac with a retina display so I can't test this.

sokol_app.h
```diff
@@ -928,27 +928,49 @@ _SOKOL_PRIVATE void _sapp_frame(void) {
...
 }
 @end
+#elif defined(SOKOL_GLCORE33)
+@interface _sapp_macos_view : NSOpenGLView
+{
+    NSTrackingArea* trackingArea;
+}
+- (void)timerFired:(id)sender;
+- (void)prepareOpenGL;
+- (void)drawRect:(NSRect)bounds;
```
gustavolsson: It seems like these function definitions are not needed here, it compiles without them. I don't know Obj-C but it feels better to be explicit here and keep them. Or is it related to what functions are public or private?

gustavolsson:
I added a few code comments in the PR. Once they are resolved, I'm ready to merge it :)

> I still wouldn't disable the GL backend on Metal-capable Macs. The current GL backend (without Mojave fixes) would still work fine for people which have a Metal-capable GPU in their Mac, but haven't upgraded to Mojave.

I think the `__MAC_OS_X_VERSION_MAX_ALLOWED` define relates to the macOS version only, so if we throw an error for GL post Mojave, people with Metal-capable macs on macOS pre-Mojave will still be able to target GL. But let's not do it for simplicity then :)

sokol_app.h
floooh: These are part of the interface definition for _sapp_macos_view, with the implementation further down below. I agree, I think it's better to keep them there, when reading the code top-to-bottom it's clearer which methods from the NSOpenGLView parent class _sapp_macos_view overrides.

sokol_app.h
floooh: I'll do some testing and report back, it may be that the GL backend didn't have correct handling for HighDPI, and that some code confuses window size and framebuffer size.

floooh:
Aha, interesting, I'm getting a deprecated warning for NSOpenGLCPSwapInterval (in prepareOpenGL):

```objc
    [ctx setValues:&swapInt forParameter:NSOpenGLCPSwapInterval];
```
...at least this explains why vsync is broken on Mojave (don't worry about this, it's part of the things I should look at after merging the PR)

sokol_app.h
gustavolsson: Ok!

sokol_app.h
floooh: Hmm, both the ```frame``` and ```bounds``` method return the same (wrong) values, I remember that this is another thing that broke under Mojave, somehow the framebuffer is always initialized at Retina resolution even when this is not requested.

I'd say: keep it as it is for now, that's something I'll need to care about when I'm looking into the vsync problem.

gustavolsson: Ok, let's merge it :)

gustavolsson:
Ok, let's merge it :)

Let's see if someone on pre-Mojave with a Retina display can test the bounds/frame logic (though a bit unlikely :p)

floooh:
Ok, merged. I'll write myself a reminder ticket for "polishing" this a bit (add note about Mojave incompatibility, and fix the remaining problems.

PR #123, opened 2019-02-27
Win DPI scaling properly taken into account on Windows

fabiopolimeni:
Hi,

I was playing around with Sokol app API and while doing some tests I found out, at least on windows, that dpi scaling gets floored because of integer casting operations. In my specific case, the monitor has a scaling of 125%, that is, 1.25, although, because of both `_sapp_win32_content_scale` and `_sapp_win32_window_scale` are declared `int` they will be floored to 1.  I also had to change `_sapp_win32_update_dimensions()` because the first time it is called windows dimensions wouldn't be different, and therefore the framebuffer wouldn't update its dimensions. In a monitor with dpi scaling of 125%, `sapp_dpi_scale()`  would return 1f, while a value of 1.25f is expected.

I don't know whether this was intentional or not. Please have a look at this pull request, it works in my test case.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2373,8 +2373,8 @@ static HWND _sapp_win32_hwnd;
 static HDC _sapp_win32_dc;
 static bool _sapp_win32_in_create_window;
 static bool _sapp_win32_dpi_aware;
-static int _sapp_win32_content_scale;
-static int _sapp_win32_window_scale;
+static float _sapp_win32_content_scale;
+static float _sapp_win32_window_scale;
 static float _sapp_win32_mouse_scale;
 static bool _sapp_win32_iconified;
 typedef BOOL(WINAPI * SETPROCESSDPIAWARE_T)(void);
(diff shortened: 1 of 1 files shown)


floooh:
Hi, if I remember right, the clamping was intentional on Windows because otherwise I would get really bad scaling artifacts (basically some pixels scaled, some not). This was especially bad in UI applications like the Dear ImGui samples.

Could you share some screenshots here what the ImGui samples look like on your machine?

I'll give your PR a try on my Windows laptop. Maybe the best solution would be to make the clamping optional via sapp_desc.

fabiopolimeni:
I've built imgui-highdpi-sapp and I don't see any artifact. 
[IMGUI-HDPI](https://ibb.co/C68FVng)

Built with the following
./fips set config sapp-win64-vstudio-debug
./fips make imgui-highdpi-sapp
./fips run imgui-highdpi-sapp

floooh:
Hm ok, looks good indeed, I'll give it a try on my Windows laptop in the afternoon :)

floooh:
Ok, I did a quick check on my HighDPI Win10 laptop and your fix looks fine, also with odd scaling factors like 125%.

I'll merge the fix in the next few days (want to see how the imgui-highdpi demo behaves on a non-highdpi Win7 machine, which I don't have at home).

Thanks for the PR!

floooh:
And merged! Looks good on an old Win7 machine too.

PR #125, opened 2019-03-01
Debugging UI extension headers via Dear ImGui

floooh:
This PR adds optional debugging UI extension headers (or rather: currently only one for sokol_gfx.h called sokol_gfx_imgui.h).

sokol_gfx.h gets a new API function to install callback functions for tracing the public sokol_gfx.h API calls.

TODO:

- [x] in sokol_gfx_imgui.h: add the error callback functions to the frame capture
- [x] write proper usage documentation in sokol_gfx_imgui.h
- [x] add a README.md to the imgui subdirectory, which generally explains how to use the debug UI headers
- [x] sokol_gfx.h: disable trace hook code by default, enable via config define, and wrap the hook calls into a macro.
- [x] sokol_gfx.h: add documentation about trace hooks
- [x] test if variadic macros work on VS2015
- [x] update the "Updates" section in the main README

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,6 +407,20 @@ Mainly some "missing features" for desktop apps:
 
 # Updates
 
+- **05-Mar-2019**: sokol_gfx.h now has a 'trace hook' API, and I have started
+implementing optional debug-inspection-UI headers on top of Dear ImGui:
+    - sokol_gfx.h has a new function sg_install_trace_hook(), this allows
+      you to install a callback function for each public sokol_gfx.h function
+      (and a couple or error callbacks). For more details, search for "TRACE HOOKS"
+      in sokol_gfx.h
+    - I have started a new 'subproject' in the 'imgui' directory, this will
+      contain a slowly growing set of optional debug-inspection-UI headers
+      which allow to peek under the hood of the Sokol headers. The UIs are
+      implemented with [Dear ImGui](https://github.com/ocornut/imgui). Again,
+      see the README in the 'imgui' directory and the headers in there 
+      for details, and check out the live demos on the [Sokol Sample Webpage](https://floooh.github.io/sokol-html5/)
+      (click on the little UI buttons in the top right corner of each thumbnail)
+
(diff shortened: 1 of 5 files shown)


Alzathar:
@floooh Thanks for this work! I was reading your code modifications when I though about its effect when no trace hooking is installed. There is a total of 72 `if (_sg.hooks.*)` conditions with some of them that can be repeated during a pass. The cost of these conditions are negligible when no trace hooking is installed, but I was wondering if we can have a way to discard them completely. You may want to use the tracing tool in development mode, but discard any trace of it in production mode.

What about the possibility to include the trace code only when a macro is defined? For example 
```c++
#if defined(SG_HOOK_ENABLED)
if (_sg.hooks.query_feature) {
         _sg.hooks.query_feature(f, res, _sg.hooks.user_data);
     }
#endif
```

The code above may be simplified by another macro.

```c++
SG_HOOK(query_feature, f, res);
```

The definition of `SG_HOOK` could be like below.

```c++
// NOT TESTED
#if defined(SG_HOOK_ENABLED)
#define SG_HOOK(fn, ...) \
  if (_sg.hooks.fn) { \
    _sg.hooks.fn(__VA_ARGS__, _sg.hooks.user_data); \
  }
#else
#define SG_HOOK(fn, ...)
#endif
```

floooh:
Hmm, yeah, but I'm tending more towards a single "all or nothing" macro. Not sure yet whether default would be hooks enabled or disabled. I'll sleep over it ;)

floooh:
Ok, I made up my mind :)

First, it seems all compilers support variadic macros, at least according to godbolt, which only seems to go have VS2017, I'll check on VS2015 too, but it looks like it's ok to use them in headers.

So basically, how you suggested, trace hooks are disabled by default and must be enabled (the public API declaration will remain the same though, but the implementation code will be removed), and I'll wrap the call to the hooks in a variadic macro.

Thanks for the suggestion :)

floooh:
Ok, trace hook code is optional now, off by default, and the callback invocations are in a macro (see 23aadb04853add02ecda47f88f7f2fb1e15a1249)

Issue #112, opened 2019-02-06
Extend and cleanup supported pixel formats.


floooh:
The list of pixel formats should be cleaned up so that it it contains a "core" of pixel formats that are supported in all backends, and a more relaxed list of optional backend-specific pixel formats for which support must be queried (similar to how the compressed texture formats are currently handled).

There's also the problem of renderable pixel formats vs pixel formats that are only supported for texturing, how should this be queried? Should we keep the simple sg_query_feature(...) and just extend the feature enum with new items? Or should there be a new query-feature method just for pixel formats, with additional query params (such as "renderable")?

amerkoleci:
That makes sense, I think we should allow to query feature for every single format (core also) and have something to vulkan (vkGetPhysicalDeviceFormatProperties) and therefore know max size (width, height) and usage allowed, if sampled, render target, depth stencil etc

floooh:
Hmm yes, probably a new ```sg_pixel_format_info sg_query_pixel_format_info(sg_pixel_format)``` function, this would return a struct with the most important info (what's in this struct should again be the common set of information available in the backend APIs. In OpenGL, at least some of this stuff can be queried via glGetIntegerv() (e.g. GL_MAX_TEXTURE_SIZE), others (like whether a pixel format is renderable) is (AFAIK) defined by the GL spec for a specific GL version.

mattiasljungstrom:
Any plans to support linear output to screen framebuffers? (The naming conventions are a bit confusing, basically support for GL_FRAMEBUFFER_SRGB / EXT_framebuffer_sRGB)

floooh:
>  Any plans to support linear output to screen framebuffers? (The naming conventions are a bit confusing, basically support for GL_FRAMEBUFFER_SRGB / EXT_framebuffer_sRGB)

I'm not sure if that's worth the trouble to be honest, is there any advantage versus doing the srgb->linear->srgb conversion in the pixel shader (I think the performance difference is negligable, even on mobile GPUs)?

mattiasljungstrom:
Makes sense, I haven't actually tested the performance difference.

floooh:
...work is happening here: https://github.com/floooh/sokol/tree/pixelformat

floooh:
Closing this as duplicate for new issue: #187

PR #136, opened 2019-03-15
Integer texture pixel formats

iboB:
This is not a complete feature. It's only implemented for GL and tested on Linux and WebGL 2.0.

Still it can be a start to something quite useful. Integer textures (and integer samplers) can be of great help in certain situations like clever sampling optimizations by packing multiple data within a single texture.

So perhaps this can be extended to add the complete feature support to sokol?...

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -330,9 +330,9 @@
     - define SOKOL_NO_ENTRY before including the sokol_app.h implementation
     - do *not* provide a sokol_main() function
     - instead provide the standard main() function of the platform
-    - from the main function, call the function ```sapp_run()``` which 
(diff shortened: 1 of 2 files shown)


floooh:
Hi, @iboB, this is a good and useful PR, so apologies for not merging it :)

There's a (somewhat longstanding ticket) to improve pixel format support in sokol-gfx, along with support for querying whether a pixel format is supported, and I think it's better to do this all in one swoop:

https://github.com/floooh/sokol/issues/112

My current thinking is to take the texture format list from BGFX as a base:

https://bkaradzic.github.io/bgfx/bgfx.html#_CPPv2N4bgfx13TextureFormatE

PR #147, opened 2019-05-01
sokol_gfx_imgui.h: fix use of deprecated ImGui enums

code-disaster:
... when compiled with IMGUI_DISABLE_OBSOLETE_FUNCTIONS


floooh:
Oh whoopsi, I should probably start to use this IMGUI_DISABLE_OBSOLETE_FUNCTIONS in all my code, thanks!

PR #149, opened 2019-05-09
added cimgui wrapper and gfx debugger

prime31:
I'm not sure if you're interested in something like this, but I figured I'd send over a PR and you can either accept it or close it. I converted the sokol Dear ImGui renderer and gfx debugger over to work with cimgui for projects that don't use C++.

### DIFF
--- a/util/sokol_cimgui.h
+++ b/util/sokol_cimgui.h
@@ -1,0 +1,985 @@
+#pragma once
+/*
+    sokol_cimgui.h -- drop-in Dear ImGui renderer/event-handler for sokol_gfx.h
+
+    Do this:
+        #define SOKOL_CIMGUI_IMPL
+    before you include this file in *one* C file to create the
+    implementation.
+
+    The following defines are used by the implementation to select the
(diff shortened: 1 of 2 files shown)


floooh:
Hi @prime31, yes, this is extremely cool! I wasn't sure how closely cimgui follows the latest imgui versions, AFAIK in the past it was always quite a bit behind. But it looks it is now automatically generated, so everything good :)

I'll play around a bit with your PR and will also add a cimgui sample to the sokol-samples repo (and at least one using the cimgui gfx debugger), so it'll be a few days until I merge the PR.

floooh:
...cimgui sample added: https://floooh.github.io/sokol-html5/wasm/cimgui-sapp.html

floooh:
...and the microui-sample now uses sokol_gfx_cimgui.h for its debugging-UI:

https://floooh.github.io/sokol-html5/wasm/sgl-microui-sapp-ui.html

prime31:
Woohoo! Now we get to stay in C mode with its beautiful compound literal syntax for structs everywhere ;)

PR #152, opened 2019-05-12
sokol_gfx.h: Cleanup resource 'unbinding' behaviour...

floooh:
This implements the following new "clarifications" in the resource binding behaviour:

- in the GL backend: creating or updating a resource restores the previous resource binding (this may have behaved wrong before when a resource was created or updated between a call to sg_apply_bindings() / sg_update_...() and sg_draw()

- in sg_commit(), make sure to unbind all resources that were bound in the frame (just to prevent any accidental resource leaks)

- sg_apply_bindings() will still NOT unbind any "unused" resource slots, this is too much hassle in the GL backend. Instead this happens at the end of the frame in sg_commit().

- minor GL backend fix: any attempts to bind a texture to a too high slot number (GL_MAX_COMBINED_IMAGE_UNITS) will be silently dropped (before this would have produced a GL error... however this would only be an issue for very old GLES2 implementations which are only guaranteed to have 8 texture slots).

"Correct code" shouldn't require any changes.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2867,6 +2867,7 @@ typedef int  GLint;
 #define GL_POINTS 0x0000
 #define GL_ONE_MINUS_SRC_COLOR 0x0301
 #define GL_MIRRORED_REPEAT 0x8370
+#define GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS 0x8B4D
 
 typedef void  (GL_APIENTRY *PFN_glBindVertexArray)(GLuint array);
 static PFN_glBindVertexArray _sapp_glBindVertexArray;
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -1311,7 +1311,7 @@ typedef struct sg_image_desc {
     sg_shader_desc
 
     The structure sg_shader_desc describes the shaders, uniform blocks
-    texture images and vertex-attribute-names/semantics (required 
+    texture images and vertex-attribute-names/semantics (required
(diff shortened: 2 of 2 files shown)


radgeRayden:
Unfortunately when trying to compile with `SOKOL_GLCORE33` it fails by not finding `GL_MAX_COMBINED_IMAGE_UNITS`, so I couldn't test.

floooh:
...ah alright, I haven't added that define yet to the "mini GL loader" in sokol_app.h for the Windows platform. On macOS where I'm currently testing, the system GL headers are used.

floooh:
Should work now on GL on Windows, needs an update on sokol_app.h

radgeRayden:
Tested on windows with [this](https://gist.github.com/radgeRayden/e3483b00ae4d0267328b5770de17bd5e) (slightly) modified dyntex sample code, black texture on GLCORE33 and assertion fail on D3D11.

floooh:
Ok, kinda expected, since with the modified code sokol_gfx.h has no knowledge that the fragment shader has a texture sampler. I guess the assertion in D3D11 is coming from D3D's validation layer itself.

With the new shader-code-generation stuff, the shader compiler will provide reflection information to the C code, so that such a situation should become a proper validation error in sokol_gfx.h.

floooh:
...note to self: the Metal backend already conforms to the "behaviour clarification" outlined above.

floooh:
...and D3D11 also conforms to the "clarification", ready for merge.

PR #153, opened 2019-05-16
Compute feature check support

amerkoleci:
This add initial feature for compute check in order to support compute as WebGL 2.0 is gaining support as well (https://github.com/9ballsyndrome/WebGL_Compute_shader).
After will add new flag to create texture with different usage (shader read, shader write, render target).

Tell me what you think

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -2514,6 +2514,7 @@ typedef struct {
     bool valid;
     ID3D11Device* dev;
     ID3D11DeviceContext* ctx;
+    D3D_FEATURE_LEVEL feature_level;
     const void* (*rtv_cb)(void);
     const void* (*dsv_cb)(void);
     bool in_pass;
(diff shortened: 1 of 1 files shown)


floooh:
Hi, I actually need to read up on computer shader support in D3D11 and Metal, and especially the extension for WebGL2, your changes so far look good.

I'm not sure yet whether it's better to have the entire compute support code in the sokol_gfx.h header itself, or in an extension header included after sokol_gfx.h (but in any way, basic things like checking and querying for compute-support need to go into sokol_gfx.h anyway).

amerkoleci:
Yes and texture and buffer need the flag to allow uav creation for d3d11 and Shader Write on metal

amerkoleci:
Opengl need just new GL_STORAGE_BUFFER type for buffer

raizam:
Will this ever get merged ?

floooh:
Not sure if it's ready. I'm also not sure how useful that feature is without sokol-gfx actually supporting compute (even if actually running a compute shader is implemented as an extension function, the resource creation functions would need to be extended as well).

PR #154, opened 2019-05-17
Pixel format info

amerkoleci:
This initially implements #112 .
Tell me if you like the API so I can add to other backends.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -1572,6 +1572,9 @@ typedef struct sg_pass_desc {
 
 typedef struct sg_pixel_format_info {
     bool supported;
+    uint32_t max_width;
+    uint32_t max_height;
+    uint32_t max_depth;
 } sg_pixel_format_info;
 
 /*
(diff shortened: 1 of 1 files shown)


floooh:
Yes this looks good!

Two comments:

- Calls which require non-GLES2 features like this: https://github.com/floooh/sokol/pull/154/files#diff-a8b646082914484fa2639b5ad5947b55R4066 need to behind a compile- *and* runtime-check:

```c
#if !defined(SOKOL_GLES2)
if (!_sg.gl.gles2) {
   ...
}
#endif 
```

- I think there should be another query-info-member ```max_layers``` for the max number of array-texture layers, queried from GL with GL_MAX_ARRAY_TEXTURE_LAYERS. This must also be behind a "not GLES2" check like above.

amerkoleci:
Can you take look if it's ok?

floooh:
Looking over the changes it looks good to me (I haven't merged or run the code).

floooh:
Hi @amerkoleci, I have merged this PR into the branch: https://github.com/floooh/sokol/tree/pixelformat, and will continue working there on the whole "pixel format cleanup topic". Your commit history is intact so I guess that's ok ;)

PR #156, opened 2019-05-19
Fix crash when glGetString returns NULL on GL_EXTENSIONS

eloraiby:


eloraiby:
Happens on old AMD drivers for linux

floooh:
Merged, thanks!

Issue #162, opened 2019-06-04
Ignore key repeat.


slembcke:
It seems that there is no way to ignore key repeat events? I get repeated key down events on Linux. I can test on Mac later, but that computer is currently occupied. (I don't see a check for NSEvent.isARepeat though)

It would be pretty easy to add as a boolean to the sapp_event struct. NSEvent makes it pretty easy on Apple, and I could look into Linux.

floooh:
Sounds like a good feature. New ```bool key_repeat``` in sapp_desc?

If you want to, you can provide a partial pull request, and I can fill in the rest :)

slembcke:
Yup, that was exactly my thought. I'm hoping to have time this week to look into it.

PR #164, opened 2019-06-04
Implement key repeat flag.

slembcke:
https://github.com/floooh/sokol/issues/162

Implemented for Mac. Will look into Linux next and update the pull request as I go.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3828,20 +3828,22 @@ _SOKOL_PRIVATE void _sapp_win32_scroll_event(float x, float y) {
     }
 }
 
-_SOKOL_PRIVATE void _sapp_win32_key_event(sapp_event_type type, int vk) {
+_SOKOL_PRIVATE void _sapp_win32_key_event(sapp_event_type type, int vk, bool repeat) {
     if (_sapp_events_enabled() && (vk < SAPP_MAX_KEYCODES)) {
         _sapp_init_event(type);
         _sapp.event.modifiers = _sapp_win32_mods();
         _sapp.event.key_code = _sapp.keycodes[vk];
+        _sapp.event.key_repeat = repeat;
         _sapp_call_event(&_sapp.event);
     }
 }
 
-_SOKOL_PRIVATE void _sapp_win32_char_event(uint32_t c) {
+_SOKOL_PRIVATE void _sapp_win32_char_event(uint32_t c, bool repeat) {
     if (_sapp_events_enabled() && (c >= 32)) {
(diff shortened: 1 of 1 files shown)


slembcke:
A reasonable solution I found for Linux was to use [XkbSetDetectableAutoRepeat()](https://linux.die.net/man/3/xkbsetdetectableautorepeat) to disable the virtual key up events, then keep shadow copy of the key state to check if multiple key down events happen in a row.

According to the XLib manual, the XEvent keycodes are always less than 255, so I just used a small static array. https://tronche.com/gui/x/xlib/input/keyboard-encoding.html The keycode type being int made me paranoid so I mask it with 0xFF just to be thorough.

I might have time to look into Windows too, but I know nothing about Emscripten. Assuming that iOS/Android don't need to worry about repeat events anyway.

floooh:
Looks like in Emscripten there's a field in the keyboard event struct:

https://emscripten.org/docs/api_reference/html5.h.html#c.EmscriptenKeyboardEvent.repeat

Looks simple enough, just let me know when I should take over ;)

slembcke:
Ok. Windows was easy enough to implement. Just needs more magic numbers for masks. (Getting VS reinstalled on my Windows machine again took several hours though >.<)

Anyway, I'll let you take care of the emscripten bits. Sounds like it's straightforward, but I don't have a setup for testing it.

Thanks for your time!

floooh:
I guess this is ready for merging, and me doing the emscripten stuff right?

I think I'll also add a better sample with Dear ImGui UI for testing input events (or maybe a full-blown debugging UI for sokol_app.h like sokol_gfx_imgui.h)

floooh:
...ok I went ahead and merged it, working on the emscripten part now.

Thanks a lot for the suggestion and very clean PR! :)

floooh:
@slembcke FYI when pressing one of the control keys on Mac (Alt, Ctrl, etc...) I'm getting a "non-fatal assert" in Xcode, apparently this event type doesn't support the isARepeat method:

```
Assertion failure in -[NSEvent isARepeat],
...
Invalid message sent to event "NSEvent: type=FlagsChanged loc=(148.277,-15.7578) time=32738.2 flags=0x80140 win=0x1010110b0 winNum=2327 ctxt=0x0 keyCode=61"
...
```

Happens here:

https://github.com/floooh/sokol/blob/b4d9bb9e538739b5f9f58ecce5c41ed0433eea36/sokol_app.h#L1491

I'll simply pass a ```false``` there.

floooh:
...ok all finished, new event-inspector sample: https://floooh.github.io/sokol-html5/wasm/events-sapp.html

PR #173, opened 2019-06-24
Linux wayland sapp

fabiopolimeni:
Are you interested in a Wayland-EGL implementation for Linux?

This initial attempt seems to work for me with all the sokol samples and cpp tests.

The high dpi though looks like it makes ImGui render out of the clip space. I am not sure the high dpi implementation is correct, the GUI appears but it goes out of screen as if the framebuffer is not "shrunk" within the window's size.

I have tested this in Sway and Weston compositors.

It supports only GLESv2/v3 at the moment, I plan to include GL, as it should fairly similar to your X11 implementation.

Let me know if you are interested and if so we can iterate a little bit more to make it more reliable.


floooh:
Yes I think Wayland support directly in sokol_app.h is a good thing to have, and it looks like it isn't too much additional code to maintain.

As for the ImGui HighDPI problem, I fixed a problem there quite recently, previously this had the effect that ImGui's "canvas" extended 2x to the right and bottom beyond the screen, not sure if this could also have other effects in different situations.

https://github.com/floooh/sokol/commit/fe159f4a37821cb66f8dfed1da51e31800412aff#diff-7686f41e5451a87e22bd7ad87ce5f024

It may also be different behaviour for what size sapp_width/height and sapp_dpi_scale() is returning, sapp_width/height should return the framebuffer dimensions, not the window client area dimensions.

fabiopolimeni:
I did look at that code change and it is in the repository I have used. I think the problem is more from Wayland compositor. I will need to investigate a little bit more in depth, the worst case scenario highdpi would not be supported at the beginning.

Some of the sapp events don't make sense for Wayland, such as window's minimize, because it is up to the compositor how to implement it based on its functionalities. For examples my two compositors of reference don't have it.

There is one more thing though. Wayland needs the implementation of the protocol, and the standard way is to call `wayland-scanner` to generate a shell .h and .c files from an .xml. 

An example of this is in this makefile https://github.com/eyelash/tutorials/blob/master/wayland-compositor/makefile

Now, protocols usually don't get updates too often, I think the latest minor version is from 2018. (https://github.com/wayland-project/wayland-protocols) We need to decide how to implement this.

I figure options are:
1. The user has to provide those files
2. I can upload those files from my current protocols version
3. I can simply copy-paste those few functions into sokol-app.h (similar to what you did with the Android version)
4. Require a pre-step, similar to what you did in for sokol shaders, and make it copy those files into their final position (while at point 1 the user would have do it herself) needed at build time. With this last option I would need quite some help from you.

Any thoughts?

floooh:
Could you try to find out how the GLFW Wayland support solves the wayland-scanner problem?

If their solution is okay, we should just do it the same way I think?

I'm surprised there's a C file though, shouldn't this just be some sort of API header with typedefs and function signatures?

I see 2 options:

1. embed all Wayland types and function signatures (and any code from that mysterious .c file) right in sokol_app.h (only the stuff that's needed, not everything)...similar to how I did the adhoc-GL-loader for the Windows platform

2. require the user to include this generated wayland file before including the sokol_app.h implementation

fabiopolimeni:
This is the version with the implementation of the protocol directly inside the file. I have added only the needed functions. The .c file is just used to declare some `extern` references, these are not mandatory as far as I can tell, although that file was very short and it comes in handy because I don't need to use literals in the code, I can resolve a reference to it instead.

As far as I understand from GLFW, they actually create the definition of the protocol files with WaylandScanner.

https://github.com/glfw/glfw/blob/a4d910b4a72f3a85d7be42ea2f91162cdfd28375/src/CMakeLists.txt#L34

https://github.com/glfw/glfw/blob/a4d910b4a72f3a85d7be42ea2f91162cdfd28375/CMake/modules/FindWaylandProtocols.cmake#L25

It seems that `[ecm_add_wayland_client_protocol()](https://api.kde.org/ecm/find-module/FindWaylandScanner.html)` is part of a KDE extensions module. I have no idea whether this modules are there in all Linux distros or CMake for Linux builds, all I know is I can build GLFW for Wayland.

Here you can see where they include those files from in the source code.
https://github.com/glfw/glfw/blob/a4d910b4a72f3a85d7be42ea2f91162cdfd28375/src/wl_platform.h#L59

pixelherodev:
The KDE extensions module is in all standard distros as far as I know - I have it in Debian and Alpine and I'm pretty sure it's in most other distros as well.

Under Debian it's called extra-cmake-modules IIRC.

Regarding GL, from what I remember from when I patched Oryol to expose GLFW's wayland support, it's no different from with X11, except using EGL instead of GLX.

If you need someone to test something out or anything like that let me know.

pixelherodev:
What's the status on this? @fabiopolimeni 

If you're no longer interested in working on this, I'd be interested in taking it over. I exclusively use Wayland, and I've had some minor annoyances (read: compositor segfaults) as a result of XWayland issues.

fabiopolimeni:
> What's the status on this? @fabiopolimeni
> 
> If you're no longer interested in working on this, I'd be interested in taking it over. I exclusively use Wayland, and I've had some minor annoyances (read: compositor segfaults) as a result of XWayland issues.

I would be more than happy for you to take this over. The reason I've never pushed it furtehr was because I wanted to fix the DPI problem not taking properly into account the cursor when scaling. Since then I didn't have a chance to run on Wayland again. So, please give it all the love you want to this.

fleischie:
@pixelherodev I am also interested in wayland support.

But I checked the current PR is more than outdated to how `sokol_app.h` is structured. I would naively guess the best bet is to start over with the implementation (especially splitting the currently default linux declarations and includes from the X11-related includes).

I might put some work on it, but honestly I am still currently learning how to use Wayland itself, let alone implement it into sokol. :D

PR #178, opened 2019-07-02
new sokol_fetch.h

floooh:
This PR adds the sokol_fetch.h header for asynchronous data loading/streaming.


ArnCarveris:
@floooh `fokol_flush.h` for writing will be planned?

RandyGaul:
Should add some code samples in the docs to demonstrate a full workable example!

RandyGaul:
Could you please add in a copy-pasteable example to the docs so we can easily see what a use-case looks like?

floooh:
> Could you please add in a copy-pasteable example to the docs so we can easily see what a use-case looks like?

Yes, I'll add simplest-possible example at the top of the documentation block, I also just finished a doc segment where I'm describing different loading/streaming scenarios, which each come with some sample code (mostly what the sfetch_send() call and the response-callback function looks like).

In the meantime (search for ```sfetch_```):

- loading a PNG file into a sokol_gfx.h texture: https://github.com/floooh/sokol-samples/blob/sokol-io/sapp/loadpng-sapp.c

- streaming an MPEG1 video: https://github.com/floooh/sokol-samples/blob/sokol-io/sapp/plmpeg-sapp.c

The samples will also soon be up on the sample page (here: https://floooh.github.io/sokol-html5/)

floooh:
> @floooh fokol_flush.h for writing will be planned?

Not right now. The next IO-related header will probably be a sokol_fsutil.h header with various filesystem utility functions (getting the directory where the executable is, along with some other common "special locations"), some helper functions for building path strings, and some cross-platform filesystem functions which are not covered very well by the POSIX functions (check if file/directory exists, make a directory or a whole cascade of subdirectories, etc...).

At least the "give me the directory where the executable resides" is important for iOS and Android, with that, sokol_fetch.h is currently quite useless.

ArnCarveris:
@floooh Hmm, reminds me to `VFS` pattern, mounting path to folder, BigFile(`.fat` & `.dat` like since `FarCry 2`), ZipFile(`.pak` like in `FarCry`) or _any-thing-else_. Would be great if this will be possible in future.

PR #188, opened 2019-07-26
Add sapp_android_get_native_activity function

kochol:
I need it to access AAssetManager* in my project. Maybe others need it in their projects too.


floooh:
small typo in the comment, but I'll fix that in post ;)

floooh:
ah, and many thanks btw :)

PR #189, opened 2019-07-28
Add for checking the ETC2 texture support.

kochol:
OpenGLES 3 always support ETC2.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -4092,11 +4092,13 @@ _SOKOL_PRIVATE void _sg_setup_backend(const sg_desc* desc) {
                 _sg.gl.features[SG_FEATURE_MSAA_RENDER_TARGETS] = true;
                 _sg.gl.features[SG_FEATURE_PACKED_VERTEX_FORMAT_10_2] = true;
                 _sg.gl.features[SG_FEATURE_MULTIPLE_RENDER_TARGET] = true;
+                _sg.gl.features[SG_FEATURE_TEXTURE_COMPRESSION_ETC2] = true;
             }
             else {
                 _sg.gl.features[SG_FEATURE_INSTANCING] = strstr(ext, "_instanced_arrays");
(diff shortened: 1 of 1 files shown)


floooh:
Hi, thanks for the PR, however, I'm currently reworking the whole area of capability checks and pixel formats in this branch:

https://github.com/floooh/sokol/tree/pixelformat

So the PR would be obsolete.

Some sort of ETC2 support check in the GLES3 branch is still needed for WebGL2 though, even though ETC2 is guaranteed in GLES3, it isn't in WebGL2.

In the new branch there's a check whether ETC2 support is guaranteed, or must be checked through an extension, the code starts here: 

https://github.com/floooh/sokol/blob/de71620b3e560fc0bcbd818bdc3da8b7ea7e9b1f/sokol_gfx.h#L4440

I'm closing this PR, and hopefully I'll be able to merge the pixelformat branch back to master some time this week.

Cheers!

kochol:
Hello happy to hear that you are working on it.

I have some questions about your work.
1. Why you don't support ETC1?
2. Why you don't have SG_PIXELFORMAT_ETC2_RGBA8?
3. I did a little research and extention for ETC2 contains "_compressed_ETC2_" string but you check for "_compressed_texture_etc" in your new branch. Is it a bug?

floooh:
Any missing ETC formats are basically because I didn't have much of a chance to test all the different variations yet. The two ETC2 formats are basically what I supported in Oryol and which could be generated by the texture compression tool I used at the time.

Nowadays it's probably better to look at what BasisU supports (ETC1, and an ETC2 variant made of "an 8-bit alpha block followed by an ETC1 block").

ASTC is mainly because there are so many variations of it (every block size is its own pixel format), and because I have no hardware with ASTC support, so I would have no way to test that). Also BasisU currently doesn't support ASTC as output format.

As far as I have seem, the WebGL extension for ETC2 formats is called WEBGL_compressed_texture_etc, see here: https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_compressed_texture_etc.

Issue #195, opened 2019-08-19
sokol_app: RaspberryPI support


septag:
I wanted to know if you are interested in RaspberryPI support for sokol_app with Direct FB access. In case you don't know, there is a mode in RaspberryPI which does not need X11 or any kind of graphical shell, you can setup GL surface directly from the command shell and bring up the graphics. which is the thing that I'm trying to add to sokol.

I'm new to the platform, but I've managed to run it with the bare-bone features, mouse and keyboard inputs can be read from `/dev/input` files and will be emulated to match the API behavior. Some other features like window events will not be supported obviously.

If you are interested, I can submit a pull request tomorrow.

floooh:
Yep that's the plan. I have this working for Oryol, but didn't get around adding this to sokol_app.h.

This is the GL context setup for RaspberryPi (most of this is EGL and shared with the Android code path:

https://github.com/floooh/oryol/blob/master/code/Modules/Gfx/private/egl/eglDisplayMgr.cc

The RaspberryPi specific stuff is here:

https://github.com/floooh/oryol/blob/043683dcb3181beb64ae1c85ea76e4a4eb71c124/code/Modules/Gfx/private/egl/eglDisplayMgr.cc#L118-L151

The lowlevel mouse/keyboard input code is here (this read from /dev/input):

https://github.com/floooh/oryol/blob/master/code/Modules/Input/private/raspi/raspiInputMgr.cc

A PR would be nice :)

septag:
Oh cool.
Yes, as for EGL I already took most of it from the android code with very minor changes.
Thanks for the code samples and resources, the actual documentation for rpi APIs are very much lacking.

PR #199, opened 2019-08-20
sokol_app raspberry pi 

septag:
here's a draft version of rpi.  #195 

I didn't find any convenient way of finding the platform with compiler defines, so I defined `__RPI__` in the build system.

looking forward for your input and suggestions.


floooh:
I'll try to have a look soon-ish, as with the sokol_audio.h PR it can take a little while though, apologies in advance :)

meshula:
> I didn't find any convenient way of finding the platform with compiler defines, so I defined __RPI__ in the build system.

I've searched as well, and can't find anything.

floooh:
I've actually thought today that I should get myself a Pi4 ;)

meshula:
It's really fun to hack with.

AndrewCapon:
Hi guys, is PI support going to be coming to Sokol?

floooh:
@AndrewCapon: It's currently not a high priority. The X11/GLX-backend in sokol_app.h might actually work on modern Raspis (I'm currently away from my RPi4 and can't check though). The port in this PR is EGL+VideoCore based which would also work without window system.

(but I'll go ahead and close the PR now because it has diverged too much from the current code, it's probably easier to start from scratch).

colesnicov:
> here's a draft version of rpi. #195
> 
> I didn't find any convenient way of finding the platform with compiler defines, so I defined `__RPI__` in the build system.
> 
> looking forward for your input and suggestions.

What model RPI you are use?

Issue #205, opened 2019-08-31
Problems with #pragma once


martincohen:
I'm using single-translation unit approach and `#pragma once` makes it difficult for me to work with stb-style libraries that use it. In single-translation unit I do this, for example:

`main.h`
```
#include <sokol_gfx.h>
// ...
```

`main.c`
```
#include "main.h"

#define SOKOL_IMPL
#include <sokol_gfx.h>
#undef SOKOL_IMPL
````
In `main.c` the implementation from `sokol_gfx.h` is not included because I've already included `sokol_gfx` in header mode in `main.h`. Fix is using old-style `#ifdef` / `#define` / `#endif` include guard.

I'd be happy to fix this in PR, if you'd agree.

floooh:
Oh interesting, I haven't thought of this case before. Yes, we can switch to traditional include guards, AFAIK compilers have special handling for this pattern so it's just as efficient.

I suggest that you reuse the existing SOKOL_xxx_INCLUDED defines as include guard define, e.g. for sokol_gfx.h it would be SOKOL_GFX_INCLUDED:

https://github.com/floooh/sokol/blob/d476a22a51f195976fb6a61a449603125b8cbb9f/sokol_gfx.h#L502

Most sokol headers should already have such a define, it just needs to move up above the header comment and surrounded by an ifdef/endif.

martincohen:
Yes, that's how I keep patching it when I'm updating. :) I'll try to push out a PR.

floooh:
fixed in #206

PR #206, opened 2019-09-02
ifdef guards

martincohen:
As per https://github.com/floooh/sokol/issues/205, I've changed `#pragma once` to `ifdef` guards.

Some formatting fixes happened in there. I can re-PR if you don't want them in.


floooh:
Thanks! I was a bit confused at first because I would have expected that the closing #endif should go at the end of the entire file, not at the end of the declaration block, but the STB headers seem to use the same convention.

Issue #204, opened 2019-08-30
sokol_gfx: Clamp to border option for sg_image


martincohen:
Hello, I'd like to discuss possibility of adding *clamp-to-border* wrap mode and *border-color* option to `sg_image_desc` in `sokol_gfx`. Is this planned? Should I attempt to add it?

I generally use it to render geometric primitives with white border and UV set to -1, -1. It bypasses the need for having a white pixel in texture and simplifies the drawing code.

By my lousy research it seems that WebGL, OpenGL, DirectX and Metal all do support it.

floooh:
I can't find support in WebGL (there's an GLES2 extension but that doesn't seem to exist for WebGL).

But we can implement it as an optional feature, so:

- add a feature bit "image_border_color" (or a better name) to the sg_features struct
- add a SG_CLAMP_TO_BORDER enum to sg_wrap enum
- add a "float border_color[4]" to sg_image_desc

A PR would be welcome (also if you just implement for a single backend, I can then take over for the others).

martincohen:
What's the strategy for testing the new features? I'm now using only DirectX, and tested it there with my own code. However, so far I have no code running for OpenGL, and Metal is out of my comfort zone. Do we add a sample testing the feature out?

floooh:
Yes, a simple sample for testing/visualizing the texture addressing modes would be nice, nothing fancy.

Just implement the things you want, and then I can take over.

floooh:
See: https://github.com/floooh/sokol/pull/209 and https://github.com/floooh/sokol-samples/pull/46

I'll update the documentation and READMEs next (tomorrow or so) and then I think it's ready for merging.

floooh:
This has now been merged to master.

PR #207, opened 2019-09-03
Border color for image in DirectX

martincohen:
- Added support for `CLAMP_TO_BORDER` in `sg_image` for DirectX.
- Added feature `image_border_color`, now only true for DirectX.

https://github.com/floooh/sokol/issues/204

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -739,6 +739,7 @@ typedef struct sg_features {
     bool msaa_render_targets:1;
     bool imagetype_3d:1;        /* creation of SG_IMAGETYPE_3D images is supported */
     bool imagetype_array:1;     /* creation of SG_IMAGETYPE_ARRAY images is supported */
+    bool image_border_color:1;
 } sg_features;
 
 /*
@@ -964,6 +965,7 @@ typedef enum sg_wrap {
     _SG_WRAP_DEFAULT,   /* value 0 reserved for default-init */
     SG_WRAP_REPEAT,
     SG_WRAP_CLAMP_TO_EDGE,
+    SG_WRAP_CLAMP_TO_BORDER,
     SG_WRAP_MIRRORED_REPEAT,
     _SG_WRAP_NUM,
     _SG_WRAP_FORCE_U32 = 0x7FFFFFFF
@@ -1469,6 +1471,7 @@ typedef struct sg_image_desc {
     sg_wrap wrap_v;
     sg_wrap wrap_w;
     uint32_t max_anisotropy;
+    float border_color[4];
     float min_lod;
     float max_lod;
     sg_image_content content;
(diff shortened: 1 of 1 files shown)


floooh:
...btw, not sure if you're aware of it, but the whole PR workflow works much better when you create a separate branch (I noticed that the PR is created from a forked master branch) :)

I'm starting to look into Metal and GL support now...

floooh:
...I'll be working on this branch, you can continue adding changes to this PR, I'll track and merge those to my branch.

martincohen:
> ...btw, not sure if you're aware of it, but the whole PR workflow works much better when you create a separate branch (I noticed that the PR is created from a forked master branch) :)
> 
> I'm starting to look into Metal and GL support now...

I wasn't a contributor before, so forking was the only way. And since I've had it setup from the first PR, I've used it for this PR as well. :)

floooh:
> I wasn't a contributor before, so forking was the only way. And since I've had it setup from the first PR, I've used it for this PR as well. :)

Ah, no, I mean first do a fork of course, but then create a branch to work on just for that PR. After the PR is merged, that branch can be deleted.

Then when starting work on another PR, first bring the forked repo's master branch uptodate by merging changes from "upstream", and create a new branch.

But I guess if you were not running into trouble working on the master branch that's fine, from my side I haven't had problems (for instance when merging the PR) either yet :)

martincohen:
Yah, working from master was without a problem. :)

floooh:
Stumbled over a little problem:

Metal only allows 3 hardwired border colors: black, white and transparent:

https://developer.apple.com/documentation/metal/mtlsamplerbordercolor?language=objc

WebGPU doesn't seem to support clamp-to-border at all:

https://gpuweb.github.io/gpuweb/#samplers

Opinions? Would your use-case work with one of the 3 hardwired border colors from Metal?

floooh:
PS: since you mention in https://github.com/floooh/sokol/issues/204 that you use a white border, I think I'll just go ahead and use the "3 color approach", the ```float border_color[4]``` member in sg_image_desc would be replaced with an enum ```sg_border_color``` with SG_BORDERCOLOR_TRANSPARENT_BLACK and SG_BORDERCOLOR_OPAQUE_BLACK, SG_BORDERCOLOR_OPAQUE_WHITE.

floooh:
...oh, and border color also isn't available at all in Metal on iOS... :(

That's the reason why WebGPU doesn't implement it.

What a mess...

martincohen:
@floooh White is perfectly sufficient. I'm also thinking whether to revert back and keep the library simple...

floooh:
I decided to go on with the black/white/transparent version. Metal is done, I'm currently working on the GL version. It's very little code and doesn't change the public API (except for some new members in structs), so it's ok to keep it in, even though it's not really useful for cross-platform usage. I also started an extremely simple example:

Screen Shot 2019-09-07 at 2 34 34 PM

I also decided to fall back to clamp-to-edge if the feature isn't available without a hard validation error (but applications can check the sg_features flag of course.

floooh:
Closing this PR now, as it's become the base of this PR: https://github.com/floooh/sokol/pull/209

martincohen:
> Closing this PR now, as it's become the base of this PR: #209

Sure thing!

Issue #208, opened 2019-09-05
sokol_app: Show/hide mouse cursor (followup tasks)


martincohen:
A way to show/hide cursor (useful for games with cursor shown again in debug mode).

Maybe `sapp_show_mouse_cursor(bool)` that has no effect on platforms where it doesn't make sense?

I can PR Windows implementation.

floooh:
Yes, this would also be a good first step towards a proper pointer-lock mode (I started a while ago with code from GLFW, but it's quite a lot of code unfortunately).

floooh:
This is now implemented, just forgot to close the ticket.

PR #210, opened 2019-09-08
208: sokol_app: Show/hide mouse cursor

martincohen:
Implementing [issue 208](https://github.com/floooh/sokol/issues/208) for Windows:

- Showing and hiding mouse cursor with `sapp_show_mouse(bool)`.
- Querying mouse visibility with `bool sapp_mouse_shown()`.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3866,7 +3866,9 @@ _SOKOL_PRIVATE void _sapp_win32_show_mouse(bool shown) {
 }
 
 _SOKOL_PRIVATE bool _sapp_win32_mouse_shown(void) {
-    CURSORINFO cursor_info = { .cbSize = sizeof(CURSORINFO) };
+    CURSORINFO cursor_info;
+    memset(&cursor_info, 0, sizeof(CURSORINFO));
+    cursor_info.cbSize = sizeof(CURSORINFO);
     GetCursorInfo(&cursor_info);
     return (cursor_info.flags & CURSOR_SHOWING) != 0;
 }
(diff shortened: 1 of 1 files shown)


sokol_app.h
```diff
@@ -3857,6 +3861,16 @@ _SOKOL_PRIVATE bool _sapp_win32_utf8_to_wide(const char* src, wchar_t* dst, int
     }
 }
 
+_SOKOL_PRIVATE void _sapp_win32_show_mouse(bool shown) {
+    ShowCursor((BOOL)shown);
+}
+
+_SOKOL_PRIVATE bool _sapp_win32_mouse_shown(void) {
+    CURSORINFO cursor_info = { .cbSize = sizeof(CURSORINFO) };
```
floooh: Please no C99 in the implementation code, must also compile as C++ :)

floooh: PS: since something like ```CURSORINFO cursor_info = {0};``` also creates warnings on some compilers, the recommended way to clear out a struct in the sokol implementations is a good old memset().

martincohen:
I'll check C++ compiler then too for the future PRs. :)

floooh:
> I'll check C++ compiler then too for the future PRs. :)

In the sokol-samples project there's a C++ and ObjC++ compilation test which should catch most incompatible parts (but that needs to be compiled for each backend on different platforms, so it's mostly useful for running during CI like Travis or AppVeyor.

https://github.com/floooh/sokol-samples/blob/master/tests/cpp-compile-test.cc

https://github.com/floooh/sokol-samples/blob/master/tests/objcpp-compile-test.mm

PR #211, opened 2019-09-08
141: sokol_app: UTF-8 argv on Windows

martincohen:
Adding support for UTF-8 argv on Windows.

@floooh I'm not sure what's the preference for dealing with unexpected states (OS calls failing), I've used `SOKOL_LOG` for now, but am willing to change to whatever your preference is.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -19,6 +19,8 @@
     SOKOL_NO_ENTRY      - define this if sokol_app.h shouldn't "hijack" the main() function
     SOKOL_API_DECL      - public function declaration prefix (default: extern)
     SOKOL_API_IMPL      - public function implementation prefix (default: -)
+    SOKOL_CALLOC        - your own calloc function (default: calloc(n, s))
+    SOKOL_FREE          - your own free function (default: free(p))
 
     Optionally define the following to force debug checks and validations
     even in release mode:
@@ -867,7 +869,7 @@ SOKOL_API_DECL const void* sapp_android_get_native_activity(void);
     #include <assert.h>
     #define SOKOL_ASSERT(c) assert(c)
 #endif
-#if !defined(SOKOL_CALLOC) && !defined(SOKOL_FREE)
+#if !defined(SOKOL_CALLOC) || !defined(SOKOL_FREE)
     #include <stdlib.h>
 #endif
 #if !defined(SOKOL_CALLOC)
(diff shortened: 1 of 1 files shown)


floooh:
Looks good, but a few minor points:

1. It seems CommandLineToArgvW() can fail, see here: https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-commandlinetoargvw, this should be handled at least as far as preventing the nullptr access that would happen when w_argv is used later. 

2. I think in case of any error in _sapp_win32_command_line_to_utf8_argv(), a single dummy arg should be returned, so that sokol_main() isn't called with an "unusual" argv/argc pair that would need special handling in sokol_main(). Maybe name that dummy arg something obvious like "INVALID_CMDLINE_ARGS".

3. Please replace the HeapAlloc() / HeapFree() with SOKOL_MALLOC / SOKOL_FREE to be somewhat consistent with the other sokol headers:
    - mention SOKOL_MALLOC / SOKOL_FREE at the top with the other macro overrides: https://github.com/floooh/sokol/blob/b45530fdca04887f42037de2034efb2c254d9e6f/sokol_gfx.h#L36-L37
    - use malloc/free as defaults down in the implementation like here: https://github.com/floooh/sokol/blob/b45530fdca04887f42037de2034efb2c254d9e6f/sokol_gfx.h#L2115-L2119
   No way around CommandLineToArgvW() doing its own allocation unfortunately I guess :)

martincohen:
@floooh For error handling, I think failing silently is much more dangerous for certain uses of `argv`. Most of the cases where OS calls fail are really either insufficient memory or invalid data. I'd simply log and abort immediately. For future, we can provide error callback or `SOKOL_PANIC` macro that will be called with these kinds of errors, if needed.

What do you think?

martincohen:
@floooh There already is `SOKOL_CALLOC` used, but not documented. So I'll use (and document) that instead of adding `SOKOL_MALLOC`, if you're ok with it.

floooh:
> error handling...

Agreed. But I just remembered that there's already something along those lines. Look for the _sapp_fail() function: 

https://github.com/floooh/sokol/blob/425e906d2af8c976c30f7e925a303fea9f0bf879/sokol_app.h#L945-L956

It's used already for fatal error situations:

https://github.com/floooh/sokol/blob/425e906d2af8c976c30f7e925a303fea9f0bf879/sokol_app.h#L3631-L3633

I think you can use this as is.

> SOKOL_MALLOC vs SOKOL_CALLOC:

Yes, SOKOL_CALLOC is fine.

floooh:
PS: there's a chicken-egg-situation for that error-callback in _sapp_fail() though, since the cmdline arg conversion needs to happen before sokol_main() is even called, there's no other way then aborting (_sapp_fail() should still work as expected with a SOKOL_LOG() + SOKOL_ABORT() since static globals are zero-initialized, so the callback function pointers are zero)

martincohen:
> PS: there's a chicken-egg-situation for that error-callback in _sapp_fail() though

Yeh, maybe that callback can be defined as a macro as opposed to a function-pointer, but that change would break the API.

floooh:
> Yeh, maybe that callback can be defined as a macro as opposed to a function-pointer, but that change would break the API.

IMHO just don't worry about that yet and use _sapp_fail(), at least it's somewhat centralized then when we clean that up at a later time.

floooh:
Ok, merged, thanks!

floooh:
Whoops I'm getting a couple type conversion warnings. I'll clean those up on the master branch :)

floooh:
...and I had to do some more changes for the case where SOKOL_WIN32_FORCE_MAIN is defined, nothing big though, I just properly split those two cases (main vs WinMain) into their own functions now.

PR #212, opened 2019-09-12
support 16-bit unsigned normalized vertex formats

aumfer:
Added `SG_VERTEXFORMAT_USHORT2N` and `SG_VERTEXFORMAT_USHORT4N`

I don't have the ability to test these across all platforms atm, but I think they should work.

https://docs.microsoft.com/en-us/windows/win32/api/dxgiformat/ne-dxgiformat-dxgi_format
https://developer.apple.com/documentation/metal/mtlvertexformat?changes=latest__3_3_1&language=objc

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -10314,7 +10314,7 @@ _SOKOL_PRIVATE bool _sg_validate_apply_pipeline(sg_pipeline pip_id) {
             /* default pass */
             SOKOL_VALIDATE(pip->color_attachment_count == 1, _SG_VALIDATE_APIP_ATT_COUNT);
             SOKOL_VALIDATE(pip->color_format == _sg_default_rendertarget_colorformat(), _SG_VALIDATE_APIP_COLOR_FORMAT);
-            SOKOL_VALIDATE(pip->depth_format == _sg_default_rendertarget_depthformat(), _SG_VALIDATE_APIP_DEPTH_FORMAT);
+            SOKOL_VALIDATE(pip->depth_format == _sg_default_rendertarget_depthformat() || pip->depth_format == SG_PIXELFORMAT_NONE, _SG_VALIDATE_APIP_DEPTH_FORMAT);
             /* FIXME: hmm, we don't know if the default framebuffer is multisampled here */
         }
         return SOKOL_VALIDATE_END();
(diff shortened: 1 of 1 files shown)


floooh:
I'm pretty sure I had to leave those out because they're not supported on some GL implementations, I need to find the places in the spec again though.

floooh:
Hmm nevermind, it's allowed in WebGL: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/vertexAttribPointer, guess I was confusing this with the cross-platform restrictions which I already mentioned in the documentation comment block above the sg_vertex_format enum.

floooh:
The vertex format changes look good. But can you undo the depth-format validation check in this PR please (because it's unrelated to the PR, and I don't know yet whether it's a good change... this change should better go together with providing the default framebuffer pixel format and sample count as part of the sg_desc structure in sg_setup(), and check against those. That's a separate feature though.

amerkoleci:
Make sure to set normalized state when enable the attribute on opengl backend

floooh:
> Make sure to set normalized state when enable the attribute on opengl backend

As far as I can see, the PR handles that already in this change block:

https://github.com/floooh/sokol/pull/212/commits/cc9ae1e4d55c1112afdabbaf08c9fe2f8b2e47ce#diff-a8b646082914484fa2639b5ad5947b55R3895-R3903

(that's inside the function ```_sg_gl_vertexformat_normalized```)

amerkoleci:
My bad then

aumfer:
I removed the validation check.

The reasoning for that is that I have a pipeline used for an offscreen render pass that doesn't need a depth buffer (gaussian blur), and I wanted to be able to reuse that pipeline to draw into the default framebuffer. As it is, I'd have to duplicate the pipeline or add an unneeded depth buffer.

Then again, I don't think the pipeline actually manages and GL resources, so duplicating is probably no big deal (reusing the shader instead of the whole pipeline)

floooh:
> and I wanted to be able to reuse that pipeline to draw into the default framebuffer

That approach wouldn't work in Metal unfortunately (see here: https://developer.apple.com/documentation/metal/mtlrenderpipelinedescriptor?language=objc, search for "Specifying Rendering Pipeline State"), you'd need two pipeline objects for rendering into differently configured target framebuffers.

But generally I'd like the configuration of the default framebuffer more flexible, I'll write a reminder ticket that the configuration of the default framebuffer should be passed into sg_setup() (with the defaults being the same as now, so its backward compatible).

floooh:
Ok merged. Thanks!

PR #217, opened 2019-09-17
Add support for srgb texture creation in sokol_app.

amerkoleci:
Currently it supports Direct3D11 and Metal, not sure how it would work on OpenGL/GLES.


floooh:
Sorry, sRGB was left intentionally out of sokol_gfx.h because it has all sorts of problems on OpenGL (mostly GLES2 and GLES3). Doing it only in sokol_app.h alone also isn't enough, sokol_gfx.h also needs support for sRGB textures and render passes.

It's trivial to do the sRGB/linear decoding/encoding in the pixel shader as needed, and it's much more "localized" to do it there (not leaking out into the C code and asset pipeline).

PR #220, opened 2019-09-20
Fix _sapp_android_init_egl() failure on Xiaomi Redmi 6A device 

elix22:
This failure might also happen on other Android devices.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -4528,6 +4528,11 @@ _SOKOL_PRIVATE bool _sapp_android_init_egl(void) {
     EGLint alpha_size = _sapp.desc.alpha ? 8 : 0;
     const EGLint cfg_attributes[] = {
         EGL_SURFACE_TYPE, EGL_WINDOW_BIT,
+#if defined(SOKOL_GLES3)
+		EGL_RENDERABLE_TYPE, EGL_OPENGL_ES3_BIT,
+#else	
+		EGL_RENDERABLE_TYPE, EGL_OPENGL_ES2_BIT,
+#endif
         EGL_RED_SIZE, 8,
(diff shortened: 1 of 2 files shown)


fabiopolimeni:
Just to let you know this also worked for me on the _RETROID_ 2+ (Android 9). It would fail to create the EGL surface otherwise. It looks a simple fix/change, maybe worth a merge?

floooh:
Yes, makes sense, thanks for bringing the PR to my attention again. I need to look into a sokol_audio.h problem first and then will look into this PR.

floooh:
I haven't merged the PR (since it's outdated and also contains unrelated sokol_audio.h changes, but instead I added this to the sokol_app.h backend:

```c
        EGL_SURFACE_TYPE, EGL_WINDOW_BIT,
        #if defined(SOKOL_GLES3)
		    EGL_RENDERABLE_TYPE, _sapp.desc.gl_force_gles2?EGL_OPENGL_ES2_BIT:EGL_OPENGL_ES3_BIT,
        #else
		    EGL_RENDERABLE_TYPE, EGL_OPENGL_ES2_BIT,
        #endif
        EGL_RED_SIZE, 8,
```

(note how the gl_force_gles2 config flag is checked, this also happens below when creating the GL context)

Also see commit https://github.com/floooh/sokol/commit/67c321cdbb9c46e2944e77ca7a4ba3a611b80555

Unfortunately I currently don't have a way to test this on my (Pixel4) phone, since I don't seem to have an USB to UCB-C cable, I'll see if I can buy one tomorrow.

floooh:
Ok, tested a bit on my Pixel4 with the sokol-samples, and at least there it's working.

PR #238, opened 2019-12-05
sokol_app.h: fix for assignment within conditional expression

incrediblejr:


floooh:
Oops how embarrasing :) I need to figure out why I didn't get a warning. Thanks for the fix!

floooh:
...ok I didn't get a warning because it's Win32 specific code, it's a warning-level-4 warning on MSVC and I only compile the sokol-sapp samples with warning level 3. I'll see what it takes to make the code warning-level-4 clean on MSVC.

PR #258, opened 2020-01-30
Fix typos.

waywardmonkeys:


floooh:
As always: thanks a lot Bruce :)

PR #259, opened 2020-01-31
[WIP] WebGPU backend for sokol_gfx.h

floooh:
Initially only for WASM / emscripten, but plan is to make it also work for the native WebGPU implementation libraries (Google's Dawn and Mozilla's wgpu).


amerkoleci:
Do you plan also to refactor the api? maybe expose command buffer or command encoder?

floooh:
> Do you plan also to refactor the api? maybe expose command buffer or command encoder?

Nope, this will just be a straight backend without changing the public API. Command buffers definitely won't be exposed in sokol-gfx because this would require a lot of work for 3D APIs that don't have them.

One thing that might make sense, or might even be needed is this though:

https://github.com/floooh/sokol/issues/262

PR #264, opened 2020-02-18
Warning fixes

iboB:


floooh:
Looks good, thanks! That gcc9 warning is interesting, I'm surprised that this doesn't happens in many more places in my code.

PR #269, opened 2020-02-28
sokol_app.h should be included before sokol_gfx.h

caiiiycuk:
Hi, if I understand correctly idea then proposal changes is correct.


floooh:
Indeed, thanks for catching that :)

PR #271, opened 2020-02-29
Do not register animation frame if frame callback is not set

caiiiycuk:
Hi @floooh 
As you probably know js-dos works in synchronous mode, and used asyncify for drawing. In my special case frame callback is useless, and I does not provide it. Of course you current code will work without error, but I am worry about useless listener registration. 

I am not sure if I need to set `_sapp.first_frame = false`, and `_sapp.frame_count = 1`  in else branch. I totally if agree if you dislike proposed changes. I think I can do them locally anyway, if nothing will break.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2836,7 +2836,12 @@ _SOKOL_PRIVATE void _sapp_run(const sapp_desc* desc) {
     emscripten_set_touchcancel_callback(_sapp.html5_canvas_name, 0, true, _sapp_emsc_touch_cb);
     emscripten_set_webglcontextlost_callback(_sapp.html5_canvas_name, 0, true, _sapp_emsc_context_cb);
     emscripten_set_webglcontextrestored_callback(_sapp.html5_canvas_name, 0, true, _sapp_emsc_context_cb);
-    emscripten_request_animation_frame_loop(_sapp_emsc_frame, 0);
+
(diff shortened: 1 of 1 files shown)


floooh:
I like the idea but not this specific PR ;)

Can you share the code how your frameloop looks like (mostly the part the part where control is yielded back to emscripten).

I'm thinking about some sort of optional "inversion of control mode" which would only work on some platforms, where sokol-app doesn't offer any callbacks, but instead traditional sapp_init(), sapp_shutdown(), sapp_poll_events() and sapp_present() (or similar).

I'm absolutely not sure yet whether it makes sense to split the API like that though, so for now I'd suggest you apply the necessary hacks in a fork of yours.

caiiiycuk:
The js-dos loop is traditional. Dosbox start with init code, then inifinite loop starts. Each loop do dos emulation ticks, poll events, and update surface. So surface is updated inside this loop, not from raf callback. End of each loop have check how long it's doing in infinite mode, and if it's more than 16ms, asyncify stop execution for very short time.

This scheme works not so bad as it sounds :)  

BUT, after some thinking, I realized that it's not a problem to have frame callback. And I decided to rewrite code. Now frame is updated inside raf callback. 

Dosbox loop stay the same, but instead of updating surface (frame) inside loop, it's just updates the rgba array, which used in raf callback to update texture that renders to canvas.

It's requires little bit more memory (+1 rgba array) and one addtitional memcpy (that can be avoided in cost of uglifying code).

PR #276, opened 2020-03-15
Fix #275

### LINKED ISSUE #275: html5: sapp_quit
Hi. Looks like `sapp_quit` is not working for html5 backend. I have a test suite for js-dos, that runs new instances of js-dos one by one on same canvas. Everything works, but previous instances of js-dos tries to update canvas even if they are stopped. To stop them I used `sapp_quit`.

It seems that sapp_quit is no sense for html backend. I think it should:
* unregister all events listeners
* unregister raf callback

OT: How you develop sokol_app project? For me it's very hard to find functions relative to emscripten implementation. I using just emacs, can you give any hints? Thanks!

caiiiycuk:
First try to fix #275 
What might be wrong:
* Myabe is better to call `_sapp_emsc_unregister_events` from `_sapp_call_cleanup/_sapp_emsc_call_cleanup`
* Probably need to unregister `html5_ask_leave_site` hook to
* Not sure about clipboard listeners

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -91,7 +91,7 @@
     RESTORED            | YES     | YES   | YES   | ---   | ---     | ---   | ---
     SUSPENDED           | ---     | ---   | ---   | YES   | YES     | ---   | TODO
     RESUMED             | ---     | ---   | ---   | YES   | YES     | ---   | TODO
-    QUIT_REQUESTED      | YES     | YES   | YES   | ---   | ---     | TODO  | ---
+    QUIT_REQUESTED      | YES     | YES   | YES   | ---   | ---     | TODO  | YES
     UPDATE_CURSOR       | YES     | YES   | TODO  | ---   | ---     | ---   | TODO
     IME                 | TODO    | TODO? | TODO  | ???   | TODO    | ???   | ???
     key repeat flag     | YES     | YES   | YES   | ---   | ---     | TODO  | YES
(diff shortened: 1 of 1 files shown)


floooh:
Hi, what's there so far looks good IMHO, I'll try to do some testing in the next few days. Currently I'm a bit distracted because of various work-from-home issues :)

caiiiycuk:
Np, fyi this changes are enough to pass js-dos tests.

caiiiycuk:
Hi, updated this PR with correct unregister leave & clipboard hook.
Probably something similar need to do with textinput hooks, but I will do this when I face this problem (in fat future) :)

floooh:
Thanks a lot! As you have noticed, I took a little break from working on PRs and other fixes to focus on the sokol-gfx/app WebGPU branch. I'll hope to finish this up "soon-ish" :)

floooh:
Ok... going to merge this now and do some testing...

floooh:
There's quite a few conflicts now after my WebGPU merge. I'm going to perform the merge "manually" in a separate branch.

floooh:
I think I found a bug in your code while merging: when unregistering events, you're calling ```sapp_add_js_hook_clipboard``` instead of ```sapp_remove_js_hook_clipboard```, I'll fix that right away in the merge though. Need to shuffle things around a bit anyway because of the collisions with the WebGPU-specific stuff.

floooh:
Ok, merged to master, but as I said I had to shuffle things around a bit, would be good if you can test whether the code still behaves as expected on your side.

I changed the quit-requested code in _sapp_emsc_frame() so that it is possible to perform a "soft quit" where the application has a chance to display a custom "Really quit" dialog box (I'll enable this in the imgui-high-sapp demo):

https://github.com/floooh/sokol/blob/87eb805c313197b35af6cec90fe9153ead56594b/sokol_app.h#L3075-L3089

I'll also add a blurb to the documentation block and the README.

And thanks for the PR again :)

floooh:
PS: While updating the whole "application quit" documentation block in sokol_app.h I was thinking that maybe the "opt-in behaviour" isn't actually needed. I think I'm going to remove the ```html5_enable_shutdown``` flag (I know, I suggested that flag in the first place, but I didn't think this properly through at the time).

caiiiycuk:
Thanks! I also think that `html5_enable_shutdown` isn't needed. I will test new version of sokol asap. Currently I refactoring everything and not have any working version of js-dos. :p

PR #289, opened 2020-05-05
Improve `sokol_app.h` support for `SOKOL_GLCORE33` on Windows

garettbass:
Two small adjustments to `sokol_app.h`:

1. Add `#pragma comment (linker, "/subsystem:windows")` to streamline command line compilation.
2. Link with `gdi32.lib` and `user32.lib` in `_WIN32` builds to support `SOKOL_GLCORE33`.


garettbass:
I have decided a slightly different refactor better maintains the configuration that relies on `WindowsApp.lib`

PR #290, opened 2020-05-05
[sokol_app.h] Windows linker/lib pragmas for command line compilation

garettbass:
1. Added `#pragma comment (linker, "/subsystem:windows")` to avoid having to specify this on the command line (works with cl.exe and clang)
2. Added `#pragma comment (lib, "user32.lib")` for both `SOKOL_D3D11` and `SOKOL_GLCORE33`
3. Added `#pragma comment (lib, "gdi32.lib")` for `SOKOL_GLCORE33`

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3271,6 +3271,20 @@ _SOKOL_PRIVATE const _sapp_gl_fbconfig* _sapp_gl_choose_fbconfig(const _sapp_gl_
 
 #pragma comment (linker, "/subsystem:windows")
 
+#if (defined(WINAPI_FAMILY_PARTITION) && !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP))
+#pragma comment (lib, "WindowsApp.lib")
+#else
+#pragma comment (lib, "user32.lib")
+#if defined(SOKOL_D3D11)
+#pragma comment (lib, "dxgi.lib")
+#pragma comment (lib, "d3d11.lib")
+#pragma comment (lib, "dxguid.lib")
+#endif
+#if defined(SOKOL_GLCORE33)
+#pragma comment (lib, "gdi32.lib")
+#endif
+#endif
+
 #if defined(SOKOL_D3D11)
 #ifndef D3D11_NO_HELPERS
 #define D3D11_NO_HELPERS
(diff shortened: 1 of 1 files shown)


floooh:
Good fixes, thanks! (I wish all compilers would have a pragma for that, it's perfect for single-file libs)

floooh:
Oopsi, looks like the change broke the Windows CI build (and...uh the Android build?):

https://github.com/floooh/sokol/runs/646555690?check_suite_focus=true

Investigating...

floooh:
...ok the Windows builds breaks because of the subsystem:windows switch, sokol_app.h has a mode where the regular main.cpp can be used by defining SOKOL_WIN32_FORCE_MAIN, and this is used in the cpp-compile-test code:

https://github.com/floooh/sokol-samples/blob/bcb8550d541171acd4e13f0a797c659f03e3e859/tests/cpp-compile-test.cc#L6

I'll try to fix this by simply putting that pragma inside an ifdef-check...

Not sure yet what's up with the Android build, that's probably something entirely different (maybe a build system problem)

floooh:
Ok, fix in https://github.com/floooh/sokol/commit/45a9269444abdcb8b3e6553c326de89ecf80d477

...I really need to come up with a better CI configuration, the test build is currently broken for branches and PRs, because the CI simply clones the sokol-samples repository and runs a build on it, and this will always clone the master branch of the sokol-repository as dependency, which in turn is very misleading for the PR workflow, because it *looks* like all checks were passing when they actually didn't...

floooh:
...hmm and for some mysterious reason the Android CI build now works again too. Definitely cannot have been the code changes, more likely some spooky Github Actions thing...

garettbass:
Thanks for sorting it out.  The CI gave me a false sense of confidence :)

garettbass:
Is this because fips is choosing which branch of sokol to pull?

floooh:
> Is this because fips is choosing which branch of sokol to pull?

The "problem" is that the sokol-repository itself cannot be built on its own, so in the CI for the sokol repo I actually clone sokol-samples and run a build on that. This in turn clones fresh repositories for the sokol-sample dependencies, and this causes the sokol master branch to be built instead of the wanted branch. 

It only works as intended for commits to the sokol master branch, but not for branches or PRs.

fips allows to define a tag- or branch-name, or a specific commit hash in the dependency list in fips.yml, but that also doesn't help much in this specific situation.

garettbass:
Ok!  Next time, I will commit and push to my fork and wait for CI to pass there before making a PR.

garettbass:
Now I see that running your workflow on my forked repo won't help, because the workflow explicitly checks out `floooh/sokol-samples`, which will in turn check out `floooh/sokol` :P

PR #291, opened 2020-05-07
reference-based equivalents for c++

garettbass:
I actually wrote this last night before posting to Twitter about my stupid C++ trick 🤓
I hope it passes CI!

### DIFF
--- a/sokol_args.h
+++ b/sokol_args.h
@@ -294,6 +294,10 @@ SOKOL_API_DECL const char* sargs_value_at(int index);
 
 #ifdef __cplusplus
 } /* extern "C" */
+
+/* reference-based equivalents for c++ */
+inline void sargs_setup(const sargs_desc& desc) { return sargs_setup(&desc); }
+
 #endif
 #endif // SOKOL_ARGS_INCLUDED
 
--- a/sokol_audio.h
+++ b/sokol_audio.h
@@ -417,6 +417,10 @@ SOKOL_API_DECL int saudio_push(const float* frames, int num_frames);
 
 #ifdef __cplusplus
 } /* extern "C" */
+
+/* reference-based equivalents for c++ */
+inline void saudio_setup(const saudio_desc& desc) { return saudio_setup(&desc); }
+
 #endif
 #endif // SOKOL_AUDIO_INCLUDED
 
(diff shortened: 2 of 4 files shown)


garettbass:
I have only tested `sg_setup()` and `sg_make_buffer()` so far, because my C++ sample is still pretty basic, but I expect these all to work, since they compile, and are pretty simple.

floooh:
I just noticed that you have returns for void functions:

```cpp
inline void sargs_setup(const sargs_desc& desc) { return sargs_setup(&desc); }
```

Don't know if those cause warnings on some compilers, might be better to remove them. If you don't get around to do it, I can do this too after merging (I'll try to catch up with merging in a few days latest).

floooh:
Hmm, and when adding such reference-based functions to the new sokol_debugtext.h header I noticed another problem:

Clang on macOS doesn't let me override those functions unless I remove the ```extern "C" { }``` around the other API functions. Kinda makes sense in a way because the ```extern "C" { }``` disables the C++ name mangling... We may need to add a separate config-define to not do the ```extern "C"``` thing (maybe when a SOKOL_CPP_API is defined).

garettbass:
return void from a void function is a standard C++ feature to support templates.  I used it here because I wanted to ensure that the return types of the new functions remain compatible with the return types of the old functions.  For example, it would a compile error if `sargs_setup(&desc)` were changed in the future to return an `int`.  If we remove the `return` and just call the function, we lose this safety check.

garettbass:
I'm frustrated that macOS clang is behaving in this way.  Windows clang does not object to C++ overloads of `extern "C"` functions.

garettbass:
If you would prefer to abandon this PR, I can work around the issue by writing a C++-specific header file as a wrapper.

garettbass:
[Overload resolution with extern “C” linkage](https://stackoverflow.com/questions/4866043/overload-resolution-with-extern-c-linkage)

Are you sure the C++ overload is ***outside*** of the `extern "C"` block?

garettbass:
Seems like an important standard thing to support, e.g.:
```
complex sqrt(complex); // C + + linkage by default
extern "C" {
    double sqrt(double); // C linkage
}
```

floooh:
>  Are you sure the C++ overload is outside of the extern "C" block?

Erm, no. I've put it inside. I'm an idiot. Please ignore :D

floooh:
> return void from a void function is a standard C++ feature to support templates...

Oki doki thanks for the explanation. I'll try to do a "merge day" towards the end of the week, there's a couple other PR waiting which I also need to look at first.

floooh:
And merged. Thanks!

garettbass:
Thank you!

PR #294, opened 2020-05-10
support none full mipmap chian

kochol:
I added GL_TEXTURE_MAX_LEVEL to sokol_gfx.h to support none full mipmap chain. Without out this fix, the textures are rendered as black.

https://gamedev.stackexchange.com/questions/84398/texture-is-black-when-manually-building-mipmap

https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexParameter.xhtml


kochol:
I tested this on android and windows

floooh:
I vaguely remember compatibility problems with GL_TEXTURE_MAX_LEVEL, IFIR WebGL and GLES2 only support complete mipmap chains, so incomplete mipmap chains is probably something that needs to live behind a runtime feature flag. I basically only anticipated only the two use cases "1 mip surface" or "all mip surfaces", for incomplete mip chain support I also need to check how the other backend APIs behave (it basically needs a new test/sample).

Instead the sg_image_desc.min/max_lod members are intended to limit the mipmap range actually used by the GPU.

I'll try to catch up on PRs over the weekend.

PR #296, opened 2020-05-11
sapp_toggle_fullscreen(), implemented for win32

mattiasljungstrom:
Implements sapp_toggle_fullscreen(). Each call changes window style and layout to either windowed or fullscreen based on the value in _sapp.desc.fullscreen.

The code tries to center the window on switch from fullscreen. If the size of the window is too large for the current monitor, it uses the monitor size.

Not sure if you want this or not, but I figured I'll do a pull request. I've only implemented this for win32, can take a look at macOS later if you want to keep it.

I've only tested this under latest Win10.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -869,6 +869,8 @@ SOKOL_API_DECL float sapp_dpi_scale(void);
 SOKOL_API_DECL void sapp_show_keyboard(bool visible);
 /* return true if the mobile device onscreen keyboard is currently shown */
 SOKOL_API_DECL bool sapp_keyboard_shown(void);
+/* query fullscreen mode */
+SOKOL_API_DECL bool sapp_is_fullscreen(void);
 /* toggle fullscreen mode */
 SOKOL_API_DECL void sapp_toggle_fullscreen(void);
 /* show or hide the mouse cursor */
@@ -1071,6 +1073,7 @@ typedef struct {
     int sample_count;
     int swap_interval;
     float dpi_scale;
+    bool fullscreen;
     bool gles2_fallback;
     bool first_frame;
     bool init_called;
@@ -1201,6 +1204,7 @@ _SOKOL_PRIVATE void _sapp_init_state(const sapp_desc* desc) {
         _sapp_strcpy("sokol_app", _sapp.window_title, sizeof(_sapp.window_title));
     }
     _sapp.dpi_scale = 1.0f;
+    _sapp.fullscreen = _sapp.desc.fullscreen;
 }
 
 _SOKOL_PRIVATE void _sapp_discard_state(void) {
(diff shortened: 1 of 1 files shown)


mattiasljungstrom:
Wanted to check if you think this is something that should be in sokol or not? Otherwise I'll keep it in my sokol_ext.h file.

floooh:
> Wanted to check if you think this is something that should be in sokol or not? Otherwise I'll keep it in my sokol_ext.h file.

It's a good addition, I'm just lagging behind with the Pull Requests :)

I'm not sure yet if a toggle function is enough, usually I have at least one other function which explicitely sets to fullscreen or windowed (and is a no-op if it's already in the right state). I'll try to catch up on PR's after I'm happy with the new sokol_debugtext.h header.

floooh:
A bit of feedback:

- The function shouldn't modify the desc.fullscreen flag because the existing function ```sapp_query_desc()``` is expected to return the original desc struct. So there should be a new "mutable: _sapp.fullscreen flag which gets its initial value from ```_sapp.desc.fullscreen```. The other places where ```_sapp.desc.fullscreen``` is inspected should be changed to look at the new ```_sapp.fullscreen``` flag instead.

- A new function ```bool sapp_is_fullscreen(void)``` which returns the _sapp.fullscreen flag. That way that "explicit" function I mentioned in the above comment isn't needed, instead an application can check the sapp_is_fullscreen() return value to decide whether sapp_toggle_fullscreen needs to be called or not.

mattiasljungstrom:
The feedback makes sense, will update the code.

This is maybe another ticket, but it's related: the size of the non-fullscreen window is in window size, not in render target size. So, if you ask for a 1920x1080 on windows, it's instead 1900x1071 or something in render target size. What's your take on this? One option would be to query windows for the size of the borders, etc. and add them all up.

mattiasljungstrom:
Re my last comment, I noticed that this is correctly handled if the window is setup as windowed directly. Will investigate more why it's incorrect when toggled.

mattiasljungstrom:
Updated based on your feedback, and fixed sizing bug. Tested with high dpi as well.

floooh:
There's a small problem with using designated initialization

```
MONITORINFO minfo = { .cbSize = sizeof(MONITORINFO) };
```

...but I'll fix that manually after the merge.

floooh:
...and another small thing I'm going to "fix in post"

```c
... _sapp_win32_toggle_fullscreen() {
```

should be
```c
... _sapp_win32_toggle_fullscreen(void) {
```

floooh:
...ok, another small fix I did (only slightly related) is to improve the default-value replacement in sapp_desc a bit, so that the _sapp_def() macros in _sapp_win32_toggle_fullscreen() are not needed:

https://github.com/floooh/sokol/commit/061b56ead450de774a4b6d193af8e2ccf53ac42b

floooh:
...oh, and I added a test button to the imgui-highdpi-sapp sample to toggle windowed/fullscreen :)

And I forgot: Thanks for the PR :)

...I'll see if I can also add support for macOS while I'm at it...

mattiasljungstrom:
cool, thanks for merging and fixing!

PR #297, opened 2020-05-16
fixed bunch of unused parameter warnings on metal backend

septag:


floooh:
Whoops, strange that I'm not getting those in Xcode, guess I need to turn up the warning level a bit.

Thanks for the PR! I'll try to catch up with merging after hopefully in a few days!

floooh:
PS: that compilation error on emscripten isn't related to your PR, looks like something's broken when installing the emscripten SDK.

septag:
yes I knew that from my previous PRs :) thanks

floooh:
Ah damn :D I just merged another related PR (https://github.com/floooh/sokol/pull/305), and after that I went ahead and fixed all the other unused-param warnings. I forgot about your pending PR, and this is now redundant (I'll just close it).

The reason that I wasn't seeing those unused-param warnings was that I had a global -Wno-unused-parameter in all gcc and clang build configs in fips (which in hindsight wasn't such a great idea).

PR #305, opened 2020-05-30
Fix unused warnings when compiling with NDEBUG

edubart:
This fixes the following warnings when compiling with `-DNDEBUG -Wall -Wextra` on clang:
```
../sokol_gfx.h:5898:61: warning: unused parameter 'pip' [-Wunused-parameter]
_SOKOL_PRIVATE void _sg_gl_destroy_pipeline(_sg_pipeline_t* pip) {
                                                            ^
../sokol_gfx.h:13394:76: warning: unused parameter 'trace_hooks' [-Wunused-parameter]
SOKOL_API_IMPL sg_trace_hooks sg_install_trace_hooks(const sg_trace_hooks* trace_hooks) {
                                                                           ^
../sokol_gfx.h:14021:53: warning: unused parameter 'name' [-Wunused-parameter]
SOKOL_API_IMPL void sg_push_debug_group(const char* name) {
```

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -5897,6 +5897,7 @@ _SOKOL_PRIVATE sg_resource_state _sg_gl_create_pipeline(_sg_pipeline_t* pip, _sg
 
 _SOKOL_PRIVATE void _sg_gl_destroy_pipeline(_sg_pipeline_t* pip) {
     SOKOL_ASSERT(pip);
+    _SOKOL_UNUSED(pip);
     /* empty */
 }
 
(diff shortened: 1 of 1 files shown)


floooh:
Looks good, thanks! That just made me investigate why I'm not seeing those warnings, and it looks like I added -Wno-unused-parameter a long time ago to the global settings in fips. Guess I should remove those.

floooh:
Ok, I removed that global -Wno-unused-parameters and fixed all other unused-param warnings in sokol and sokol-samples which resulted from this change.

edubart:
Oh nice, that way will be easier to catch, now I've found other warnings under D3D11 backend, made a new PR #307

PR #306, opened 2020-06-01
sokol_gfx: skip unecessary GL calls when we don't care about depth/stencil buffer

edubart:
I was doing some benchmarks and noticed by tracing all GL calls that `glDepthMask` and `glStencilMask` is called twice every frame even though I do not use them at all. So this commit skip calling then when the depth/stencil action is `SG_ACTION_DONTCARE` in `_sg_gl_begin_pass`.  This effectively removes any call to both functions on my tests.


floooh:
Ok merged, thanks! I think more code can be skipped if the action is dont-care, I'll play around with it a bit more.

floooh:
Ok, I did some more changes to the _sg_gl_begin_pass() function. There shouldn't be any clear-related GL calls happening now at all if no clearing is requested:

https://github.com/floooh/sokol/commit/0a61719b98e97a0467f3c0c6ad44d67f6ca14489

I'll do some more testing on Linux and with WebGL.

Issue #308, opened 2020-06-04
sokol_gfx: optimize away uneeded glUniform1i and glActivateTexture when applying different image binding


edubart:
I was doing some benchmarks analysis with sokol_gfx using https://github.com/ValveSoftware/vogl and I noticed sokol_gfx calls repeatably `glUniform1i` and `glActivateTexture` when applying a different texture binding for the same program, those both calls can be optimized away, look at this screenshot of the output of VOGL in my test case:

![](https://i.gyazo.com/5d97659e51c5e53faeb1c79dadf14e35.png)

My test case is just drawing rectangles using the same pipeline, the only change between draw calls is the texture through `sg_apply_bindings`. You can notice in the table that each `glUniform(1, 0)` takes about ~1000 nanoseconds and `glActivateTexture(GL_TEXTURE0)` takes about ~500 nanoseconds, replacing both calls using an if on cached data should take less time. Doing this for `glActivateTexture` should be easy because this is a global state, however doing for `glUniform1i` would need to make sokol_gfx cache the texture id uniform value per shader program, or perhaps easier would be caching the texture sampler id uniform and invalidate it

edubart:
Oh the texture uniform sampler id could be cached in a easy way per pipeline binding and invalidated every time the applied pipeline changes.

I want to share that removing both calls in hacky way (just commenting out makes my test case work the same) improved my test case where I draw a lot of 2D rectangles over ~20% fps under GL 3.3 core backend.

edubart:
I made a PR for the two GL optimizations as shown above, they improved FPS under my 2D textured  rectangles rendering benchmark about 15~20%.

floooh:
Ok this looks promising, thanks for looking into this :)

From quickly glancing over the PRs they look fine, but I'll have a closer look and do some testing before merging. I had some tricky and non-obvious bugs in the past involving the GL backend state cache, so I learned to be extra careful in that area ;)

edubart:
Ok, thanks for considering these patches. Indeed, this should be tested more and thinked about any missing spot with care. As I'm not that familiar with sokol_gfx internals some details could be missing.

floooh:
Ok, everything merged. I think this can be closed now.

PR #309, opened 2020-06-04
sokol_gfx: cache active textures on GL backend

edubart:
This is an optimization on GL backend to avoid redundant calls to glActiveTexture when binding different textures. This slightly improves my FPS on my benchmark tests with sokol_gfx using GLCORE33. As pointed in #308


floooh:
Ok merged. Thanks! I'll check all the sokol-samples and will also update the HTML5 sample webpage.

PR #310, opened 2020-06-04
sokol_gfx: cache texture slots per shader on GL backend

edubart:
This skip unnecessary calls to `glUniform1i` when binding different textures in a shader program under GL backend. This was proposed on #308

Note I take advantage of that a new allocated shaders is initialized to zeros by sokol pool allocator, so `texture_slots` should be all zeros. And that GLSL spec states that program uniforms variables are always initialized to zeros, so this means all slots are 0 for new shaders on both sokol cache and GL program. I also take advatange that the uniform values in a shader persists until linked again to cache per shader.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -2923,6 +2923,7 @@ typedef struct {
         GLuint prog;
         _sg_gl_shader_attr_t attrs[SG_MAX_VERTEX_ATTRIBUTES];
         _sg_gl_shader_stage_t stage[SG_NUM_SHADER_STAGES];
+        int texture_slots[SG_MAX_SHADERSTAGE_IMAGES];
     } gl;
 } _sg_gl_shader_t;
 typedef _sg_gl_shader_t _sg_shader_t;
@@ -6435,6 +6436,13 @@ _SOKOL_PRIVATE void _sg_gl_apply_pipeline(_sg_pipeline_t* pip) {
     }
 }
 
+_SOKOL_PRIVATE void _sg_gl_shader_texture_slot(_sg_shader_t* shd, GLint loc, int slot) {
+    if(shd->gl.texture_slots[loc] != slot) {
+        shd->gl.texture_slots[loc] = slot;
+        glUniform1i(loc, slot);
+    }
+}
+
 _SOKOL_PRIVATE void _sg_gl_apply_bindings(
     _sg_pipeline_t* pip,
     _sg_buffer_t** vbs, const int* vb_offsets, int num_vbs,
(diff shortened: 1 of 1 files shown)


sokol_gfx.h
```diff
@@ -6435,6 +6436,13 @@ _SOKOL_PRIVATE void _sg_gl_apply_pipeline(_sg_pipeline_t* pip) {
     }
 }
 
+_SOKOL_PRIVATE void _sg_gl_shader_texture_slot(_sg_shader_t* shd, GLint loc, int slot) {
+    if(shd->gl.texture_slots[loc] != slot) {
```
floooh, line 6440: This array access will be a problem I think. The index is the GL uniform location, which isn't related to SG_MAX_SHADERSTAGE_IMAGES and might overflow.

I think the texture_slots array should go into _sg_gl_shader_stage_t struct, the array size remains at SG_MAX_SHADERSTAGE_IMAGES, but one entry needs to have two items:

- gl_loc: the GL uniform location (== loc function arg)
- gl_tex_slot: the GL texture slot (== slot argument)

(this should probably become a new struct _sg_gl_shader_stage_texture_slot).

You'd need two new parameters to the _sg_gl_shader_texture_slot() function:

- int stage: the shader stage index
- int img_index: the shader image index (or bind slot)

Both are available in the nested loop that calls the _sg_gl_shader_texture_slot() function.

floooh:
...ok... it just occured to me...

If the glUniform1i() call is part of the GL shader state, then we can probably move the call up into shader creation, and we don't need all that caching stuff?

Both the uniform location and the "texture slot" are resolved here:

https://github.com/floooh/sokol/blob/0a61719b98e97a0467f3c0c6ad44d67f6ca14489/sokol_gfx.h#L5814-L5834

...and these are the parameters to the glUniform1i() call. And if this glUniform1i() is part of the GL shader state then it can happen right there. I'll try that out...

floooh:
Ok, I made a little experimental branch where I moved the glUniform1i() call into the shader creation:

https://github.com/floooh/sokol/compare/tex-uniform-in-create

Can you give this branch a whirl? I need to do more testing here too, but at first glance it seems to work as expected. If it works then this PR can be closed (and the other PR https://github.com/floooh/sokol/pull/309 would work nicely with moving the uniform call into shader creation too) :)

edubart:
I've tested your solution, it worked good on my tests, that approach seems better too because it set the uniforms only once on shader creation. So I guess sokol_gfx doesn't need to change texture slots on the fly and can do only once (I didn't notice this before).

Side note: Would be cool if sokol could optionally cache user defined uniform values to avoid other unneeded glUniform calls. Like omitting redundant` glUniform` calls for a `color` uniform that does not change often, even between different pipelines changes.

floooh:
> Would be cool if sokol could optionally cache user defined uniform values to avoid other unneeded glUniform calls.

I've been thinking about this too a bit but I think that goes too far because uniform data can be fairly big (several KBytes), and just the test might be worse then just doing the glUniform call.

A better strategy to reduce the number of glUniform calls is flattening uniforms into a single vec4 array. The sokol-shdc tool does this automatically for all GLSL dialects (this is also used in "big engines": https://www.epicgames.com/fortnite/en-US/news/fortnite-on-android-launch-technical-blog, look under "RENDERING PERFORMANCE").

With this method, each sg_apply_uniforms() call always results in exactly one glUniform4fv() call.

floooh:
Ok I have merged my change (move glUniform1i() into _sg_gl_create_shader(), and will close this PR.

PR #312, opened 2020-06-04
added `#define NOMINMAX`, removed redundant `#include <windows.h>`

garettbass:
Just some minor cleanup.  Curious how you feel about `NOMINMAX`.  Seems important to me because without it, `<windows.h>` stomps any variant of `min()`, and `max()` with a macro.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3263,6 +3263,8 @@ _SOKOL_PRIVATE const _sapp_gl_fbconfig* _sapp_gl_choose_fbconfig(const _sapp_gl_
 #if defined(_WIN32)
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
+#endif
+#ifndef NOMINMAX
 #define NOMINMAX
 #endif
 #include <windows.h>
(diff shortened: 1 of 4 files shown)


garettbass:
I removed an `#include <windows.h>` directly before an `#include <d3d11.h>` because `d3d11.h` includes `windows.h` itself.

floooh:
Ok merged. The NOMINMAX makes sense (although I didn't think of it as highly important, because the windows header is only included in the implementation block, but this will of course still leak into other code in the same compilation unit).

I guess this was the reason why I defined my own _sg_min()/_sg_max() macros in sokol_gfx.h :)

floooh:
Btw, did you notice that something seems to be wrong with your git commit name? You're not showing up as contributor (for instance here: https://github.com/floooh/sokol/graphs/contributors), and when I'm looking at one of your commits (e.g. here: https://github.com/floooh/sokol/commits/master) only your name shows up but it's not clickable.

...just wanted to know if you're aware, because when I started merging your PRs I wondered if I did something wrong on my side (because I merged some PRs manually).

garettbass:
Thanks for pointing it out.  I don’t understand why it is broken.

floooh:
Maybe you haven't setup the user.email address in your local git config, or the email address is different from the one used for your github account? I don't know either how github associates the git commit information with the github account, but I guess the email address plays an important part :)

PR #314, opened 2020-06-05
soft fullscreen mode on macos

medvednikov:
As discussed.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3326,6 +3326,9 @@ _SOKOL_PRIVATE const _sapp_gl_fbconfig* _sapp_gl_choose_fbconfig(const _sapp_gl_
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
 #endif
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
 #include <windows.h>
 #include <windowsx.h>
 #include <shellapi.h>
@@ -3359,7 +3362,6 @@ _SOKOL_PRIVATE const _sapp_gl_fbconfig* _sapp_gl_choose_fbconfig(const _sapp_gl_
 #ifndef COBJMACROS
 #define COBJMACROS
 #endif
-#include <windows.h>
 #include <d3d11.h>
 #include <dxgi.h>
 #endif
--- a/sokol_audio.h
+++ b/sokol_audio.h
@@ -473,6 +473,9 @@ inline void saudio_setup(const saudio_desc& desc) { return saudio_setup(&desc); 
     #ifndef WIN32_LEAN_AND_MEAN
     #define WIN32_LEAN_AND_MEAN
     #endif
+    #ifndef NOMINMAX
+    #define NOMINMAX
+    #endif
     #include <windows.h>
     #include <synchapi.h>
     #pragma comment (lib, "kernel32.lib")
--- a/sokol_fetch.h
+++ b/sokol_fetch.h
@@ -1007,6 +1007,9 @@ inline sfetch_handle_t sfetch_send(const sfetch_request_t& request) { return sfe
     #ifndef WIN32_LEAN_AND_MEAN
     #define WIN32_LEAN_AND_MEAN
     #endif
+    #ifndef NOMINMAX
+    #define NOMINMAX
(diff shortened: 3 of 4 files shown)


floooh:
That' surprisingly little code :) 

Do you know what to do in the new sapp_toggle_fullscreen() function? Is it enough to change the style flags?

https://github.com/floooh/sokol/blob/26c6fb874b6daa6193194a9ea3fc7e34bf2e4bee/sokol_app.h#L1495-L1502

medvednikov:
From my limited experience with Cocoa, yes, this should work.

floooh:
I started testing this PR in a local branch and ran into some problems:

- macOS the menu bar is still visible 
- there's no way to quit the application currently (e.g. through the quit menu icon or shortcut - so at least a Quit menu item would need to be added, which is probably a good idea anyway)
- the "programmatic quit behaviour" in sokol_app.h is messed up (can be tested with the imgui-highdpi-sapp example), the reason seems to be that the ```windowShouldClose``` event handler is never called after calling ```performClose``` on the window (maybe because the ```NSWindowStyleMaskClosable``` flag isn't set

I'll try to find solutions to those problems when I get around it, but at the current state I don't want to merge the PR.

Btw, on macOS Catalina (10.15.6 Beta) I'm not seeing the into-fullscreen animation when starting the imgui-highdpi-sapp example, the animation only happens when toggling between windowed and fullscreen in the running application.

medvednikov:
> the menu bar is still visible
oops, sorry, I have the menu bar always hidden

> there's no way to quit the application currently
yeah for some reason macos requires a Quit menu for Cmd + Q to work

> I'm not seeing the into-fullscreen animation
do you see it when alt tabbing between different apps?

floooh:
> do you see it when alt tabbing between different apps?

Yes, after starting the application (when the fullscreen animation is skipped), all other interactions lead to the animation.

medvednikov:
Interesting. I get the initial animation as well on Catalina.

elimisteve:
Hi there! I would just like to encourage @floooh to tweak and merge this in! The V community, of which I am a part (and that @medvednikov started), is excited about Sokol and now uses it by default in our graphics library.

medvednikov:
@elimisteve the issues @floooh mentioned need to be resolved first.

With this PR merged at this stage, it will be impossible to quit full screen apps, the menu bar will be visible, etc.

elimisteve:
@medvednikov Yes, hence why I said "tweak and merge" and not merely "merge". Maybe more than tweaking is required, but _improving_ then merging would be great!

medvednikov:
We can't expect the maintainer to fix an unfinished PR for us immediately. These things take time, and this is clearly not on @floooh's priority list.

We can take care of all the issues ourselves, then it'll get merged right away.

medvednikov:
@elimisteve I use this patch in my fork of Sokol, so we are not in a hurry :)

floooh:
I'll have a look at how GLFW handles borderless fullscreen windows, I just played around with this and it creates a window that also covers the menu bar. However one downside is that one cannot tab away from a GLFW borderless fullscreen window on Mac, instead the window will always force itself to the front. But I think some variation of the GLFW fullscreen code might be useful.

ylluminate:
> That' surprisingly little code :)

Hahahahaha, this just made me chuckle. I think "little code" is somewhat of a religion for @medvednikov and thus we have beautiful V.  I'd love to eventually see Sokol c2v'd (at least after 0.3 release) and maintained in V. 😆

medvednikov:
Github closed this PR after force pushing, can you reopen it please and make it use the `fullscreen_macos` branch? Thanks

floooh:
> can you reopen it please and make it use the fullscreen_macos branch?

I'd love to but I don't know how :D The "Reopen" button is greyed out with the following note when trying to click it:

Screen Shot 2020-06-15 at 12 06 46 PM

lazalong:
Can this be reopened please?

IMO the difficulty of closing a borderless window is a non-issue : the programmer using this feature has the responsibility to provide a closure/switch option.  The main usage is to create more 'esthetic' application where the menu & min/max/close button are on the same bar. Instead of wasting top space.
top_bar

PR #316, opened 2020-06-07
sokol_gfx: cache current program in GL backend

edubart:
This cache redundant calls to `glUseProgram` when actually no program has changed in the GL state. This happens often in my use case because I execute the same shader often but in a separate pipeline with different blend mode, while the shader (and GL program) doesn't change. And from my benchmarks `glUseProgram` is expansive relative to other calls, look the table in this screenshot as reference, look it can take over ~10000ns while other calls is much less expensive like `glUniform`.

Notice `glUseProgram` is called three times with the same value, this commit omit this uneeded calls.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -2998,6 +2998,7 @@ typedef struct {
     GLuint index_buffer;
     GLuint stored_vertex_buffer;
     GLuint stored_index_buffer;
+    GLuint prog;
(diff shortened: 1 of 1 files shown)


floooh:
Ok merged, thanks! At first I thought the decision to write the current program into the state cache is an odd idea (because if sg_reset_state_cache() is called after calling GL functions directly it might "pollute" the cache with a program not owned by sokol-gfx), but since shader program GL names created from the same context can't collide that should be fine.

PR #317, opened 2020-06-07
sokol_gfx: call draw buffers once on framebuffer create for GL backend

edubart:
OpenGL core 3.3 spec says that the `glDrawBuffers` state is store per framebuffer object. So we can skip calls to this function on every `sg_begin_pass` by moving them to the `_sg_gl_create_pass`.

I've tested `mrt-glfw` sample with this and works fine.


floooh:
Ooooh that's a very nice fix (if it works as intended). I'll try to look into it in detail tomorrow (need to do some reading in the GL spec and some testing).

edubart:
https://www.khronos.org/opengl/wiki/Framebuffer

> The state set by glDrawBuffers is part of the state of the framebuffer. So you can generally set this up once and leave it set.

PR #319, opened 2020-06-11
include dpi scale 1.5 in high dpi

mattiasljungstrom:
Very minor tweak: A dpi of exactly 1.5 is fairly common and I think it should be considered a high dpi screen? Most 4k screens run at 150% in Win10 for instance.


floooh:
No objections :)

PR #321, opened 2020-06-14
fullscreen on linux

medvednikov:
right now it's marked as TODO

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -7226,6 +7226,23 @@ _SOKOL_PRIVATE void _sapp_x11_create_window(Visual* visual, int depth) {
         _sapp_fail("X11: Failed to create window");
     }
 
+    if (_sapp.desc.fullscreen) {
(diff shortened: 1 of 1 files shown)


floooh:
I'll look into this next.

floooh:
Minor nitpick: Tab characters ;) (there's an .editorconfig file checked in, but I guess the editor you're using ignores that).

But I'll fix that while merging. I'm doing the merge manually and try to add the toggle-fullscreen functionality while I'm at it.

floooh:
Ok, I changed things around a bit while implementing sapp_toggle_fullscreen(). The interesting stuff happens now in a function _sapp_x11_set_fullscreen(bool):

https://github.com/floooh/sokol/blob/c829f8990d1c9a9aabe70c489fb5a01ca08a9576/sokol_app.h#L7207-L7226

...and I tested on Ubuntu 20.04 with the imgui-highdpi-sapp sample (this starts in fullscreen and has a UI button to toggle between fullscreen and windowed).

Thanks for the PR (and if you encounter problems with my changes please let me know!)

PR #332, opened 2020-07-08
add temp solution to macos sapp_mouse_shown

slmjkdbtl:
Since `CGCursorIsVisible` is deprecated there seem to not be an API for querying mouse visibility state on macOS. This commit adds a quick bool to keep track of the state.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1329,6 +1329,7 @@ static NSOpenGLPixelFormat* _sapp_macos_glpixelformat_obj;
 static NSTimer* _sapp_macos_timer_obj;
 #endif
 static uint32_t _sapp_macos_flags_changed_store;
+static bool _sapp_macos_mouse_is_shown = true;
 
 _SOKOL_PRIVATE void _sapp_macos_init_keytable(void) {
     _sapp.keycodes[0x1D] = SAPP_KEYCODE_0;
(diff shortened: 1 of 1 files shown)


floooh:
Good fix. I'll work on integrating this now (since sokol_app.h has changed a lot I'll need to resolve the conflicts, but I'll take care of that). Next thing I'll work on is pointer-lock btw, which is somewhat related to show/hide mouse.

floooh:
Hmm ok, unfortunately the manual merge ended up being too much work after my big code-cleanup merge (where I shuffled a lot of code and data around), so I had to "reimplement" the feature, sorry about that (I mentioned you in the README though).

Another reason for not using your code directly was that I decided to change the behaviour of sapp_show_mouse(). So far this was simply forwarding the call to Win32's ShowCursor() or the macOS NSCursor class, which both implement a "hidden counter" (so hiding the cursor twice also requires showing it twice to actually make the cursor visible again). I have now changed that so that hide always hides and show always shows.

My changes are in this commit: d9692d33a18fb4d0cd1b70f6121ac4fb7986de1b

slmjkdbtl:
Thanks for the explanation! Yes `CGDisplayShowCursor` is a better solution than `NSCursor`, glad these got implemented.

PR #334, opened 2020-07-14
sokol_app.h: fix macos with glcore33 to update dimensions / dpi on startup

slmjkdbtl:
macOS with GLCORE33 doesn't call `_sapp_macos_update_dimensions` on startup like with METAL does, which keeps DPI at default value until a resize event comes in.

Not sure if it's the best place to put this function tho I'm not familiar with `sokol_app.h` codebase

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1595,6 +1595,7 @@ _SOKOL_PRIVATE void _sapp_macos_toggle_fullscreen(void) {
             userInfo:nil
             repeats:YES];
         [[NSRunLoop currentRunLoop] addTimer:_sapp_macos_timer_obj forMode:NSDefaultRunLoopMode];
+        _sapp_macos_update_dimensions();
     #endif
     _sapp.valid = true;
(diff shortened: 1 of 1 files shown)


slmjkdbtl:
oh I'm sorry didn't see this is already fixed in the master branch, nvm

floooh:
Hi, yes sorry, this change slipped in as part of the big code cleanup merge I did yesterday. Also, I haven't looked at your other PRs yet, but I'll try to work through them in the next few days.

Cheers!

slmjkdbtl:
Thanks for the reply! Love the new updates and fixes! It also fixes metal not being able to startup on my macOS device.

floooh:
>  It also fixes metal not being able to startup on my macOS device.

Oh... interesting. Do you know what sort of error you got before?

slmjkdbtl:
@floooh Sorry I failed to reproduce, there might be a problem with my setup the time I tried.

PR #337, opened 2020-07-19
Fix sokol_fetch.h spinning on Windows

bqqbarbhg:
Instead of waiting on the event as it's supposed to `_sfetch_thread_dequeue_incoming()` spins while waiting as the address of the event handle is passed to `WaitForSingleObject()` instead of the handle itself causing the function fail immediately as the handle is invalid.


floooh:
Whoops, good catch, thanks! I need to figure out why this didn't trigger a compiler warning (I would expect that at least compiling as C++ would give some sort of warning, and sokol_fetch.h is already included in the cpp-compile-test.cc

PR #343, opened 2020-07-25
Adding sokol_audio UWP support, no MMDevice initialization

albertofustinoni:
While Windows UWP supports WASAPI/Core Audio, it does not support the IMMDevice* API from `mmdeviceapi.h`, which `sokol_audio` uses to get an `IAudioClient`.

There is, however, an alternative way of getting an `IAudioClient` interface, and from there use the same codepath as desktop Win32, by using [ActivateAudioInterfaceAsync](https://docs.microsoft.com/en-us/windows/win32/api/mmdeviceapi/nf-mmdeviceapi-activateaudiointerfaceasync).

This method is possible for desktop apps too, starting from Windows 8.

I have modified `sokol_audio` to:
- Obtain `IAudioClient` via `ActivateAudioInterfaceAsync` instead of IMMDevice if `SOKOL_WIN32_NO_MMDEVICE` is defined
- Fix import libs, define `SOKOL_WIN32_NO_MMDEVICE` when compiling the header for an UWP app. 

While still being C only.

Tested:
- On a desktop app with `SOKOL_WIN32_NO_MMDEVICE` defined everything seems to work
- Compiling for an UWP dll works
- Compiling both from a C and a C++ source

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Simple
 [STB-style](https://github.com/nothings/stb/blob/master/docs/stb_howto.txt)
 cross-platform libraries for C and C++, written in C.
 
-[See what's new](#updates) (**22-Sep-2020** sokol_app.h: new UWP backend)
+[See what's new](#updates) (**26-Sep-2020** sokol_app.h: new function to change window title)
 
 [Live Samples](https://floooh.github.io/sokol-html5/index.html) via WASM.
 
@@ -431,6 +431,11 @@ See the sokol_args.h header for a more complete documentation, and the [Tiny
 Emulators](https://floooh.github.io/tiny8bit/) for more interesting usage examples.
 
 # Updates
+
+- **26-Sep-2020**:
+    sokol_app.h gained a new function sapp_set_window_title() to change
+    the window title on Windows, macOS and Linux. Many thanks to
+    @medvednikov for the initial PR!
 
 - **23-Sep-2020**:
     sokol_app.h now has initial UWP support using the C++/WinRT set of APIs.
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1026,6 +1026,8 @@ SOKOL_API_DECL uint64_t sapp_frame_count(void);
 SOKOL_API_DECL void sapp_set_clipboard_string(const char* str);
 /* read string from clipboard (usually during SAPP_EVENTTYPE_CLIPBOARD_PASTED) */
 SOKOL_API_DECL const char* sapp_get_clipboard_string(void);
+/* set the window title (only on desktop platforms) */
+SOKOL_API_DECL void sapp_set_window_title(const char* str);
 
 /* special run-function for SOKOL_NO_ENTRY (in standard mode this is an empty stub) */
 SOKOL_API_DECL int sapp_run(const sapp_desc* desc);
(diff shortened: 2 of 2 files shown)


floooh:
Looks like a good fix, thanks :) Do you have any advice on how to best test this?

Ideally I'd like to add a small UWP-specific set of samples to the sokol-samples repository, similar to the D3D11- and Metal-specific samples. Do you have some simple code around to create an UWP windows and D3D11 device and swapchain, similar to this: https://github.com/floooh/sokol-samples/blob/master/d3d11/d3d11entry.c ?)

albertofustinoni:
Glad to hear you like the code :)

As a matter of fact, the next thing I need to do as a part of a larger project I'm working on is making an UWP implementation of sokol_app (so far I've been testing on desktop Windows by #defining `SOKOL_WIN32_NO_MMDEVICE` manually, in the c file that has `SOKOL_IMPL`)

Question becomes: are you interested in an UWP implementation of sokol_app? My plan is to use one of the starting Visual Studio Templates (either DirectX 11 and XAML app or DirectX 11 app), getting rid of any rendering code and hook up sokol_app calls from there.

Otherwise the best recommendation I have is to use one of the Visual Studio UWP DirectX 11 templates and hook up sokol_gfx to the context/device they create.

floooh:
> Question becomes: are you interested in an UWP implementation of sokol_app?

Even though I'm not a big fan of UWP (see https://floooh.github.io/2016/08/20/uwp.html) I think it would make sense for sokol_app.h at least if it fits into the sokol_app.h API model.

Would UWP support in sokol_app.h plus sokol_gfx.h's D3D11 backend allow hobbyists to run their code on the Xbox One? If yes then at least I could muster up some motivation for adding an UWP backend ;)

albertofustinoni:
I know you don't have a very high opinion of UWP (I think I stumbled on the post you linked a few moths back), but at least I can confirm [it allow hobbyist developers to run their code on the Xbox One](https://github.com/Aftnet/RetriX/wiki/Sideloading-on-Xbox-One) via developer mode.

As the crazy person who made a [Libretro frontend](https://www.retrix.me/) and ported a bunch of cores to UWP (so I think I understand the pain from a dev perspective), allow me to say this in defense of the platform: it does right by the user.

With a UWP app you no longer have to play Russian roulette with your system: apps are sandboxed and you get a list of required permissions at installation time + consent dialogs in most cases when accessing most sensitive things (like camera or microphone). No more [apps sneakily adding kernel code](https://www.pcgamesn.com/doom-eternal/denuvo-anti-cheat) is nice too.

albertofustinoni:
One more thing: I think a good starting point for making an UWP implementation of sokol_app is [here](https://github.com/microsoft/Xbox-ATG-Samples/tree/master/UWPSamples/IntroGraphics/SimpleTriangleCppWinRT_UWP) or you can use one of the Visual Studio templates for DirectX11 UWP apps.

While the link is a sample for a bigger toolkit, the hooks are minimal (this sample renders a triangle to screen and uses the toolkit for mouse/pad support) and it is made using pure C++ instead of C++/CX.

Both the sample and the VS templates handle management of the swapchain in response to window resizing and rotation.

albertofustinoni:
@floooh I have gotten the basics of an UWP implementation of `sokol_app` [in this repo](https://github.com/Aftnet/sokol_app_uwp) (master branch).

Right now it only handles the basics of creating a window and event pump/render loop, setting up the Direct3D rendering context and interacting with the sokol_app render/init/cleanup callbacks.

What do you think so far?

floooh:
Is this no longer good for merging? It was one of the next things I wanted to merge now that the UWP sokol_app.h stuff had been merged :D

floooh:
...ah I see, CoInitializeEx was changed independently in sokol_audio.h to ignore the return code, is this all that's needed for making it work in UWP?

albertofustinoni:
Nono, it’s still good. I just made a mistake yesterday when creating the other PR it seems.

albertofustinoni:
This got closed since I made the mistake of force pushing your repo's master to my fork's master... forgetting this PR was from my fork's master.

I do not see a way of reopening this PR so I created #394 with the same code as the one I'm currently using.

Sorry for the trouble.

PR #344, opened 2020-07-27
sokol_audio: Don't include platform headers if SOKOL_DUMMY_BACKEND is defined

bqqbarbhg:
Currently using `SOKOL_DUMMY_BACKEND` still includes platform-specific headers. For example I couldn't build the dummy backend on a Linux server without ALSA libraries installed. This PR adds more `#if defined(SOKOL_DUMMY_BACKEND)` checks to skip including headers and defines a dummy mutex as well.

Empty structs have a dummy variable as some MSVC versions can't handle empty structs when compiling in C mode. I also added a dummy variable to the existing empty `_saudio_mutex_t` struct for that reason.


floooh:
Ok, merged. Thanks!

PR #348, opened 2020-07-29
sokol_app.h: initial mouse-lock implementation

floooh:
For getting unrestricted mouse-movement information.

Two new functions:

```c
void sapp_lock_mouse(bool b);
bool sapp_mouse_locked(void);
```

When mouse-lock is active, the mouse pointer is hidden, the mouse position reported in SAPP_EVENTTYPE_MOUSE_MOVE appears frozen, and the new sapp_event items mouse_dx/dy contain relative mouse movement.

It is not guaranteed that mouse movement has the same "trajectory" in lock mode (because on some platforms, the "raw mouse movement" APIs are used). In locked mode, MOUSE_MOVE events will also be generated when the mouse leaves the window area or hits the screen border.

If mouse-lock is inactive, mouse_dx/dy will be computed by subtracting the new mouse position from the previous. 

Implementation status:

- [x] Win32
- [x] macOS
- [x] Linux
- [x] WASM

...this PR also contains the missing sapp_show_mouse() implementation for Linux

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1216,7 +1216,6 @@ typedef struct {
     _sapp_macos_app_delegate* app_dlg;
     _sapp_macos_window_delegate* win_dlg;
     _sapp_macos_view* view;
-    float mouse_locked_x, mouse_locked_y; // 'raw' frozen mouse position during mouse lock
     #if defined(SOKOL_METAL)
         id<MTLDevice> mtl_device;
     #endif
@@ -2425,32 +2424,8 @@ _SOKOL_PRIVATE const char* _sapp_macos_get_clipboard_string(void) {
     return _sapp.clipboard.buffer;
 }
 
-_SOKOL_PRIVATE void _sapp_macos_center_mouse(void) {
-    const NSRect r_screen = [_sapp.macos.window convertRectToScreen:[_sapp.macos.view frame]];
-    const CGFloat x = NSMidX(r_screen);
-    const CGFloat y = CGDisplayBounds(kCGDirectMainDisplay).size.height - NSMidY(r_screen);
-    CGWarpMouseCursorPosition(CGPointMake(x, y));
-}
-
-_SOKOL_PRIVATE void _sapp_macos_store_locked_mouse_pos(void) {
-    const NSPoint locked_pos = [_sapp.macos.window mouseLocationOutsideOfEventStream];
-    _sapp.macos.mouse_locked_x = locked_pos.x;
-    _sapp.macos.mouse_locked_y = locked_pos.y;
-}
-
-_SOKOL_PRIVATE void _sapp_macos_restore_locked_mouse_pos(void) {
-    const NSRect r_win = NSMakeRect(_sapp.macos.mouse_locked_x, _sapp.macos.mouse_locked_y, 0, 0);
-    const NSRect r_screen = [_sapp.macos.window convertRectToScreen:r_win];
-    const CGFloat x = NSMinX(r_screen);
-    const CGFloat y = CGDisplayBounds(kCGDirectMainDisplay).size.height - NSMinY(r_screen);
-    CGWarpMouseCursorPosition(CGPointMake(x, y));
-}
-
 _SOKOL_PRIVATE void _sapp_macos_update_mouse(void) {
-    if (_sapp.mouse.locked) {
-        _sapp_macos_center_mouse();
-    }
-    else {
+    if (!_sapp.mouse.locked) {
         const NSPoint mouse_pos = [_sapp.macos.window mouseLocationOutsideOfEventStream];
         float new_x = mouse_pos.x * _sapp.dpi_scale;
         float new_y = _sapp.framebuffer_height - (mouse_pos.y * _sapp.dpi_scale) - 1;
@@ -2484,11 +2459,8 @@ _SOKOL_PRIVATE void _sapp_macos_lock_mouse(bool lock) {
     if (_sapp.mouse.locked) {
         [NSEvent setMouseCoalescingEnabled:NO];
         CGAssociateMouseAndMouseCursorPosition(NO);
-        _sapp_macos_store_locked_mouse_pos();
-        _sapp_macos_center_mouse();
     }
     else {
-        _sapp_macos_restore_locked_mouse_pos();
         CGAssociateMouseAndMouseCursorPosition(YES);
         [NSEvent setMouseCoalescingEnabled:YES];
     }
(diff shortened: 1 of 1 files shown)


meshula:
I've been playing with this branch - it looks like it will do the job for me. There is however, a basic mismatch versus ImGui though. In ImGui, the sapp events are submerged and translated into the ImGuiIO object by sokol_imgui.h, ImGuiIO hasn't got a spot for relative mouse motion. I played with some different approaches. 

- Using the delta values to synthesize a MousePos doesn't work, because the relative mouse motion from a trackpad on Windows has a lot of "interesting" heuristics that make the result very unintuitive to a user. 
- Another approach is to add a MouseDelta to the ImGuiIO object, and possibly a flag in the ImGuiIO object, or have client code query sapp for mouse lock. 
- Another approach is to cache the last mouse delta from the event in sapp itself, so that client apps can fetch it from there in preference to modifying ImGui. 
- Another approach, is to cache the mouse motion in ImGuiIO's MousePos, and have client code ask if sapp_mouse_lock is active.

The last seems the least intrusive since it doesn't involve a local hack to ImGui, or sapp, just a modification to sokol_imgui.h, but does mean that if the last "actual" mouse position is important, it'll be lost without extra caching in client code.

I was wondering if you'd given any thought to a best practice versus ImGui?

floooh:
I see. Hmm... the mouse-lock feature is more useful for things like camera control or first-person-shooter cameras, less for dragging in UIs... for those it's usually also better to report raw mouse movement with as much precision as possible, but which doesn't correspond with "cooked" high-level mouse movement.

Is the main problem in your case that dragging is essentially cancelled when the mouse cursor leaves the window area (e.g. for ImGui sliders and color-widegts)?

Would a separate "mouse capture mode" work which would work similar to Win32 mouse capture? E.g. the only difference to regular mouse events would be that they are reported outside the window client area, but it would have restrictions like that it only works while the mouse button is pressed down? (e.g. see here: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setcapture).

This mouse-capture mode would make sense for UI manipulation, but less for camera controls.

It probably could even be activated/deactivated automatically (any mouse button pressed would activate capture and release would deactivate it).

Anyway, I'll merge the mouse-lock stuff back to master now. But I would appreciate your feedback on whether a separate mouse capture feature for UI manipulation sounds like a good idea :)

meshula:
I have both use cases. 

I have sliders on the edge of my window, and it interferes with usability that the dragging is canceled at the edge of the window area. In this case, a visible cursor is valuable for detailed control of the value.

I also have button initiated-and-held camera controls, where I am synthesizing relative motion from ImGui's cursor position. Camera motion initiated by a mouse drag near the edge of the screen also suffers from the edge cancel. In this case, a visible cursor is not valuable, and hiding would eliminate the visual distraction of the nearly meaningless cursor position.

I don't have a first person shooter mode, where the mouse lock activation is independent of button action, and where the button may have another function such as target selection. Clearly other applications would want that.

This morning I started to play with exposing the mouse deltas via sokol_imgui itself, thus avoiding modifying sapp or imgui, which I prefer to simply pull and use. That's basically equivalent to the idea of a separate mouse-capture mode, the separate mode is probably the cleanest.

floooh:
> mouse drag near the edge of the screen also suffers from the edge cancel

Hmm... this isn't solveable by the regular Win32 SetCapture function is it (not 100% sure though)? I think getting mouse movement past the screen border would require the tricks used in the mouse-lock code (like using a raw-input API, and/or warping the mouse position back to a 'safe position' away from any screen borders).

meshula:
On Windows, this patch solves the "edge of screen" problem, I believe you suggested something similar in another thread a while ago. I haven't visited the other platforms yet.

```diff
index 97ac980..b453e94
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -4752,6 +4752,9 @@ _SOKOL_PRIVATE void _sapp_win32_app_event(sapp_event_type type) {

 _SOKOL_PRIVATE LRESULT CALLBACK _sapp_win32_wndproc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam) {
     /* FIXME: refresh rendering during resize with a WM_TIMER event */
+
+    static uint8_t mouse_buttons = 0;
+
     if (!_sapp_win32_in_create_window) {
         switch (uMsg) {
             case WM_CLOSE:
@@ -4811,21 +4814,33 @@ _SOKOL_PRIVATE LRESULT CALLBACK _sapp_win32_wndproc(HWND hWnd, UINT uMsg, WPARAM
                 break;
             case WM_LBUTTONDOWN:
                 _sapp_win32_mouse_event(SAPP_EVENTTYPE_MOUSE_DOWN, SAPP_MOUSEBUTTON_LEFT);
+                if (!mouse_buttons && ::GetCapture() == NULL) ::SetCapture(hWnd);
+                mouse_buttons |= 1;
                 break;
             case WM_RBUTTONDOWN:
                 _sapp_win32_mouse_event(SAPP_EVENTTYPE_MOUSE_DOWN, SAPP_MOUSEBUTTON_RIGHT);
+                if (!mouse_buttons && ::GetCapture() == NULL) ::SetCapture(hWnd);
+                mouse_buttons |= 2;
                 break;
             case WM_MBUTTONDOWN:
                 _sapp_win32_mouse_event(SAPP_EVENTTYPE_MOUSE_DOWN, SAPP_MOUSEBUTTON_MIDDLE);
+                if (!mouse_buttons && ::GetCapture() == NULL) ::SetCapture(hWnd);
+                mouse_buttons |= 4;
                 break;
             case WM_LBUTTONUP:
                 _sapp_win32_mouse_event(SAPP_EVENTTYPE_MOUSE_UP, SAPP_MOUSEBUTTON_LEFT);
+                if (!mouse_buttons && ::GetCapture() == hWnd) ::ReleaseCapture();
+                mouse_buttons &= ~1;
                 break;
             case WM_RBUTTONUP:
                 _sapp_win32_mouse_event(SAPP_EVENTTYPE_MOUSE_UP, SAPP_MOUSEBUTTON_RIGHT);
+                if (!mouse_buttons && ::GetCapture() == hWnd) ::ReleaseCapture();
+                mouse_buttons &= ~2;
                 break;
             case WM_MBUTTONUP:
                 _sapp_win32_mouse_event(SAPP_EVENTTYPE_MOUSE_UP, SAPP_MOUSEBUTTON_MIDDLE);
+                if (!mouse_buttons && ::GetCapture() == hWnd) ::ReleaseCapture();
+                mouse_buttons &= ~4;
                 break;
             case WM_MOUSEMOVE:
                 _sapp.mouse_x = (float)GET_X_LPARAM(lParam) * _sapp_win32_mouse_scale;
````

Issue #345, opened 2020-07-28
sokol_app.h: drag'n'drop discussion


floooh:
Issue for discussing details of file drag'n'drop in sokol_app.h.

General implementation should be:

- a new file-dropped event type to notify the application that a file had been dropped on the window (Q: can this include coordinates where exactly the file was dropped?)
- a separate function to get the file path or URL of the dropped file (since this might be pretty long it shouldn't be embedded in the event, but instead live in a separate internal buffer like the clipboard content).

It's up to the application what to do with the file path.

floooh:
For the web platform I have implemented very basic drag'n'drop support in my emulators completely on the Javascript side (just as example how the whole thing works in JS):

https://github.com/floooh/chips-test/blob/df11a99d4ab131fa90aef644effb9d1a64cdd591/webpage/emsc.html#L68-L90

hb3p8:
I've extracted our changes to https://github.com/floooh/sokol/pull/346
We'll probably need iterate on it to unify with implementations for mac and linux. Though having simple file name pointers in event seem kinda fine to me?

floooh:
> Though having simple file name pointers in event seem kinda fine to me?

It probably would since the internal buffer would never move elsewhere, but I'd prefer not to have it somewhat similar to the clipboard access (where you'd receive a pasted event and then call sapp_get_clipboard_string(). Since all event types  share the same event structure, I'd like to not put too much stuff in there for "exotic" events :)

prime31:
I’ll finish up macOS today and remove my clipboard hijacking antics ;)

Since this is a short lived event is the plan with the buffer to do something like this or a different strategy:
- os drop event occurs
- dupe the path to a buffer
- fire the sapp dropped event
- (user is expected to fetch the file path here)
- free throws buffer

prime31:
Disregard previous comment! I see that the wip PR already does exactly what I described.

@floooh can you merge that PR into a wip-file-drop branch so that I can put my PR on top of it?

prime31:
I pulled in the changes from @hb3p8 for Windows and made the macOS version follow the same paradigm. This is all in PR #347.

@floooh: this currently passes along the paths and count along with the event. Should we separate out the file path data and add a `const char** sapp_get_dropped_files(void)` method and just leave the `drop_files_count` in the event?

prime31:
[...]

PR #349, opened 2020-07-29
Simple drag-n-drop support on macOS and Windows

prime31:
Draft with new API per #345. This is tested working on macos and needs to be tested on Windows next.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -654,6 +654,7 @@ typedef enum sapp_event_type {
     SAPP_EVENTTYPE_UPDATE_CURSOR,
     SAPP_EVENTTYPE_QUIT_REQUESTED,
     SAPP_EVENTTYPE_CLIPBOARD_PASTED,
+    SAPP_EVENTTYPE_FILE_DROPPED,
     _SAPP_EVENTTYPE_NUM,
     _SAPP_EVENTTYPE_FORCE_U32 = 0x7FFFFFFF
 } sapp_event_type;
@@ -822,6 +823,7 @@ typedef struct sapp_event {
     int window_height;
     int framebuffer_width;
     int framebuffer_height;
+    int drop_files_count;
 } sapp_event;
 
 typedef struct sapp_desc {
@@ -908,6 +910,8 @@ SOKOL_API_DECL uint64_t sapp_frame_count(void);
 SOKOL_API_DECL void sapp_set_clipboard_string(const char* str);
 /* read string from clipboard (usually during SAPP_EVENTTYPE_CLIPBOARD_PASTED) */
 SOKOL_API_DECL const char* sapp_get_clipboard_string(void);
+/* read paths from dropped files during SAPP_EVENTTYPE_FILE_DROPPED. The paths are only valid during the event callback! */
+SOKOL_API_DECL const char** sapp_get_dropped_files(void);
 
 /* special run-function for SOKOL_NO_ENTRY (in standard mode this is an empty stub) */
 SOKOL_API_DECL int sapp_run(const sapp_desc* desc);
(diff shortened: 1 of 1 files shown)


floooh:
Just a reminder for myself (I can easily fix this during merge):

The calls to calloc() and free() should be replaced with SOKOL_CALLOC and SOKOL_FREE macros (one day I actually want to get rid of SOKOL_CALLOC, and only have SOKOL_MALLOC, but that's for another time).

sokol_app.h
```diff
@@ -4848,6 +4916,34 @@ _SOKOL_PRIVATE LRESULT CALLBACK _sapp_win32_wndproc(HWND hWnd, UINT uMsg, WPARAM
...
+                for (i = 0;  i < _sapp.num_dropped_files;  i++) {
+                    const UINT length = DragQueryFileW(drop, i, NULL, 0) + 1;
+                    WCHAR* buffer = (WCHAR*)SOKOL_CALLOC(length, sizeof(WCHAR));
+
+                    DragQueryFileW(drop, i, buffer, length);
+
+                    const int dst_needed = WideCharToMultiByte(CP_UTF8, 0, buffer, -1, 0, 0, NULL, NULL);
+                    int len = dst_needed > _sapp.max_dropped_file_path_length ? _sapp.max_dropped_file_path_length : dst_needed;
+                    _sapp_win32_wide_to_utf8(buffer, &_sapp.dropped_files[i * _sapp.max_dropped_file_path_length], len);
+                    free(buffer);
```
floooh, line 4937: ...this is the free() call that slipped through...

floooh:
Ok I have started merging in branch https://github.com/floooh/sokol/tree/prime31-file-drop, apart from resolving conflicts (already done) I'll also do some changes related to the intenal structure cleanup that happened on the master branch in the meantime, and some "personal taste" things :)

I'm not sure yet whether I'll add the emscripten code too before merging back to master or doing that separately.

floooh:
Just found a long-standing bug, _sapp_win32_wide_to_utf8() wouldn't zero terminate the result string if it doesn't fit into the buffer. 

This was only used in the clipboard code before.

I'm fixing this now so that no conversion takes place if the string wouldn't fit the buffer (because simply putting a zero byte into the last buffer position might result in a corrupted UTF-8 string if this is in the middle of a multi-byte sequence).

prime31:
> Just found a long-standing bug, _sapp_win32_wide_to_utf8() wouldn't zero terminate the result string if it doesn't fit into the buffer.

Good find! I’m not sure how I missed that when looking through that method.

floooh:
> Good find! I’m not sure how I missed that when looking through that method.

TBH from the documentation of WideCharToMultiByte() I was assuming that the function would zero-terminate the string if it needs to be clamped. But it was easy to test by setting a very small max_dropped_file_path_length.

floooh:
FYI I've done some minor public API changes in my merge-branch:

- renamed the event to ```SAPP_EVENTTYPE_FILES_DROPPED``` (file => files)
- add a ```bool enable_dragndrop``` item to sapp_desc, similar to ```enable_clipboard```, by default this is false, and the whole drag'n'drop system is deactivated (e.g. no allocations taking place)
- bump the default path length to 2048 bytes (a few more KBytes doesn't matter much, and this should be a pretty safe max path length on all systems)

I'm also adding demo code to the events-sapp sample:

image

image

The mouse deltas don't make much sense, because at least on windows, no regular mouse movement is reported during a drag'n'drop operation, but I guess that's fine (the only important info is the mouse position)

floooh:
Ok, I have also added Linux drag'n'drop (taken from GLFW). Some more testing and updating the documentation and then I'll merge this PR back to master.

I haven't implemented anything for HTML5 yet, I'll do that in a separate branch.

floooh:
Ok, this has been merged.

Initial documentation (no WASM/HTML5 yet):

https://github.com/floooh/sokol/blob/03dec6b442ba8111324b879bd3ed4c3241109265/sokol_app.h#L502-L577

prime31:
Woohoo! Can finally get off the fork and back to the mainline. Thanks @floooh!

PR #350, opened 2020-08-01
sokol_audio.h: Implement WebAudio worklet backend

bqqbarbhg:
Hi! Turns out my game/engine is doing a lot of nasty blocking stuff on the main thread so I wanted to lift off the audio to a separate thread (especially now that Firefox supports it, albeit with headers). Anyways here's my stab at doing a WebAudio worklet implementation, as always feel free to use any parts of this for your own implementation if you'd prefer!

This PR enables audio worklets if `__EMSCRIPTEN_PTHREADS__` is defined and the browser supports them, otherwise it will fall back to the `ScriptProcessorNode` backend. I tried to test with a minimal repro case that both push/pull API work with worklets and that the fallback works as intended (https://github.com/bqqbarbhg/sokol-audio-worklet-test).

So turns out that the worklets are kind of a pain, getting WASM code to run in the worklet callback is super tricky, maybe even impossible (without a binary WASM blob I guess) in a single header context. I implemented a small JavaScript worker that communicates with a background thread through a single atomic variable and a single producer single consumer queue in lockstep.

#### Things to consider:

The worklets operate on 128 frame "quantum" batches, my first implementation called the user callback at this rate but I didn't like that it differed from the other backends that much so now it rounds the requested buffer size to a suitable value and calls the user callback at that rate. On the other hand operating at the worklet quantum size would make internal code and worklet synchronization a bit simpler.

The internal worklet queue is double the size of the requested buffer size so that it can double buffer with the worklet. You could increase this size independently from the buffer size to have higher latency (safer) without having to grow the user callback buffer size. Not sure if it's worth it though.

The embedded worklet source code is done with ES6 backtick strings which should be supported pretty much everywhere WebAudio is. I don't know if you can catch the syntax error on other browsers for a fallback. The most library-user-friendly approach would be to use `"line1" + "line2"` or a pre-condensed string but that's a lot less nice to edit.

### DIFF
--- a/sokol_audio.h
+++ b/sokol_audio.h
@@ -275,16 +275,12 @@
 
     THE WEBAUDIO BACKEND
     ====================
-    The WebAudio backend is currently using a ScriptProcessorNode callback to
-    feed the sample data into WebAudio. ScriptProcessorNode has been
-    deprecated for a while because it is running from the main thread, with
-    the default initialization parameters it works 'pretty well' though.
-    Ultimately Sokol Audio will use Audio Worklets, but this requires a few
-    more things to fall into place (Audio Worklets implemented everywhere,
-    SharedArrayBuffers enabled again, and I need to figure out a 'low-cost'
-    solution in terms of implementation effort, since Audio Worklets are
-    a lot more complex than ScriptProcessorNode if the audio data needs to come
-    from the main thread).
+    The WebAudio backend can use either Audio Worklets or a ScriptProcessorNode.
+    If __EMSCRIPTEN_PTHREADS__ is defined and the browser supports them the
+    Audio Worklet backend is used. Using the Worklet backend Sokol Audio will
+    create a background thread to feed the Worklet. If Worklets or threads are
+    not available the ScriptProcessorNode backend will be updated from the
+    main thread.
 
     The WebAudio backend is automatically selected when compiling for
     emscripten (__EMSCRIPTEN__ define exists).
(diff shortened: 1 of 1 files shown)


bqqbarbhg:
Alternatively you could do an implementation where instead of using `SharedMemoryBuffer` to communicate you could use an internal ring buffer in the worklet and use `postMessage()` (with `workletNode.port.onmessage` callback on the main thread) to request more frames to the back of the buffer and `workletNode.port.postMessage()` the frame as a response from the main thread if there are no threads. Unfortunately I don't think I have time to experiment with this. You would probably want to keep the shared memory in the `__EMSCRIPTEN_PTHREADS__` case anyway since the `postMessage()` introduces a lot of copying and maybe some latency as well.

floooh:
> Alternatively you could do an implementation where instead of using SharedMemoryBuffer...

Yeah that what I was thinking too after seeing that CORS-header requirement in Firefox. But I wonder if that's 'low-latency' enough to be better than ScriptProcessorNode.

When I've been thinking about Audio Worklet support I came to the same conclusion that running WASM on the audio thread side would be a PITA, so I though it might be better to implement all the audio-thread code as a Javascript function using the "inline worker" technique mentioned in this thread (essentially through a data URL):

https://groups.google.com/forum/#!searchin/emscripten-discuss/audio$20worklet%7Csort:date/emscripten-discuss/AwdVJGcRj7c/iBKqnuEiAgAJ

The linked article also mentions several other techniques to reduce latency for getting data into web worker threads.

PS: it will take a while until I can have a deeper look at the PR, I have a little backlog to work of other PRs and issues to work through first, and I also currently have a week of vacation in front of me where I will probably tinker around with Zig a bit ;)

bqqbarbhg:
This PR actually uses exactly the technique you mentioned with JS a inline worker (worklet?) and blob/data URLs! But yea I have no hurry for this PR as I'm using forked headers at the moment anyway, just thought to contribute it back if you're interested :D

bqqbarbhg:
I actually had to disable this myself from production as Chrome had some weird crackling issues, might have just been the API not working properly but more likely there's something broken in this PR..

floooh:
Ok good to know. TBH I'd like to find a solution that doesn't require WASM pthreads support (because of that SharedArrayBuffer response header requirement which requires control over the web server).

PR #351, opened 2020-08-01
sokol_app UWP implementation

albertofustinoni:
Implemented `sokol_app` for UWP.

Added supported features in table: basically close to parity to desktop Win32 (clipboard and cursor capture/replacement only missing). Hopefully this can do for a start :)

Tested and working on my PC and Xbox One via project [here](https://github.com/Aftnet/sokol_app_uwp)

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -81,42 +81,42 @@
 
     FEATURE/PLATFORM MATRIX
     =======================
-                        | Windows | macOS | Linux |  iOS  | Android | Raspi | HTML5
-    --------------------+---------+-------+-------+-------+---------+-------+-------
-    gl 3.x              | YES     | YES   | YES   | ---   | ---     | ---   | ---
-    gles2/webgl         | ---     | ---   | ---   | YES   | YES     | TODO  | YES
-    gles3/webgl2        | ---     | ---   | ---   | YES   | YES     | ---   | YES
-    metal               | ---     | YES   | ---   | YES   | ---     | ---   | ---
-    d3d11               | YES     | ---   | ---   | ---   | ---     | ---   | ---
-    KEY_DOWN            | YES     | YES   | YES   | SOME  | TODO    | TODO  | YES
-    KEY_UP              | YES     | YES   | YES   | SOME  | TODO    | TODO  | YES
-    CHAR                | YES     | YES   | YES   | YES   | TODO    | TODO  | YES
-    MOUSE_DOWN          | YES     | YES   | YES   | ---   | ---     | TODO  | YES
-    MOUSE_UP            | YES     | YES   | YES   | ---   | ---     | TODO  | YES
-    MOUSE_SCROLL        | YES     | YES   | YES   | ---   | ---     | TODO  | YES
-    MOUSE_MOVE          | YES     | YES   | YES   | ---   | ---     | TODO  | YES
-    MOUSE_ENTER         | YES     | YES   | YES   | ---   | ---     | TODO  | YES
-    MOUSE_LEAVE         | YES     | YES   | YES   | ---   | ---     | TODO  | YES
-    TOUCHES_BEGAN       | ---     | ---   | ---   | YES   | YES     | ---   | YES
-    TOUCHES_MOVED       | ---     | ---   | ---   | YES   | YES     | ---   | YES
-    TOUCHES_ENDED       | ---     | ---   | ---   | YES   | YES     | ---   | YES
-    TOUCHES_CANCELLED   | ---     | ---   | ---   | YES   | YES     | ---   | YES
-    RESIZED             | YES     | YES   | YES   | YES   | YES     | ---   | YES
-    ICONIFIED           | YES     | YES   | YES   | ---   | ---     | ---   | ---
-    RESTORED            | YES     | YES   | YES   | ---   | ---     | ---   | ---
-    SUSPENDED           | ---     | ---   | ---   | YES   | YES     | ---   | TODO
-    RESUMED             | ---     | ---   | ---   | YES   | YES     | ---   | TODO
-    QUIT_REQUESTED      | YES     | YES   | YES   | ---   | ---     | TODO  | YES
-    UPDATE_CURSOR       | YES     | YES   | TODO  | ---   | ---     | ---   | TODO
-    IME                 | TODO    | TODO? | TODO  | ???   | TODO    | ???   | ???
-    key repeat flag     | YES     | YES   | YES   | ---   | ---     | TODO  | YES
-    windowed            | YES     | YES   | YES   | ---   | ---     | TODO  | YES
-    fullscreen          | YES     | YES   | YES   | YES   | YES     | TODO  | ---
-    pointer lock        | TODO    | TODO  | TODO  | ---   | ---     | TODO  | TODO
-    screen keyboard     | ---     | ---   | ---   | YES   | TODO    | ---   | YES
-    swap interval       | YES     | YES   | YES   | YES   | TODO    | TODO  | YES
-    high-dpi            | YES     | YES   | TODO  | YES   | YES     | TODO  | YES
-    clipboard           | YES     | YES   | TODO  | ---   | ---     | ---   | YES
+                        | Windows | macOS | Linux |  iOS  | Android | UWP  | Raspi | HTML5
+    --------------------+---------+-------+-------+-------+---------+------+-------+-------
+    gl 3.x              | YES     | YES   | YES   | ---   | ---     | ---  | ---   | ---
+    gles2/webgl         | ---     | ---   | ---   | YES   | YES     | ---  | TODO  | YES
+    gles3/webgl2        | ---     | ---   | ---   | YES   | YES     | ---  | ---   | YES
+    metal               | ---     | YES   | ---   | YES   | ---     | ---  | ---   | ---
+    d3d11               | YES     | ---   | ---   | ---   | ---     | YES  | ---   | ---
+    KEY_DOWN            | YES     | YES   | YES   | SOME  | TODO    | YES  | TODO  | YES
+    KEY_UP              | YES     | YES   | YES   | SOME  | TODO    | YES  | TODO  | YES
+    CHAR                | YES     | YES   | YES   | YES   | TODO    | YES  | TODO  | YES
+    MOUSE_DOWN          | YES     | YES   | YES   | ---   | ---     | YES  | TODO  | YES
+    MOUSE_UP            | YES     | YES   | YES   | ---   | ---     | YES  | TODO  | YES
+    MOUSE_SCROLL        | YES     | YES   | YES   | ---   | ---     | YES  | TODO  | YES
+    MOUSE_MOVE          | YES     | YES   | YES   | ---   | ---     | YES  | TODO  | YES
+    MOUSE_ENTER         | YES     | YES   | YES   | ---   | ---     | YES  | TODO  | YES
+    MOUSE_LEAVE         | YES     | YES   | YES   | ---   | ---     | YES  | TODO  | YES
+    TOUCHES_BEGAN       | ---     | ---   | ---   | YES   | YES     | TODO | ---   | YES
+    TOUCHES_MOVED       | ---     | ---   | ---   | YES   | YES     | TODO | ---   | YES
+    TOUCHES_ENDED       | ---     | ---   | ---   | YES   | YES     | TODO | ---   | YES
+    TOUCHES_CANCELLED   | ---     | ---   | ---   | YES   | YES     | TODO | ---   | YES
+    RESIZED             | YES     | YES   | YES   | YES   | YES     | YES  | ---   | YES
+    ICONIFIED           | YES     | YES   | YES   | ---   | ---     | YES  | ---   | ---
+    RESTORED            | YES     | YES   | YES   | ---   | ---     | YES  | ---   | ---
+    SUSPENDED           | ---     | ---   | ---   | YES   | YES     | YES  | ---   | TODO
+    RESUMED             | ---     | ---   | ---   | YES   | YES     | YES  | ---   | TODO
+    QUIT_REQUESTED      | YES     | YES   | YES   | ---   | ---     | ---  | TODO  | YES
+    UPDATE_CURSOR       | YES     | YES   | TODO  | ---   | ---     | TODO | ---   | TODO
+    IME                 | TODO    | TODO? | TODO  | ???   | TODO    | ---  | ???   | ???
+    key repeat flag     | YES     | YES   | YES   | ---   | ---     | YES  | TODO  | YES
+    windowed            | YES     | YES   | YES   | ---   | ---     | YES  | TODO  | YES
+    fullscreen          | YES     | YES   | YES   | YES   | YES     | YES  | TODO  | ---
+    pointer lock        | TODO    | TODO  | TODO  | ---   | ---     | ---  | TODO  | TODO
+    screen keyboard     | ---     | ---   | ---   | YES   | TODO    | TODO | ---   | YES
+    swap interval       | YES     | YES   | YES   | YES   | TODO    | ---  | TODO  | YES
+    high-dpi            | YES     | YES   | TODO  | YES   | YES     | YES  | TODO  | YES
+    clipboard           | YES     | YES   | TODO  | ---   | ---     | TODO | ---   | YES
 
     TODO
     ====
@@ -998,9 +998,23 @@ inline int sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
     #endif
 #elif defined(_WIN32)
     /* Windows (D3D11 or GL) */
-    #define _SAPP_WIN32 (1)
-    #if !defined(SOKOL_D3D11) && !defined(SOKOL_GLCORE33)
-    #error("sokol_app.h: unknown 3D API selected for Win32, must be SOKOL_D3D11 or SOKOL_GLCORE33")
+    #include <winapifamily.h>
+    #if (defined(WINAPI_FAMILY_PARTITION) && !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP))
+        #define _SAPP_UWP (1)
+        #if !defined(SOKOL_D3D11)
+        #error("sokol_app.h: unknown 3D API selected for UWP, must be SOKOL_D3D11")
+        #endif
+        #if defined(CINTERFACE) || defined(D3D11_NO_HELPERS) || defined(COBJMACROS) || defined(WIN32_LEAN_AND_MEAN) || defined(NOMINMAX)
+        #error("sokol_app.h: incompatible preprocessor directives found. Do not copmile implementation in the same translation unit as sokol_gfx")
+        #endif
+        #if !defined(__cplusplus)
+        #error("UWP bindings require C++/17")
+        #endif
+    #else
+        #define _SAPP_WIN32 (1)
+        #if !defined(SOKOL_D3D11) && !defined(SOKOL_GLCORE33)
+        #error("sokol_app.h: unknown 3D API selected for Win32, must be SOKOL_D3D11 or SOKOL_GLCORE33")
+        #endif
     #endif
 #elif defined(__ANDROID__)
     /* Android */
@@ -1127,23 +1141,21 @@ inline int sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
     #include <windows.h>
     #include <windowsx.h>
     #include <shellapi.h>
-    #pragma comment (lib, "Shell32.lib")
     #if !defined(SOKOL_WIN32_FORCE_MAIN)
         #pragma comment (linker, "/subsystem:windows")
     #endif
-    #if (defined(WINAPI_FAMILY_PARTITION) && !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP))
-        #pragma comment (lib, "WindowsApp.lib")
-    #else
-        #pragma comment (lib, "user32.lib")
-        #if defined(SOKOL_D3D11)
-            #pragma comment (lib, "dxgi.lib")
-            #pragma comment (lib, "d3d11.lib")
-            #pragma comment (lib, "dxguid.lib")
-        #endif
-        #if defined(SOKOL_GLCORE33)
-            #pragma comment (lib, "gdi32.lib")
-        #endif
+
+    #pragma comment (lib, "user32.lib")
+    #pragma comment (lib, "Shell32.lib")
+    #if defined(SOKOL_D3D11)
+        #pragma comment (lib, "dxgi.lib")
+        #pragma comment (lib, "d3d11.lib")
+        #pragma comment (lib, "dxguid.lib")
     #endif
+    #if defined(SOKOL_GLCORE33)
+        #pragma comment (lib, "gdi32.lib")
+    #endif
+    
     #if defined(SOKOL_D3D11)
         #ifndef D3D11_NO_HELPERS
             #define D3D11_NO_HELPERS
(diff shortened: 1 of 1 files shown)


floooh:
Wow that's awesome! It'll be a little while until I'll be able to look into and merge the PR, but I definitely want to get this in.

floooh:
Ok, I'll start looking into this PR now. Is the PR ready for merging? (apart from the conflicts which I'll take care of). I've been quickly skimming over the code and found nothing to complain about, but I'll take some time to thoroughly understand the code (and I'll also add a uwp sample subdirectory to sokol-samples).

Btw, what's the code in this file for?

https://github.com/Aftnet/sokol_app_uwp/blob/master/src/sokol/sokol_uwp_impl.cpp

It looks like it's mostly a duplicate of what's in the PR?

albertofustinoni:
Great to hear.

How did you get to [that file](https://github.com/Aftnet/sokol_app_uwp/blob/master/src/sokol/sokol_uwp_impl.cpp)? It should not be part of the PR, or even in the code base at all at this point.

I started working on the UWP implementation in a separate file to make things easier for myself, but moved everything to the header in the end.

Also, don't remember if I mentioned this before or in the file itself, but to make things work with UWP one needs to have `sokol_app.h` and `sokol_gfx.h` included from different files for the implementation, due to gfx needing `CINTERFACE`, `COBJMACROS` and the UWP version needing the C++ interface.

Here's what I do myself:

implement.c
```
#include "sokol_app.h"

#define SOKOL_IMPL
#include "sokol_gfx.h"
#include "sokol_glue.h"
#include "sokol_audio.h"
#include "sokol_time.h"
```

implement.cpp
```
#include "sokol_gfx.h"

#define SOKOL_IMPL
#include "sokol_app.h"
```

floooh:
> one needs to have sokol_app.h and sokol_gfx.h included

Yikes ok. Did you try whether calling the D3D11 C interface would work from the UWP specific code? Or is it already some UWP specific header which includes d3d11.h and requires the D3D11 C++ interface? I'd really like to find a solution for this before the merge.

Another solution that would generally work is to call the D3D11 C++ API if the sokol_gfx.h implementation was included in a C++ source. If the sokol_app.h UWP code path can't use the D3D11 C interface, then that's something I'll work on first before the merge.

floooh:
> How did you get to that file?

I was looking around there because you posted the link to the project in the first comment ;)

I'll also try to get UWP support into fips first and create a simple uwp_entry.cc (similar to d3d11_entry.c) in the sokol-samples repository.

albertofustinoni:
Personally I'd rather go with the second option, mainly because using the C++ interface to D3D11 allows for using smart COM pointers and so prevents memory leaks.

The UWP code I used as base comes from Microsoft itself (it's the Visual Studio UWP/D3D app template, which I ported to pure C++ from the original C++/CX) and relies on smart pointers to work - and I would not want to mess with the design.

floooh:
Ok, then I'll create an issue for myself to transparently use the D3D11 C++ API in sokol_gfx.h if the implementation is compiled as C++.

albertofustinoni:
Awesome, thanks :)

trojanfoe:
I think it's great to have UWP support in Sokol; nice work!

I have a question about the use of shaders however.  In [sample code](https://github.com/Aftnet/sokol_app_uwp/blob/master/src/SampleAppCode.c) provided the shaders are compiled from text source within the exceutable however I understood UWP on XBox One to forbid this and you had to precompile them?  This may not be a problem when using the sokol-shdc cross-compiler as that generates C header files which would work fine in UWP I assume?

But I presume you've tested this on XBox one?

floooh:
> ...however I understood UWP on XBox One to forbid this and you had to precompile them?

There are several options:

- use sokol-shdc integrated into the build process as in the sampled under the sokol-samples/sapp directory
- compile shaders "offline" via sokol-shdc or the DX shader compile, and integrate the resulting byte code manually (like I did in the sokol-utility headers (e.g. https://github.com/floooh/sokol/blob/9e1d008c024149750b1f07f133fa9674274fb457/util/sokol_debugtext.h#L3079)
- ...or compile the shaders offline and load them from the file system

floooh:
FYI: sokol_gfx.h and sokol_app.h now use the D3D11 and DXGI C++ APIs when the implementation is compiled as C++. I will now start looking into this PR.

floooh:
Ok I have manually merged the PR into this branch (because of the merge conflicts, it wasn't anything "serious" though) and will continue testing there before merging into master:

https://github.com/floooh/sokol/tree/Aftnet-sapp_uwp

albertofustinoni:
Once again, thank you very much

I have updated my own project to the latest `sokol_gfx` and can confirm it works (desktop Windows/UWP/macOS/iOS) with my version of `sokol_app`

Two things I have noticed:

- For things to work one must include `sokol_app.h` before `sokol_gfx.h`
- For desktop Windows Sokol must be coupled as C; C++ breaks

The above are not deal breakers (Apple stuff already needs to be compiled as ObjC), was just wondering if you were aware.

floooh:
> For things to work one must include sokol_app.h before sokol_gfx.h

This intended, it's also required for other platforms (this behaviour may change later though, but this would for instance require to move the "MiniGL loader" from sokol_app.h into sokol_gfx.h... long term that may be better, but right now the sokol_app.h implmentation should be included before the sokol_gfx.h implementation when both are used together).

> For desktop Windows Sokol must be coupled as C; C++ breaks

This is surprising, can you maybe open a separate issue with the errors you are seeing?

floooh:
> For desktop Windows Sokol must be coupled as C; C++ breaks

...maybe a reason could be that you only replaced sokol_gfx.h but use your own sokol_app.h, not sure if that's it, just something that popped into my mind right now.

albertofustinoni:
Yeah, I also thought that may be the case after I wrote that. Everything seems good then :)

floooh:
@albertofustinoni another question: what Visual Studio version are you using? I'm getting some mysterious compilation errors and I'm stating to suspect that VS2017 doesn't have complete C++17 support (I'll try the VS2019 Community Edition next).

floooh:
Hmm, it's not VS2019 vs VS2017, I'm getting those errors (in case you've seen those too and know what's the problem, otherwise don't worry, I'll figure it out somehow):

```
2>E:\projects\sokol\sokol_app.h(6228,56): error C2672: 'IID_PPV_ARGS_Helper': no matching overloaded function found
2>E:\projects\sokol\sokol_app.h(6228,56): error C2784: 'void **IID_PPV_ARGS_Helper(T **)': could not deduce template argument for 'T **' from 'winrt::com_ptr<ID3D11Texture2D1> *'
2>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\combaseapi.h(202): message : see declaration of 'IID_PPV_ARGS_Helper'
2>E:\projects\sokol\sokol_app.h(6228,88): error C2660: 'IDXGISwapChain::GetBuffer': function does not take 2 arguments
2>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\shared\dxgi.h(1718,43): message : see declaration of 'IDXGISwapChain::GetBuffer'
2>E:\projects\sokol\sokol_app.h(6373,55): error C2672: 'IID_PPV_ARGS_Helper': no matching overloaded function found
2>E:\projects\sokol\sokol_app.h(6373,55): error C2784: 'void **IID_PPV_ARGS_Helper(T **)': could not deduce template argument for 'T **' from 'winrt::com_ptr<IDXGIFactory4> *'
2>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\combaseapi.h(202): message : see declaration of 'IID_PPV_ARGS_Helper'
2>E:\projects\sokol\sokol_app.h(6373,83): error C2660: 'IDXGIObject::GetParent': function does not take 1 arguments
2>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\shared\dxgi.h(338,43): message : see declaration of 'IDXGIObject::GetParent'
2>E:\projects\sokol\sokol_app.h(6384,49): error C2672: 'IID_PPV_ARGS_Helper': no matching overloaded function found
2>E:\projects\sokol\sokol_app.h(6384,49): error C2784: 'void **IID_PPV_ARGS_Helper(T **)': could not deduce template argument for 'T **' from 'winrt::com_ptr<IDXGIFactory4> *'
2>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\combaseapi.h(202): message : see declaration of 'IID_PPV_ARGS_Helper'
2>E:\projects\sokol\sokol_app.h(6384,78): error C2660: 'CreateDXGIFactory1': function does not take 1 arguments
2>C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\shared\dxgi.h(2168,16): message : see declaration of 'CreateDXGIFactory1'
2>C:\Program Files (x86)\Windows Kits\10\bin\10.0.18362.0\XamlCompiler\Microsoft.Windows.UI.Xaml.Common.targets(482,5): error MSB4181: The "CompileXaml" task returned false but did not log an error.
```

albertofustinoni:
You need to be on VS2019 for C++/WinRT support (in 2017 it was experimental at best), also due to improved C++ conformance of 2019.

Beyond that, as indicated [here](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/intro-to-using-cpp-with-winrt#visual-studio-support-for-cwinrt-xaml-the-vsix-extension-and-the-nuget-package), you will need a few things

- Check the exact VS components installed
> From within the Visual Studio Installer, install the Universal Windows Platform development workload. In Installation Details > Universal Windows Platform development, check the C++ (v14x) Universal Windows Platform tools option(s), if you haven't already done so
- Install the [C++/WinRT extension](https://marketplace.visualstudio.com/items?itemName=CppWinRTTeam.cppwinrt101804264) (what I have done) or add the [C++/WinRT NuGet package](https://www.nuget.org/packages/Microsoft.Windows.CppWinRT/) to the project

I have the extension installed on my systems because it adds debugger support (and no NuGet package in my build, I build with CMake and have not added it to the CMakefile) and I build reliably.

floooh:
Hmm no luck so far, the core error which is repeated a few times in other places is this:

```
'void **IID_PPV_ARGS_Helper(T **)': could not deduce template argument for 'T **' from 'winrt::com_ptr<ID3D11Texture2D1> *'
```

The other interface where this happens is ```IDXGIFactory4```.  Both seem to be in "extension" APIs (e.g. not in core D3D11 and DXGI, but instead in d3d11_3.h and dxgi1_4.h). But those *are* included (at least I find them in sokol_app.h. I'll continue looking...

floooh:
The official sample here compiles fine, but there seems to be some sort of code generation involved:

https://github.com/microsoft/Windows-universal-samples/tree/master/Samples/Simple3DGameDX/cppwinrt

After compilation there's a "Generated Files" directory which contains all the winrt headers in some sort of preprocessed form. If I remove the ```IID_PPV_ARGS_Helper``` function from those generated headers I get the same compilation error that I'm seeing when compiling sokol_app.h. I'm currently trying to understand what's going on there but damn that's some complex bullshit going on... starting to question whether this is worth the hassle ;)

floooh:
...right know it looks like the problem is that I don't have the cppwinrt NuGet package which silently overrides the Windows SDK WinRT headers. At least that's the difference between my project and the SimpleDXGame project. The SimpleDXGame project has the NuGet dependency defined in the Visual Studio solution, but I really don't know if I want to go there. I'll see if cmake has a simple solution (it seems to have with VS_PACKAGE_REFERENCES...).

floooh:
Ok it compiles when adding the CppWinRT NuGet package in Visual Studio to the sokol build target. But I really need to find a way to do this via cmake (this VS_PACKAGE_REFERENCES didn't work out of the box, but I'm probably using it wrong...).

albertofustinoni:
The extension did not help? That's weird: I also use CMake as the build system for my project and I have not included the nuget package, but things still compile. I thought that was thanks to the extension.

in any case, your analysis is correct: as far as I know what MS is moving towards is to generate the UWP (WinRT) API projections into various languages (C#, C++, Rust) as part of the project build pipeline, with them only shipping winmd files as part of the SDK.

In any case, I was planning on using the VS_PACKAGE_REFERENCES feature in CMake eventually myself to remove the dependency on the extension...

floooh:
Ok, I think I found the "proper" solution: using the latest Windows SDK (10.0.19041.0) offered by the VS2019 installer instead of the NuGet package also seems to fix the compile problems. By default, only the SDK version 10.0.18362.0 is installed, and this has a different implementation of that IID_PPV_ARGS_Helper which doesn't work with winrt::com_ptr<>, I guess the NuGet package is the "bleeding edge" which contains fixes that then will go into the next Windows SDK.

Continuing as normal now :)

floooh:
Ok, I got it basically working, just a few things:

- I have replaced wWinMain() with the regular WinMain() (e.g. applications should be compiled without the UNICODE define, instead application strings should be treated as UTF-8 and converted from and to UTF-16 when needed for calling Windows APIs)

- For some reason, sokol_imgui.h doesn't detect mouse clicks, the mouse button callbacks are called though.

I'll spend a few days getting familiar with the code, e.g. I wonder if having those separate DeviceResources and Renderer classes is really needed or useful... (I guess that's how the UWP example programs are structured, but I also don't want to change the code too much unless it really makes sense).

image

albertofustinoni:
Oh interesting: I have worked on the latest Win10 SDK (10.0.19041.0) since the beginning, which is why I never noticed anything.

Never tried using imgui, although at some point I may need to look into it (will need some in app UI at some point). Does it work for touch or is it only for mouse/keyboard?

As for command line handling, I *think* it is supported, bu I am not sure: UWP apps are meant to be launched from start menu icons for the most part, but I think MS added support for command line UWP apps too at some point.

floooh:
A little update: one thing that was messed up in the mouse button code was that left and middle mouse button was mixed up: 

https://github.com/floooh/sokol/blob/9929134012699d7677bf64eb8ded02ed4f571007/sokol_app.h#L5282-L5289

But now I'm stuck with detecting which mouse button was actually released, because I'm only getting this generic PointerRelease event, but I'm not seeing way how to detect which mouse button was released.

I'll continue with this tomorrow...

floooh:
...ok fixing the mouse button input was actually quite the rabbithole, because UWP also doesn't send Pressed/Released events when multiple buttons are pressed simultaneously, even SDL seems to get this wrong.

I came up with the following solution, which seems to work. I need to test more input edge cases though:

https://github.com/floooh/sokol/blob/2b717d88b4cb42d96390657a21dbe118bb52e91c/sokol_app.h#L5715-L5758

albertofustinoni:
Wow, I am sorry: I did the mouse implementation quickly because I don’t really need it but wanted to get as close to feature parity as I could with the other platforms and mistakes slipped in.

Never thought it would turn into this much of a rabbit hole.

floooh:
Hmm more UWP strangeness... The window.Closed event handler never seems to be called (this also means the loop in App::Run is never finished, the application simply terminates. Instead when clicking the window-close-button, App::OnSuspending() is called.

I wonder now where to best put the _sapp_call_cleanup() call. OnSuspending() seems to be the wrong place, because that implies that OnResuming() could be called later.

I'm also not seeing other event handlers like "OnTerminated()"... PS: probably like this: https://docs.microsoft.com/en-us/windows/uwp/launch-resume/app-lifecycle#app-close

floooh:
Ok, nearly done now, I think there's only one topic remaining: DPI-handling.

On my HighDPI Windows laptop the UWP backend behaves entirely different than the Win32 backend (the UWP backend always uses HighDPI rendering, no matter what the sokol-app high_dpi flag says, because the UWP backend uses some random threshold values (192 dpi, 1920x1080 pixels) to decide whether rendering should happen in high-dpi or not.

Apart from always using HighDPI rendering on my laptop, this also throws off the mouse coordinates, they seem to be off in all samples when running on this laptop.

I'll need to figure out how to rewrite the HighDPI handling in the UWP backend so that it behaves identical to the other platforms, so it'll be a couple more days until merging...

floooh:
PS: the new mouse-lock feature also requires work, but I'll simply mark this is "TODO" in the feature matrix, not a showstopper for the merge.

floooh:
Note to self: compute new mouse dx / dy in events in OnPointerMove

floooh:
I just noticed that MSAA support is missing, I'll start working on this now (good thing is that I can mostly use the same approach for the Win32/D3D11 backend, because this currently prints a deprecated warning, DXGI MSAA support in the swap chain must be implemented similar to WebGPU, with a separate MSAA render target and an explicit resolve into the framebuffer surface).

floooh:
@albertofustinoni I think you can give the Aftnet-sapp_uwp branch a whirl now, this is the version I'd like to merge to master (only thing missing is some notes in the sokol_app.h documentation header, I'll finish that today or tomorrow).

albertofustinoni:
@floooh Just gave your branch a spin and linker complains about wWinMain being undefined.

Then I noticed your comment above about Unicode and I should say Windows applications (not just UWP) *need* to be compiled with `UNICODE` defined in order to be able to actually use Unicode strings for Windows API calls. It's what you get by default both on modern Visual Studio and CMake too.

Sadly Windows just does not do UTF8 at all: when you have API calls that have a `wide_char` version and a char version (think `CreateFileA` vs `CreateFileW`) the char (*A) version **only supports a single codepage** (the one set system wide here).
image

Apps that work with UTF8 strings need to convert to wide chars when calling into the API and back.

See [here](https://stackoverflow.com/questions/4592261/windows-api-ansi-and-wide-character-strings-is-it-utf8-or-ascii-utf-16-or-u) for reference.

I would strongly recommend you have `sokol_app` at least compatible with `UNICODE`, and ideally implemented with it in mind.

floooh:
> Sadly Windows just does not do UTF8 at all

This isn't really a problem at least when using Win32 calls (I'm not sure about WinRT), you can compile Windows applications without the UNICODE define but still use the wide-string API calls by explicitly calling *W() functions. The Windows code page settings are also irrelevant, those settings are only relevant for the *A() (ASCII) Windows API functions, not for the wide-string functions.

The usual strategy is to treat all application-internal strings as UTF-8 encoded, end only covert from and to UTF-16 (in Windows lingo "UNICODE" although that's incorrect) when calling Windows API functions.

See here for instance in the sokol_app.h Win32 code:

https://github.com/floooh/sokol/blob/342821c8452fff509107fe4ddc6318335208e3ea/sokol_app.h#L5332-L5344

This is the "UNICODE" Version of CreateWindowEx() which accepts UTF-16 encoded "wide-strings", the window title is passed as UTF-8 encoded string to sokol-app, and converted to UTF-16 via MultiByteToWideChar:

https://github.com/floooh/sokol/blob/342821c8452fff509107fe4ddc6318335208e3ea/sokol_app.h#L4790-L4803

I've been using that approach for the last 15 years or so and it works great, it's also what's recommended in the utf8everywhere manifesto:

https://utf8everywhere.org/

I'll think about checking if UNICODE is defined and then use wWinMain() instead of WinMain(), but it might be confusing because all the sokol-APIs will still require strings to be passed in as UTF-8 encoded, not UTF-16, and that's where I won't budge ;)

albertofustinoni:
I think we are actually saying the same thing: app code, including sokol, should be using UTF8 and only translate to UTF16 when dealing with Windows APIs. I am definitely *not* arguing for UTF16 usage outside anywhere else.

Allowing compilation with UNICODE defined (converting `wWinMain` args to UTF8 for processing is sokol) is just part of that approach, IMO.

floooh:
@albertofustinoni do you see a Character Set configuration option in your Visual Studio project settings? I'm a bit confused because I don't see that in my C++/WinRT project (only in Win32 projects). I got a wWinMain() linker error in the original code and had to switch to WinMain(), so I guess there must be an option somewhere to switch UNICODE on and off, I'm just not finding the option anywhere. 

I want to test a hybrid entry point (use wWinMain when UNICODE is defined, and WinMain otherwise), but on one hand I don't know why my project (created with cmake) defaults to UNICODE switched off, and on the other hand I'm confused because I'm not seeing a related option in the Visual Studio project settings dialog (like I'm seeing in a Win32 project).

floooh:
Ah nvm, I think I figured it out. Instead of a separate compiler option this simply seems to be done now through standard preprocessor defines (e.g. UNICODE and _UNICODE are part of the %(PreprocessorDefinitions) "config variable".

floooh:
hmmm ..strangely enough, both ```UNICODE``` and ```_UNICODE``` *are* defined in my project, but this doesn't seem to "activate" the wWinMain() entry point... instead I'm getting a linker error when wWinMain() is used, because WinMain() is expected:

```
>MSVCRTD.lib(exe_winmain.obj) : error LNK2019: unresolved external symbol WinMain referenced in function "int __cdecl invoke_main(void)" (?invoke_main@@YAHXZ)
```
This also sort of kills my plan to use wWinMain() when UNICODE is defined and WinMain() otherwise. I need to figure out why my build config expects WinMain() first...

floooh:
Another weirdness is that the CoreApplication C++/WinRT example accepts both wWinMain() or WinMain() without any linker errors. So far I haven't found any compiler/linker option differences which would explain the different behaviour. It seems to be about the Runtime Library that's being linked, but I can't see a difference there, nor in the linker's "subsystem".

albertofustinoni:
Hello

Unfortunately I’m gonna be away from home and PC for the next three days, so I won’t be able to help much :(

In my experience too WinRT projects do not have a codepage setting, and wWinMain should be enabled by defining `UNICODE`

floooh:
Ok, defining UNICODE doesn't seem to select between WinMain or wWinMain (the Windows SDK headers always define both entry points). The linker error seems to be related to the runtime libraries that are linked. I'll dig a bit deeper today and try to find out what's different to the C++/WinRT samples that come with the Visual Studio extension (interestingly, those accept *both* WinMain or wWinMain).

floooh:
Ok, I think I found the solution, but it's very non-obvious. Basically if wWinMain() is defined in static library, the linker won't recognize it, and this magic line is needed in code that links against this library:

https://github.com/floooh/sokol/blob/88fb6cd5a225d978a80cdf9d2b03d8fe276e4d3a/sokol_app.h#L1082

(which I'm doing now in the declaration part when compiling for C++/WinRT, so that this line is "injected" into each code module which uses sokol_app.h).

If this line exists, an unresolved-error happens if there is no wWinMain(). Otherwise there seems to be some sort of autodetection happening when wither WinMain() or wWinMain() is defined in a "top-level" object-file which is fed directly into the linker (not indirectly via a static link library), because when I added either a WinMain() or wWinMain() directly to my samples (or included the sokol_app.h implementation directly into the sample source code), both WinMain() or wWinMain() worked (e.g. WinMain also worked when UNICODE is defined, which is the case in my WinRT/C++ projects, since that's automatically defined by the Visual Studio build system).

When you have a minute, please check whether this works now on your side (sokol_app.h now selected wWinMain() if UNICODE is define, otherwise WinMain():

https://github.com/floooh/sokol/blob/88fb6cd5a225d978a80cdf9d2b03d8fe276e4d3a/sokol_app.h#L6649-L6653

floooh:
Btw, question @albertofustinoni: I've tried to setup my Xbox One for testing this stuff but it seem this needs a paid "Partner Center" account now. As far as I remember this "just worked" on the 360 without a separate account. Are there no free options anymore?

albertofustinoni:
Just had a chance to test your changes and now the code in `Aftnet-sapp_uwp` works for me :)

Massive thanks for everything so far.

As for the Partner Center account: looks like they made it paid for now :(
When I got mine, back in the Windows 8/Windows Phone 8 days, it was free, and I guess those accounts just got grandfathered.
At least it's not 100 bucks and yearly renewals *glares angrily at Apple*

floooh:
Alright, thanks for testing! I'll merge the branch back to master within the next few days then :) 

I think I'll bite the bullet and pay for a private account, AFAIK it's just a one-time payment. Now that it's possible I also want to see the code run on the Xbox One.

Btw, Apple now offers dev accounts that are free and allow running code on your device, the $99 tax is only needed if you want to publish the application ;)

albertofustinoni:
Awesome, thanks again

I thought the free Apple dev account forces apps on devices to expire after one week. Has this changed recently or is there a workaround?

BTW, would you be interested in a gamepad API? I have made my own following sokol’s code style and it works on Windows (desktop/UWP via XINPUT), Mac/iOS and Android.

floooh:
> I thought the free Apple dev account forces apps on devices to expire after one week. Has this changed recently or is there a workaround?

Yeah there are definitely limitations, you can also only have up to 8 different apps on the device at the same time.

> in a gamepad API

That's definitely something I'd want to look at because my rough plan for the next sokol headers looks like this:

1. sokol_framebuffer.h: simple 2D "framebuffer" for setting pixels with the CPU and rendering via sokol_gfx.h
2. **sokol_gamepad.h**: gamepad input with roughly the same feature set as the Oryol gamepad API (including button- and axis-mapping from raw gamepad-specific buttons and axes to "semantic" buttons and axes)
3. sokol_gesture.h: simple cross-platform touch gesture gesture recognition on top of the low-level touch events from sokol_app.h

floooh:
Ok merged!

PR #364, opened 2020-08-23
set window title on macos

medvednikov:
This is a very early phase. I don't know want what kind of API you want. Probably one function `sapp_set_title(const char* title)`?

I'll add Windows and Linux a bit later.

Also, does Sokol plan to support multiple windows? Right now all functions seem to be using a global main window.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1032,6 +1032,8 @@ SOKOL_API_DECL const void* sapp_metal_get_renderpass_descriptor(void);
 SOKOL_API_DECL const void* sapp_metal_get_drawable(void);
 /* macOS: get bridged pointer to macOS NSWindow */
 SOKOL_API_DECL const void* sapp_macos_get_window(void);
+/* macOS: set window title */
+SOKOL_API_DECL void sapp_macos_set_title(const char* title);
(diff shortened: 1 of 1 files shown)


floooh:
The function name should probably be sapp_set_window_title() (because it's also called window_title in the sapp_desc struct).

There shouldn't be platform-specific public API functions for this, but only one function for all platforms, which then forks to platform specific private functions _sapp_macos_set_window_title(), _sapp_win32_set_window_title() etc... like it's done here:

https://github.com/floooh/sokol/blob/c6749d70435c4330f00f7f973eeb50a82d448d90/sokol_app.h#L8197-L8205

medvednikov:
Yeah, that's what I thought. Will push a fix.

floooh:
Ah, forgot something:

> Also, does Sokol plan to support multiple windows?

Maybe at a later time, but don't worry about that yet :)

floooh:
I'll start merging this PR next (but will make the changes discussed above).

floooh:
Ok merged. The new public function is called sapp_set_window_title(), and it's implemented for the macOS, Win32 and Linux backends. I also added a test/example to the events-sapp sample.

Thanks for the PR!

PR #374, opened 2020-09-12
fixing sdtx_putr

rbxnk:


floooh:
Ooops, good catch. I'll try to come up with a test for this in https://github.com/floooh/sokol-samples/blob/master/tests/sokol-debugtext-test.c and then merge the PR.

Thanks!

floooh:
Ok, I have added a test here which reproduced the problem:

https://github.com/floooh/sokol-samples/blob/2665dc80c484308ff3446c050fbc0b2bbda14cdd/tests/sokol-debugtext-test.c#L418-L430

...and a visual test in the debugtext-printf-sapp sample here:

https://github.com/floooh/sokol-samples/blob/2665dc80c484308ff3446c050fbc0b2bbda14cdd/sapp/debugtext-printf-sapp.c#L75-L76

rbxnk:
Very thorough! Thanks!

PR #390, opened 2020-09-24
default imgui.ini

RandyGaul:
Dear ImGui uses a default path for saving window settings as "imgui.ini", but in sokol_imgui.h there is no default, so this pull request fixes that.

If this value isn't set then window settings are not preserved across sessions, which is really very annoying.

### DIFF
--- a/util/sokol_imgui.h
+++ b/util/sokol_imgui.h
@@ -1595,6 +1595,7 @@ SOKOL_API_IMPL void simgui_setup(const simgui_desc_t* desc) {
     SOKOL_ASSERT(desc);
     memset(&_simgui, 0, sizeof(_simgui));
     _simgui.desc = *desc;
+    _simgui.desc.ini_filename = _simgui_def(_simgui.desc.ini_filename, "imgui.ini");
     _simgui.desc.max_vertices = _simgui_def(_simgui.desc.max_vertices, 65536);
     _simgui.desc.dpi_scale = _simgui_def(_simgui.desc.dpi_scale, 1.0f);
     #if !defined(SOKOL_IMGUI_NO_SOKOL_APP)


floooh:
I need to check how this affects platforms without or optional filesystems (like WASM). IFIR I've kept that path zero-initialized on purpose because it caused issues on some platforms (definitely on WASM), and setting the ini-path to zero was the easiest way to let ImGui skip any filesystem calls (and the simgui_desc item was a way for users to override that behaviour and enable ImGui's persistency code.

floooh:
As discussed on twitter, it's probably best to use the default "imgui.ini" on "regular desktop platforms" which are guaranteed to have a filesystem, but zero on other platforms.

Another unrelated problem is that "current working directory" doesn't mean much on some platforms, it can point to any directory depending on how the application is started. But I think this is something that must be fixed by the user via the regular Dear ImGui API.

RandyGaul:
After looking around it seems like Dear ImGui just defers this problem to the user and lets them save/read ini to/from memory as an optional API, which sounds like it could be used on platforms without a traditional file system as an alternative.

floooh:
I need to dig into the whole WASM topic for this again. IFIR my main concern was that the default filesystem in ImGui was causing the filesystem wrapper code to be pulled in (or required it to be enabled), which was bloating WASM modules quite a bit.

Maybe this also has changed in the meantime in ImGui itself and there's a better way.

floooh:
...ok, I played around with ImGui's state saving a bit, and I think it's better to keep it as it is (e.g. if users wants to use ImGui's automatic state save/load feature they can set ```simgui_desc.ini_filename``` to a path, or they can call the ```Load/SaveIniSettingsFromMemory/Disk()``` explicitly.

Instead I added a bit of documention to the header:

https://github.com/floooh/sokol/blob/d8c12f7414217f39f7f5573f225654e9dd00cc63/util/sokol_imgui.h#L119-L128

PR #392, opened 2020-09-28
carp_rwops.h support read from local file system include assets of android/ios

alittlesail:
I copy SDL_rwops file and modify to stb code style and you can merge this to your sokol_fetch.h


floooh:
That's not the proper way to integrate functionality like this, it shouldn't be in a separate header, and the code shouldn't be copied without the original license and attribution.

alittlesail:
ok, you are right. so I hope this codes can help you support load assets of android/ios

PR #393, opened 2020-09-28
Gamepad API

albertofustinoni:
Proposal for a sokol gamepad API

### DIFF
--- a/sokol_gamepad.h
+++ b/sokol_gamepad.h
@@ -1,0 +1,591 @@
+#ifndef SOKOL_GAMEPAD_INCLUDED
+/*
+sokol_gamepad.h -- cross-platform gamepad API
+
+    Project URL: https://github.com/floooh/sokol
+
+    Do this:
+        #define SOKOL_IMPL
+    before you include this file in *one* C or C++ file to create the
+    implementation.
+
+    This is a wrapper around each platform's native gamepad APIs:
+    - xinput on Windows (7 and up, Win32 and UWP)
+    - game controller framework on macOS/iOS/tvOS
+    - input API on Android
+
+    and abstracts gamepads to the layout of an idealized Xbox 360 gamepad:
+    - one d-pad
+    - two analog sticks acting as buttons when pressed
+    - guide and back buttons
+    - two analog shoulder buttons
+    - two analog triggers
+    This is the most popular and comprehensive configutration across game systems
+    (PS1/PS2/PS3/PS4/Xbox/Xbox360/Xbox One/Switch map directly, just for start)
+
+    To use:  
+    - In you app initialization code call sgamepad_init()
+    - At the exact time you want to record input state call sgamepad_record_state()
+    - Get the state for a particular gamepad like this:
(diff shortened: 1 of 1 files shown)


prime31:
This looks great! Really well done. Can’t wait to give it a go.

meshula:
Hi, I've successfully incorporated your proposed API into my app, but I had to make a few additions to match my use case, which requires detecting whether or not a controller is connected. I've submitted a PR here: https://github.com/Aftnet/sokol/pull/1

Querijn:
Darn, was hoping this was something that I could use, but the windows version seems to rely on xinput, which disallows you to use gamepads that aren't xbox controllers.

meshula:
I've been looking at this dinput implementation, but haven't had time to move forward yet. https://github.com/Kode/Kinc/blob/d153bf0642e7d56b1fa056f01005e979f0ce6267/Backends/System/Windows/Sources/kinc/backend/system.cpp

My current implementation is here - https://github.com/meshula/LabCamera/blob/main/examples/Navigator/sokol_gamepad.h - it's moved forward compared to the OP, but hasn't got dinput yet.

Querijn:
I've found out that XInput is the best simple approach, which allows just Xbox Controllers to work, and controllers that act like those. But I've also been looking at Raw Input, which is what seems to be what the web Gamepad API is based on (poorly).

I've come across a couple of issues while investigating this myself. :
- On Windows, Raw Input requires access to the message loop to catch WM_INPUT messages. This would mean hooking the WndProc pointer via SetWindowLongPtr
- The web's gamepad API leaves something to be desired. On Firefox, it decides to just convert every HID control value into an axis, which means the D-Pad is one messed up float for my PS5 controller. Chrome seems to convert these to buttons. The way vendors and products are indicated is also non-standard, whereas Firefox prepends the name with them, and Chrome just states them as human readable values

Left Chrome, right Firefox.
image

floooh:
Not sure if it's useful for the discussion, but in Oryol I had a mapping layer over the gamepad input because axis and button indices are not standardized in some APIs. The mapping layer is basically a dictionary of device id strings (or some sort of string pattern?) which maps raw axis and button indices to "semantic" names. Oryol came with builtin mappings for Xbox and PS4 controllers but user code could add additional mappings:

https://github.com/floooh/oryol/tree/master/code/Modules/Input#gamepad-input

(the Windows side also simply used XInput though, IFIR I had ripped the code from GLFW)

alterae:
any word on/plans to merge this or something like it soon?
it looks usable at a cursory glance, at least from C, but in its current state using it from Zig would be a (relative) hassle

floooh:
No plans to merge this PR as is, sorry. And getting consistent gamepad support across all platforms and gamepad types (e.g. anything that's not an Xbox controller on Windows, or a Playstation controller on Mac) isn't trivial.

alterae:
ah, yeah
looking through this in more detail now i see it only handles those two particular cases, oof

Querijn:
Has Firefox improved support yet? It was kind of awful last time I tried.

PR #394, opened 2020-09-28
UWP compatibility, no MMDEVICE

albertofustinoni:
Replicating #343

### DIFF
--- a/sokol_audio.h
+++ b/sokol_audio.h
@@ -480,8 +480,16 @@ inline void saudio_setup(const saudio_desc& desc) { return saudio_setup(&desc); 
     #endif
     #include <windows.h>
     #include <synchapi.h>
-    #pragma comment (lib, "kernel32.lib")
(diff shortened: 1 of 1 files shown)


floooh:
Thanks a lot for putting the PR up again, I'll start integrating this now (I'll do a manual merge like the other UWP PR, gives me an opportunity to get more familiar with the code)

floooh:
Ok merged. I had to fix two small things:

- two unused-variable-warnings in the new callback functions
- and I had to re-add a fix to ignore the return value of CoInitializeEx() in the Win32 code path, somehow that wasn't shown as a merge conflict

I also enabled the 'saudio-sapp' example in the sokol-samples repository, unfortunately the other sokol-audio-sample (modplay-sapp) doesn't work because libmodplug would require too many fixes to compile as UWP library.

albertofustinoni:
Yay!
Thanks so much again 👍

PR #398, opened 2020-10-02
sokol_gfx.h: Fix instancing when running in webgl fallback mode. (SOKOL_GLES2 & SOKOL_GLES3)

thehink:
To be honest I'm not 100% sure why this works. Would think that these defines should still be required when running webgl1.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -2502,7 +2502,7 @@ inline void sg_init_pass(sg_pass pass_id, const sg_pass_desc& desc) { return sg_
     #define GL_LUMINANCE 0x1909
     #endif
 
-    #ifdef SOKOL_GLES2
+    #if defined(SOKOL_GLES2) && !defined(SOKOL_GLES3)
     #   ifdef GL_ANGLE_instanced_arrays
     #       define SOKOL_INSTANCING_ENABLED
     #       define glDrawArraysInstanced(mode, first, count, instancecount)  glDrawArraysInstancedANGLE(mode, first, count, instancecount)


thehink:
Thanks for the explanation! That's actually what I tried first. Now that I'm looking at again it seems the problem actually lies in the shader creation step. This asserts triggers when only compiling with SOKOL_GLES3 and trying to run webgl. The shader header is created with sokol-shdc.

sokolshaderfail2

Should probably close this PR and create an issue instead.

floooh:
If this assert is hit it means sg_make_shader() is called with a nullptr. I bet this means that sokol-shdc has been told to only create shader for WebGL2, but not for WebGL. E.g. here's how it's handles in sokol-samples when compiled for WASM:

https://github.com/floooh/sokol-samples/blob/e42599040be34619b2dd88a1ed5c6873e72c8ffb/CMakeLists.txt#L44-L51

Note how ```glsl300es:glsl100``` is used as argument for the sokol-shdc -slang argument so that both WebGL2 and WebGL1 shaders are generated. If this would only be ```glsl300es```, and WebGL2 cannot be initialized and needs to fallback to WebGL1, the generated code which returns the ```sg_shader_desc``` struct would return a nullptr (because it can't find a shader for the WebGL1 backend), which in turn would trigger this assertion in sg_make_shader_desc().

I don't understand how this is related to the fix in this PR though...

floooh:
PS: here's what this generated struct should look like for WebGL2 with support for WebGL fallback:

```c
static inline const sg_shader_desc* triangle_shader_desc(void) {
  if (sg_query_backend() == SG_BACKEND_GLES2) {
    static sg_shader_desc desc;
    static bool valid;
    if (!valid) {
      valid = true;
      desc.attrs[0].name = "position";
      desc.attrs[1].name = "color0";
      desc.vs.source = vs_source_glsl100;
      desc.vs.entry = "main";
      desc.fs.source = fs_source_glsl100;
      desc.fs.entry = "main";
      desc.label = "triangle_shader";
    };
    return &desc;
  }
  if (sg_query_backend() == SG_BACKEND_GLES3) {
    static sg_shader_desc desc;
    static bool valid;
    if (!valid) {
      valid = true;
      desc.attrs[0].name = "position";
      desc.attrs[1].name = "color0";
      desc.vs.source = vs_source_glsl300es;
      desc.vs.entry = "main";
      desc.fs.source = fs_source_glsl300es;
      desc.fs.entry = "main";
      desc.label = "triangle_shader";
    };
    return &desc;
  }
  return 0; /* can't happen */
}
```

...there are two if-branches, one for SG_BACKEND_GLES2 and one for SG_BACKEND_GLES3. If sokol-shdc is told to only generate shader code for GLES3/WebGL2, than the function will fall through to the ```return 0;``` (which actually *can* happen as you are seeing).

thehink:
It's clear to me now why my PR made it work. Sokol-shdc was wrapping all shaders with ifdefs for respective backends.
So the WebGL1 code wasn't compiled at all with SOKOL_GLES3 which resulted in the nullptr assert for the WebGL1 codepath.
Really appreciate the help!
```
static inline const sg_shader_desc* triangle_shader_desc(void) {
#if defined(SOKOL_GLES2)
  if (sg_query_backend() == SG_BACKEND_GLES2) {
    static sg_shader_desc desc;
    static bool valid;
    if (!valid) {
      valid = true;
      desc.attrs[0].name = "position";
      desc.attrs[1].name = "color0";
      desc.vs.source = vs_source_glsl100;
      desc.vs.entry = "main";
      desc.fs.source = fs_source_glsl100;
      desc.fs.entry = "main";
      desc.label = "triangle_shader";
    };
    return &desc;
  }
#endif /* SOKOL_GLES2 */
#if defined(SOKOL_GLES3)
  if (sg_query_backend() == SG_BACKEND_GLES3) {
    static sg_shader_desc desc;
    static bool valid;
    if (!valid) {
      valid = true;
      desc.attrs[0].name = "position";
      desc.attrs[1].name = "color0";
      desc.vs.source = vs_source_glsl300es;
      desc.vs.entry = "main";
      desc.fs.source = fs_source_glsl300es;
      desc.fs.entry = "main";
      desc.label = "triangle_shader";
    };
    return &desc;
  }
  return 0; /* can't happen */
#endif /* SOKOL_GLES3 */
}
```

floooh:
I'm actually seeing the problem in the snippet you posted. Those #ifdef/#endif pairs shouldn't be there. You need to run sokol-shdc with the ```-noifdef``` command line option for the WebGL2=>WebGL fallback situation (note how those are missing in my snippet).

I'm also seeing a codegen bug I need to take care of (the fallthrough ```return 0``` needs to go outside the ifdefs.

In any way, thanks for the valuable input. Some things around shader-compilation seem to be much less obvious as they should be. I'll try to come up with some improvements, maybe more checks and runtime warnings.

floooh:
...sokol-shdc now defaults to not generating those ifdefs:

https://github.com/floooh/sokol-tools/blob/master/docs/sokol-shdc.md#updates

PR #409, opened 2020-10-27
Added sg_xxx_device() for retrieving metal/d3d11 device

oviano:
I've added the following two functions:

Added sg_mtl_device() for retrieving the metal device
Added sg_d3d11_device() for retrieving the d3d11 device

I have several libraries, one of which (the app framework library) creates these devices. Another library is dedicated to video playback and doesn't have direct visibility of the app framework library, but does need access to the devices for hardware accelerated decoding.

Since the precedent appears to have already been set with the sg_mtl_render_command_encoder() I have added the above two functions to the same section.


floooh:
Looks good, thanks!

PR #410, opened 2020-10-27
Allow injected d3d11 resources to be texture or srv

oviano:
I have renamed d3d11_texture to d3d11_resource in the sg_image_desc for injected d3d11 resources, and at the point this is used it checks first to see if the resource is actually a shader resource view. If it is, it uses this instead of creating its own, if not it reverts to previous behaviour and assumes it is a texture.

In my case, this is needed so I can pass a hardware decoded D3D11 texture, which is NV12 format, through sokol as two shader views that can then use an NV12 shader.

I have implemented this for 2D textures only, at this stage as I do not have code to test anything else.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -1596,7 +1596,10 @@ typedef struct sg_image_content {
 
     .gl_textures[SG_NUM_INFLIGHT_FRAMES]
     .mtl_textures[SG_NUM_INFLIGHT_FRAMES]
-    .d3d11_texture
+    .d3d11_resource
+
+    For D3D11 the d3d11_resource can be either a texture or a
+	shader resource view (2D texture only).
 
     The same rules apply as for injecting native buffers
     (see sg_buffer_desc documentation for more details).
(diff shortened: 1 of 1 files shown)


floooh:
I need to sleep over this PR... I'm don't quite like that sg_image_desc.d3d11_resource can either be a texture resource, or a texture resource view which is then type-checked, that's a bit too much "magic" ;)

Maybe it's better to have both:

```c
    void* d3d11_texture;
    void* d3d11_shader_resource_view;
```

If only one is valid, the other will be looked up or created (similar to your code), but the user can also directly provide both. That way the naming is also clearer.

...the 3D-texture path also seems to be missing (understandable, because it's pointless for your use-case).

I'll try to make up my mind, and then either use your PR as base or write something from scratch :)

oviano:
Yeah understood, separating them might be cleaner and be more understandable than hiding away some magic, I do get where you're coming from.

floooh:
Ok, I have created an alternative version:

https://github.com/floooh/sokol/commit/a49cfdc7811d90312405fdcd1c712c1e775d3b43

...and the inject-d3d11 sample has new testing code which allows to test providing a texture, or a shader-resource-view, or both:

https://github.com/floooh/sokol-samples/blob/7553b9f0d3673092af5cd7a2bfd44971b44b869e/d3d11/inject-d3d11.c#L127-L175

I'll close this PR in the meantime, let me know if the fix works for you :)

oviano:
Yes the version you creates works just fine, thanks for that :)

PR #411, opened 2020-10-27
Added gl_texture_target option for injected gl textures

oviano:
This added an additional field where the target can be specified for an injected GL texture.

I needed this so I can pass an Android GL_OES_EGL_image_external that has been provided by the mediacodec hardware video decode system through sokol and be rendered on the screen using samplerExternalOES. This is achieved by setting this new gl_texture_target field to GL_TEXTURE_EXTERNAL_OES.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -1598,6 +1598,10 @@ typedef struct sg_image_content {
     .mtl_textures[SG_NUM_INFLIGHT_FRAMES]
     .d3d11_texture
 
+    For GL, you can also specify the texture target or leave it empty
+    to use the default texture target for the image type (GL_TEXTURE_2D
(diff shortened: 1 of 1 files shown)


floooh:
Looks good, one nitpick (I'll change that during the merge): the new gl_texture_target item should be an uint32_t, because GLenum is defined as ```typedef unsigned int GLenum``` in GL headers.

floooh:
Ok, merged. Changed gl_texture_target to uint32_t, and also added a little test here:

https://github.com/floooh/sokol-samples/blob/5d68b936c307780a6ef0d7c0db1b69b21024c0d2/glfw/inject-glfw.c#L117-L128

Thanks for the PR!

PR #413, opened 2020-10-29
sokol_args.h: also parse argc/argv in emscripten build

thehink:
Basically just removed all ifdefs disabling parsing of argc/argv in emscripten build.


floooh:
Looks good to me, I'll give it a quick whirl and then merge.

floooh:
And merged. Thanks!

PR #426, opened 2020-11-16
add forward declaration in sokol_glue.h

toyboot4e:
Then we can generate FFI to `sokol_glue.h` without including other header files!
(If we define `SOKOL_GFX_INCLUDED` and `SOKOL_APP_INCLUDED`)

I mean, I'm doing this:

rokol


toyboot4e:
I'm sorry. It turns out it doesn't make sense because it makes `sapp_desc` an opaque struct. Closing

PR #429, opened 2020-11-19
Auto-generated distinct macros per lib. Resolves #428

### LINKED ISSUE #428: Distinct SOKOL_API macros for libs
This is a Windows-only feature request, as only Windows compilers have special export/import symbols for DLLs.

On Windows it's not possible to put different sokol libs in different DLLs which reference each other, as all libs use the same SOKOL_API_DECL/IMPL export macros.

My particular use case is putting imgui along with sokol_imgui in a separate dll because not all modules use/need imgui. In that dll I want sokol_imgui functions to be dllexport and other sokol functions to be dllimport.

To keep backwards compatibility this can be used:

```c++
// sokol_gfx.h
// assuming SOKOL_GFX_API is the macro for sokol_gfx
#if defined(SOKOL_API_DECL)
    #define SOKOL_GFX_API_DECL SOKOL_API_DECL
#endif

#if !defined(SOKOL_GFX_API_DECL) 
// ... 
// same defines and checks as the the ones which used to be for SOKOL_API_DECL
```

If the proposal is approved, I can whip up PRs for each library

PS I realize that a workaround by creative including is possible: `define-include-undef-define-include`, but this won't work with C++20 modules, and also it's pretty unpleasant code to write and maintain...

iboB:
This was generated by the following ruby script:

```ruby
SDECL = 'SOKOL_API_DECL'

def process_file(fname)
  lines = File.readlines(fname)

  lines.find { |line|
    line =~ /^#if.* SOKOL_([A-Z]_+)_INCLUDED/
  }
  ldecl = "SOKOL_#{$1}_API_DECL"

  decl_line = lines.index { |line|
    line =~ /^#if.* #{SDECL}/
  }

  return if !decl_line

  lines.map! { |line|
    line.
      gsub("#{SDECL}" + ' ' * (ldecl.length - SDECL.length), ldecl).
      gsub(SDECL, ldecl)
  }

  lines[decl_line, 0] = <<~EOF
    #if defined(#{SDECL}) && !defined(#{ldecl})
        #define #{ldecl} #{SDECL}
    #endif

  EOF

  File.write(fname, lines.join)
end

# process_file 'sokol_time.h'

Dir['**/*.h'].each do |f|
  process_file f
end
```

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -29,7 +29,7 @@
         SOKOL_ABORT()       - called after an unrecoverable error (default: abort())
         SOKOL_WIN32_FORCE_MAIN  - define this on Win32 to use a main() entry point instead of WinMain
         SOKOL_NO_ENTRY      - define this if sokol_app.h shouldn't "hijack" the main() function
-        SOKOL_API_DECL      - public function declaration prefix (default: extern)
+        SOKOL_APP_API_DECL  - public function declaration prefix (default: extern)
         SOKOL_API_IMPL      - public function implementation prefix (default: -)
         SOKOL_CALLOC        - your own calloc function (default: calloc(n, s))
         SOKOL_FREE          - your own free function (default: free(p))
@@ -44,7 +44,7 @@
 
         SOKOL_DLL
 
-    On Windows, SOKOL_DLL will define SOKOL_API_DECL as __declspec(dllexport)
+    On Windows, SOKOL_DLL will define SOKOL_APP_API_DECL as __declspec(dllexport)
     or __declspec(dllimport) as needed.
 
     If you use sokol_app.h together with sokol_gfx.h, include both headers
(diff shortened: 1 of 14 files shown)


iboB:
oops... forgot to mark this as a draft. Sorry about that

This is not ready to merge, yet. I saw some problems from the script. If all checks pass, I'll fix the problems manually

iboB:
Ok. All checks have passed. I fixed the bad auto-generation of sokol_gfx_imgui.

@floooh I think this is ready for merge

floooh:
Ok, I'm merging this now (manually, as I usually do). I wonder if the macro is actually needed inside the SOKOL_IMPL block too, I'll experiment with this on Windows a bit.

floooh:
..and merged, many thanks!

PR #432, opened 2020-11-22
docs: fix simple typo, simultanously -> simultaneously

timgates42:
There is a small typo in sokol_gfx.h.

Should read `simultaneously` rather than `simultanously`.


floooh:
Thanks! Note that the WebGPU currently isn't working though because it hasn't been updated for API WebGPU changes that happened in the meantime :)

PR #433, opened 2020-11-22
sokol_audio: AVAudioSession fixes for iOS

oviano:
This PR fixes two issues I have discovered on iOS:

1. With the iPhone audio only works if the AVAudioSession is explicitly activated on audio initialisation, and deactivated on shutdown so it now does this on iOS (it doesn't appear to affect anything by doing this on an iPad).

2. When there is an interruption that requires the audio to stop, then the AudioQueue no longer continues after the interruption is finished. i.e. the audio callback is no longer called. This can be observed by setting Alarm Clock to go off a minute later, and then running an app that plays audio. The fix is to explicitly Pause/Start the AudioQueue and deactive/activate the AVAudioSession in response to AVAudioSessionInterruptionNotification.


floooh:
I'll start looking into this now.

floooh:
I'll apply some fixes during the merge, this might take a few days. I'm just keeping some reminder notes here from time to time:

- need to handle ARC on vs off for the interruption handler object (like in sokol_app.h: https://github.com/floooh/sokol/blob/c8023ab515822005d74c36b3a77656315a5ce1be/sokol_app.h#L2648-L2652)

floooh:
Ok, I fixed the ARC vs non-ARC stuff and got rid of the bridged casts (ObjC id's in C structs are supported with ARC enabled since Xcode 10.2 or so, I use the same stuff in sokol_app.h and sokol_gfx.h). This stuff is currently in a merge-branch (https://github.com/floooh/sokol/tree/oviano-sokol-audio-avaudiosession).

Also tested on iPad device and iPhone simulator (I don't have a physical iPhone anymore). Actual merge will be tomorrow though.

oviano:
> Ok, I fixed the ARC vs non-ARC stuff and got rid of the bridged casts (ObjC id's in C structs are supported with ARC enabled since Xcode 10.2 or so, I use the same stuff in sokol_app.h and sokol_gfx.h). This stuff is currently in a merge-branch (https://github.com/floooh/sokol/tree/oviano-sokol-audio-avaudiosession).
> 
> Also tested on iPad device and iPhone simulator (I don't have a physical iPhone anymore). Merge will most likely be tomorrow.

Great, thank you!

floooh:
...and merged. Thanks for the PR!

I haven't updated the changelog yet, will do so now.

PR #435, opened 2020-11-25
Added sg_dealloc_xxx() and sg_uninit_xxx() suite of functions

oviano:
These are the equivalent to the sg_alloc/init functions and allow async destruction of resources. This is useful when using sokol in a multithreaded scenario; the render thread does the sg_init/sg_uninit while a separate resource/api/update thread can call the sg_alloc/sg_dealloc.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -2272,11 +2272,21 @@ SOKOL_API_DECL sg_image sg_alloc_image(void);
 SOKOL_API_DECL sg_shader sg_alloc_shader(void);
 SOKOL_API_DECL sg_pipeline sg_alloc_pipeline(void);
 SOKOL_API_DECL sg_pass sg_alloc_pass(void);
+SOKOL_API_DECL void sg_dealloc_buffer(sg_buffer buf_id);
+SOKOL_API_DECL void sg_dealloc_image(sg_image img_id);
+SOKOL_API_DECL void sg_dealloc_shader(sg_shader shd_id);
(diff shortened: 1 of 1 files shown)


floooh:
Looks good! I'll try to do a "merge day" soon where I'll focus on pending PRs.

floooh:
I'm starting to look into this PR now. I think there's two small problem:

- the uninit-functions should put the resource slot back into the SG_RESOURCESTATE_ALLOC state
- ...and the dealloc-functions need to check whether the resource slot in in the SG_RESOURCESTATE_ALLOC state

I think those changes are trivial, and I'll try to do them during the merge.

oviano:
Yes, you are right of course, I remember wondering about that when I did it, and I think the only reason I didn't at the time was that in all cases the uninit should be followed by a dealloc anyway so it seemed like it wasn't too critical for my testing.....and then I forgot to go back to it :)

floooh:
Ok merged, many thanks for the PR!

As described above, I've added the correct "state unwinding". I've also added a couple new tests to sokol-gfx-test.c in the sokol-samples repository:

https://github.com/floooh/sokol-samples/blob/4cf2a1ef723207e6d7a9b1500f79eb048a2f4ce6/tests/sokol-gfx-test.c#L674-L742

Another minor thing you missed is adding new entries to the sg_trace_hooks struct (this only shows up when compiling with SOKOL_TRACE_HOOKS enabled).

oviano:
Great, thank you and also if there is a next time I'll be sure to look out for the sg_trace_hooks!

Issue #450, opened 2021-01-03
sokol_gfx: MRT pixel_formats ?


septag:
I have noticed that we can't create multiple render target passes with different pixel_formats, sokol validates that every RT must have the same pixel_format.

however, this should be valid in d3d11/GL4/Metal APIs. in fact, this is very much necessary for GBuffer creation. (ARGB for albedo, R16G16F for normals, etc..)

Is there a reason that this check is done in `make_pass` ?

floooh:
See here:

https://github.com/floooh/sokol/issues/402#issuecomment-706687941

Definitely something that should be fixed, it will also need an extension of the sg_pipeline_desc struct though.

There may be restrictions in backend-APIs what pixel format combinations are allowed though, so before relaxing the requirement we should have another careful look at the 3D API docs and maybe allow backend-specific exceptions via the sg_features struct.

septag:
oh. alright!
The limitation is probably from ES 2.0. I'll check it out and perhaps submit a PR

0x1100:
OpenGL ES 2.0 and WebGL1 can only have a single color attachment.
See the "mrt" (multi-rendertargets) sample.

floooh:
> OpenGL ES 2.0 and WebGL1 can only have a single color attachment.

True, but general MRT support is already covered by the sg_features.multiple_render_targets field:

https://github.com/floooh/sokol/blob/5894182dddc2e0fd6531e5a2deb5d9cf7662ea2a/sokol_gfx.h#L808

The question is whether we'll need additional feature flags for pixel format restrictions.

septag:
MRT can be enabled on certain ES2 hardware with EXT_draw_buffers extension. but the limitation of same pixel_format for all RTs would always be there for ES2

floooh:
This is now implemented.

Issue #402, opened 2020-10-10
sokol_gfx.h: Render Targets with Different Pixel Format


MrMilad:
I tried to attach two color_attachment with different pixel formats (SG_PIXELFORMAT_RGBA16F and SG_PIXELFORMAT_RGBA8) in the same pass, but i get this error: 
**all pass color attachment images must have the same pixel format**.

Please allow the use of render targets with different pixel formats in the same pass.
According to the documents, this is possible for D3D11 and same for OpenGL:
> Each render target can have a different data format. These render target formats are not required to have identical bit-per-element counts.

floooh:
I'm don't quite remember why I added this restriction, it might have been related to WebGL2 (quickly browsing the spec I can't find the restriction though), or simply because ```sg_pipeline_desc``` only has a single ```color_format``` item. Apart from the pixel format, there's definitely further differences between 3D APIs though, e.g. whether alpha-blending can be set up per attachment or must be the same across all attachments.

This definitely needs a public API braking change and some research and experimentation, so it's not just removing that validation check

MrMilad:
Thanks for the information. So can I comment out that line of code as a quick and dirty way?
`SOKOL_VALIDATE(img->cmn.pixel_format == color_fmt, _SG_VALIDATE_PASSDESC_COLOR_PIXELFORMATS);`
Although I know this is wrong, I have no choice :)

floooh:
> So can I comment out that line of code as a quick and dirty way

It might work on GL and D3D11, but definitely not on Metal, because this requires pixel format information in the pipeline state object for each color attachment:

See here: https://developer.apple.com/documentation/metal/mtlrenderpipelinedescriptor?language=objc

...and search for MTLRenderPipelineColorAttachmentDescriptorArray. in the current sg_make_pipeline() implementation this array has identical entries for each attachment:

https://github.com/floooh/sokol/blob/cb50040aab19ef299117742f983bb2f99283ef46/sokol_gfx.h#L9827-L9838

floooh:
This is now possible with the latest API update.

PR #452, opened 2021-01-04
Improved MRT support. Removed same pixel_format limitation for MRT

septag:
Regarding the issue #450 and #402 , I've created a draft version. I tested on GL and D3d. but haven't tested on other platforms yet. 
I'll wait for your feedback. some notes:

- I think the limitation was for OpenGL-ES2. however, we should check for **EXT_draw_buffers** extension on ES2: https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_draw_buffers.txt . On all other APIs, there should be proper support with different formats for MRT
- question number 5 of the extension text indicates that on ES2, the pixel_format must be the same. so I only put the check for ES2 with MRT support.
- I had to change the API for sg_make_pipeline, so that we get multiple color_formats in `sg_blend_state.color_formats` instead of one.  

EDIT: I just realized that I can put a union around `color_formats` and `color_format` to not break the API, so we would have both fields depending on what user wants to use. what's ur opinion on that ?

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -1838,7 +1838,7 @@ typedef struct sg_blend_state {
     sg_blend_op op_alpha;
     uint8_t color_write_mask;
     int color_attachment_count;
-    sg_pixel_format color_format;
+    sg_pixel_format color_formats[SG_MAX_COLOR_ATTACHMENTS];
     sg_pixel_format depth_format;
     float blend_color[4];
 } sg_blend_state;
@@ -2061,7 +2061,7 @@ typedef struct sg_pass_info {
     .uniform_buffer_size    4 MB (4*1024*1024)
     .staging_buffer_size    8 MB (8*1024*1024)
 
-    .context.color_format: default value depends on selected backend:
+    .context.color_formats[SG_MAX_COLOR_ATTACHMENTS]: default values depend on selected backend:
         all GL backends:    SG_PIXELFORMAT_RGBA8
         Metal and D3D11:    SG_PIXELFORMAT_BGRA8
         WGPU:               *no default* (must be queried from WGPU swapchain)
@@ -2528,6 +2528,13 @@ inline void sg_init_pass(sg_pass pass_id, const sg_pass_desc& desc) { return sg_
     #ifndef GL_LUMINANCE
     #define GL_LUMINANCE 0x1909
     #endif
+    #ifndef GL_MAX_COLOR_ATTACHMENTS
+    #define GL_MAX_COLOR_ATTACHMENTS 0x8CDF
+    #endif
+    #ifndef GL_COLOR_ATTACHMENT0
+    #define COLOR_ATTACHMENT0 0x8CE0
+    #endif
+
 
     #ifdef SOKOL_GLES2
     #   ifdef GL_ANGLE_instanced_arrays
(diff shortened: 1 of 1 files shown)


floooh:
I haven't looked at the PR in detail yet (I'll probably do so after my vacation), but this:

> EDIT: I just realized that I can put a union around color_formats and color_format to not break the API, so we would have both fields depending on what user wants to use. what's ur opinion on that ?

Please no unions I just removed all nested unions because they're tricky to handle in language bindings :) Breaking the API is fine for situations like this (I'll do a quite massive API breakage after I merge this branch anyway where I'm overhauling the sokol-gfx and -app APIs to be more "language-binding-friendly": https://github.com/floooh/sokol/tree/bindgen-friendly). I guess it's a good idea to put other breaking changes into the same big merge (like your PR, and this sokol-shdc change: https://github.com/floooh/sokol-tools/issues/45)

septag:
yes. I noticed this yesterday, as I was syncing my own fork with your repo :)

floooh:
I'll start working on this now in the bindgen-friendly branch, however I decided to do a deeper change and allow to define the entire color-blend-state per color-attachment.

For instance see here:

https://docs.microsoft.com/en-us/windows/win32/api/d3d11/ns-d3d11-d3d11_blend_desc

...and here:

https://developer.apple.com/documentation/metal/mtlrenderpipelinecolorattachmentdescriptor?language=objc

floooh:
One limitation for GL: even though the new sg_pipeline_desc structure will allow setting a different blend-configuration and color write mask per color attachment, the GL backend will only support a single blend-configuration across all attachments (for now at least).

Separate blend-config for color attachments requires either GL4 or the ARB_draw_buffers_blend extension:

https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_draw_buffers_blend.txt

...separate pixel formats are independent from this and should work though.

septag:
That is really cool. looking fwd for your changes.
btw, regarding GL, don't u think it's a good idea to migrate to OpenGL4.2 or 4.4 ?

floooh:
I've actually been thinking moving to GL 4.0 or 4.1, that's the last version macOS supports, and GL 4.0 is the first version which allows to define per-render-target blend-functions (the "i" version here: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBlendFuncSeparate.xhtml). The sg_pipeline_desc update will allow to define per-render-target blend-funcs, but currently this will only be supported in the D3D11 and Metal backend.

Different pixel formats seems to work across all APIs that support multiple-render-targets though (all except GLES2/WebGL1). I found what looks like a bug in Metal for macOS though:

https://twitter.com/FlohOfWoe/status/1353338461235924994

Per-render-target color masks will work on all desktop backends (including GL), but not on WebGL2 (WebGL2 is missing the glColorMaski() function:

https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glColorMask.xhtml

...and per-render-target BlendFuncs will currently only support in D3D11 and Metal, with the option GL4.x too (but I'd like to do any GL4 upgrade later, because that would probably also enable other modern features in the GL backend).

floooh:
Erm, sorry for writing the same stuff I already wrote before... side effect of switching between too many tasks currently I guess :D

floooh:
Ok, closing this because it's redundant with the latest API update now.

PR #456, opened 2021-01-09
sokol_app.h: trigger a linker error if users forget -pthread

### LINKED ISSUE #455: [sokol_app] Suggestion: add a pthread call to Linux init so there's an error instead of "weird things"
I just experienced the "weird things" like the ones in #376 because I had forgotten to add `-pthread`. I lost about half an hour until I finally figured it out. More precisely with CMake I had added `${CMAKE_THREAD_LIBS_INIT}` but I had forgotten to call `find_package(Threads)`, which is sort of common.

To prevent this from happening, why not add a simple "noop" pthread call like `pthread_yield` in the Linux initialization?

Thus users which forget about `pthread` one way or the other would at least get some kind of compilation or linker error, instead of experiencing #376 and would find the problem much faster.

iboB:
My suggestion to resolve #455

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -10030,7 +10030,13 @@ _SOKOL_PRIVATE void _sapp_x11_process_event(XEvent* event) {
     }
 }
 
+extern int pthread_yield(void); // see below
 _SOKOL_PRIVATE void _sapp_linux_run(const sapp_desc* desc) {
+    // The following line is here to trigger a linker error instead of an
+    // obscure runtime error if the user has forgotten to add -pthread to
+    // the compiler or linker options. It has no other purpose.
+    pthread_yield();
+
     _sapp_init_state(desc);
     _sapp.x11.window_state = NormalState;
 


floooh:
Looks like  good solution, thanks!

floooh:
There's an error when compiled as C++:

```
/usr/include/pthread.h:470:12: error: conflicting declaration of ‘int pthread_yield()’ with ‘C’ linkage
 extern int pthread_yield (void) __THROW;
            ^~~~~~~~~~~~~
```

I'll include pthread.h instead and removed the manual declaration.

floooh:
hmmmm it's not that simple... pthread.h doesn't contain pthread_yield() by default, instead _GNU_SOURCE must be defined before, but if there's already another include of pthread.h the _GNU_SOURCE define doesn't work. What a mess :/ 

My next attempt was calling pthread_self(), but for some reason this seems to be optimized away...

iboB:
`pthread_self` (funny enough) doesn't actually require pthread to be linked. It's in libc. 

And yeah, _GNU_SOURCE was the reason I just forward declared `pthread_yield` instead of including pthread.

As for C++, that's unpleasant. How about we keep the `__THROW` for now and in a couple of hours I'll find some time to submit a better solution in a PR?

iboB:
Er... by the way won't adding `"C"` fix things properly? C functions should be nothrow

`extern "C" int pthread_yield();`

floooh:
No worries, I found an okay-ish solution: pthread_attr_init/destroy works. It shows up in libsokol.a as external undefined symbol (unlike pthread_self()).

floooh:
> won't adding "C" fix things properly

Not sure, there's a lot of C++ specific stuff in /usr/include/pthread.h, and almost all functions are marked as __THROW.

floooh:
Ok, see: https://github.com/floooh/sokol/commit/5fbb6a25501e5478674ca0600a7539033592d749

Issue #461, opened 2021-01-21
sokol_fetch: range requests are off by 1 in wasm


jfaust-fy:
According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests, range requests are inclusive of the values specified. 

sokol_fetch does this:
```
req.setRequestHeader('Range', 'bytes='+offset+'-'+(offset+bytes_to_read));
```

And ends up receiving overlapping range requests:
```
[sent] Range: bytes=0-8192
[received] Content-Length: 8193
[received] Content-Range: bytes 0-8192/46128

[sent] Range: bytes=8192-16384
[received] Content-Length: 8193
[received] Content-Range: bytes 8192-16384/46128

[sent] Range: bytes=16384-24576
[received] Content-Length: 8193
[received] Content-Range: bytes 16384-24576/46128
...
```

This means each chunk in a chunked read duplicates 1 byte from the previous chunk.

This is an easy fix, and I'll put up a PR shortly:
```
req.setRequestHeader('Range', 'bytes='+offset+'-'+(offset+bytes_to_read-1));
```

floooh:
Hmm interesting, thanks for catching this. I wonder why the plmpeg samples works now (I'll do some tests when merging the PR, I'll try to do this soon-ish).

jfaust-fy:
I took a quick look at the plmpeg sample, and I think I see the difference. In the plmpeg sample, you're managing a bunch of buffers, and they all have extra space to account for compression. So there's an extra byte being written, but it doesn't matter - you never even check the size of the data returned, because you don't need to.

Whereas I need contiguous memory, so I'm appending to a single buffer (and expanding it as necessary).

floooh:
https://github.com/floooh/sokol/pull/462 has been merged, thanks!

PR #462, opened 2021-01-21
Fix range request to not overlap reads

jfaust-fy:
See #461


floooh:
And merged, thanks a lot for the PR!

PR #470, opened 2021-01-29
fix MSVC warnings

code-disaster:
Fixes C4191 ("'type cast': unsafe conversion from 'FARPROC' to '...'") when compiled with Visual C++ /W3 or above. The warning appears to be C++ only, so the sokol samples don't complain.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -5322,15 +5322,15 @@ _SOKOL_PRIVATE void _sapp_wgl_init(void) {
         _sapp_fail("Failed to load opengl32.dll\n");
     }
     SOKOL_ASSERT(_sapp.wgl.opengl32);
-    _sapp.wgl.CreateContext = (PFN_wglCreateContext) GetProcAddress(_sapp.wgl.opengl32, "wglCreateContext");
+    _sapp.wgl.CreateContext = (PFN_wglCreateContext)(void*) GetProcAddress(_sapp.wgl.opengl32, "wglCreateContext");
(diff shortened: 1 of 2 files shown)


floooh:
Thanks!

floooh:
PS: strange that I'm not seeing those warnings when compiling the sokol-samples here in C++ mode with warning level 4:

https://github.com/floooh/sokol-samples/blob/89f5825ab5d36906b2dc219efbce3d00be9e7cdf/libs/sokol/CMakeLists.txt#L39-L48

I'm having a deja-vu btw, didn't we discuss this exact same thing a little while ago? :D

code-disaster:
Yes, on Twitter about a week ago. :D

Hm, indeed, doesn't warn in sokol-samples. The only difference I see is that I run my code with /Wall, some nitpicky stuff disabled globally, and external headers wrapped in `#pragma warning(push: 3)`. Seems like MSVC does handle this differently.

floooh:
...strange... on AppVeyor (VS2015) I'm now seeing this:

```
C:\projects\sokol\sokol_gfx.h(8131): warning C4054: 'type cast': from function pointer 'FARPROC' to data pointer 'void *'
```
...this is the double cast here:

https://github.com/floooh/sokol/blob/9dfe8cd050d7a69deff1f74c3bfd4d1ba0c18cd4/sokol_gfx.h#L8131

...I'll just go ahead and suppress this warning via pragma, makes no sense to add changes just to appease VS2015 I think ;)

PR #473, opened 2021-02-01
Added sokol_nuklear.h utility header

wmerrifield:
sokol_nuklear.h implements the initialisation, rendering and event-handling code for [Nuklear](https://github.com/Immediate-Mode-UI/Nuklear) on top of sokol_gfx.h and (optionally) sokol_app.h. The design of it is heavily based on the existing sokol_imgui.h header. 

See also this PR for the example in the sokol-samples repo: https://github.com/floooh/sokol-samples/pull/81


floooh:
FYI: I'm going to look into this PR after the "API breaking merge" I'll do sometime this week (probably tomorrow). I'll need to fix a few things because of the API changes, but I can do this during the merge.

floooh:
Ok, I'm starting to look into this now.

floooh:
BTW I just saw this:

https://github.com/floooh/sokol/blob/9d518545fcdd6e775f6fe5ff35e099bfd2c4d8fd/util/sokol_nuklear.h#L1662

Note that SOKOL_ASSERT() will be compiled out in release builds (or more specifically, when NDEBUG is defined), so putting logic into SOKOL_ASSERT() isn't a good idea (I'll fix that stuff during the merge - I'll also try updating to the latest Nuklear, and put the header right into sokol-samples/libs)

floooh:
I'm going to use the sample code from here:

https://github.com/Immediate-Mode-UI/Nuklear/blob/master/demo/overview.c

This demonstrates more Nuklear features similar to the ImGui samples.

floooh:
...I also added an overflow-check for the vertex- and index-data:

https://github.com/floooh/sokol/blob/1ddb55b177ab64667d389715717fa15c4170b962/util/sokol_nuklear.h#L1899-L1905

This is also caught by the sokol_gfx.h validation layer, but catching this higher up is a bit cleaner.

floooh:
Oh, and I added a preprocessor check and a comment that the implementation must be compiled in C, this is a bit special because most other headers can be either compiled in C or C++, but I think for Nuklear it makes sense, and the C99 designated initialization makes it "worth it" (I'm actually thinking about moving some other header-implementations from the common C/C++ subset to C99 as well).

floooh:
...and merged. Thanks! I'll update the CHANGELOG and add the sample to the sample webpage next.

wmerrifield:
Awesome. I’m glad it’s in (and thank you for Sokol in general!)

PR #474, opened 2021-02-08
Fix #pragma compilation with tcc on windows

spaceface777:
The tcc ([Tiny C Compiler](https://bellard.org/tcc/)) linker does not like the `.lib` suffix inside `#pragma comment (lib, "foo.lib")` directives - it fails with `tcc: error: library 'foo.lib' not found` errors. Removing the extension fixes tcc, and all major compilers (msvc/clang/gcc/mingw) compile fine without it, so this change should simply allow the headers to be compiled with a different compiler.

We've needed to, (and have been making) this change over at [vlang/v](https://github.com/vlang/v/tree/master/thirdparty/sokol),  ever since tcc became an officially-supported (and later on, the default for debug builds) C compiler on windows. However, every time we update the headers, we also need to reapply this patch; and we felt that, by PR'ing it upstream, it'd benefit both the V project, and anyone attempting to compile Sokol with tcc :)

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1455,16 +1455,16 @@ inline void sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
         #pragma comment (linker, "/subsystem:windows")
     #endif
 
-    #pragma comment (lib, "kernel32.lib")
-    #pragma comment (lib, "user32.lib")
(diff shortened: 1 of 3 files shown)


floooh:
Thanks! I hadn't been aware that it also works without the extension.

PR #475, opened 2021-02-11
Android IME support

miscvariables:
based on checkout of sokol and sokol-samples on 2021/02/11
Android ImGui IME support. Simple Information:

    No Java. Pure c/c++ android native activity.

    Using JNI to call JavaVM functions, show/hide soft keyboard, create Android EditText widget in top view DecorView

    Update the content between EditText and ImGui::InputText

    I use 4 states to update the content: previous text in EditText(atext_prev), current text in EditText(atext), previous text in ImGui::InputText(vtext_prev), current text in ImGui::InputText(vtext). "atext" means Android-Text, "vtext" means Virtual-Text. The state machine is in sokol_app.h

    Android main UI thread and sokol gl&imgui thread use Android looper to communicate. All JNI calls are done in Android main UI thread. It can avoid crash

    sokol_imgui.h should pass ImGui::InputText state of current focus to sokol_app.h

    sokol android seems only using c interface(no c++), so I only implement JNI c interface in sokol_app.h, JNI c++ interface has only empty functions for compiling complains.

    Pure c/c++ Android app must use many tricks, the the code looks a bit dirty. There are many commented out codes in sokol_app.h, show the sucessful and unsucessful tries I ever made.

    Patched files:
    sokol/sokol_app.h
    sokol/util/sokol_imgui.h
    sokol-samples/sapp/imgui-highdpi-sapp.cc
    sokol-samples/sapp/CMakeLists.txt (stripped. Contains only imgui-highdpi-sapp.cc Not necessary)
    sokol-samples/libs/dbgui/dbgui.cc (only add #include imgui_internal.h, for compiling complains)
    sokol-samples/tests/cpp-compile-test.cc (only add #include imgui_internal.h, for compiling complains)


miscvariables:
f21bce2a97fb790e023ae24d1b8ee2c

7859e4798e72763da44eb95eb1ce17d

miscvariables:
Bugfix and better performance in sokol_app.h, no known bugs now

floooh:
Thanks for the PR, but TBH, currently it looks too messy to be considered. There are big code blocks inside```#if 0 / #endif```,  an strdup() call which most likely leads to a memory leak (in any case it circumvents the SOKOL_MALLOC allocation function), large blocks of code are commented out, and function names like ```UnicodeStrFromUtf8``` don't follow the naming convention and might collide with other symbols.

miscvariables:
I am cleaning up the code, so many tricks when only using c, some comments are reserved for alternative options. coming soon

floooh:
> I am cleaning up the code

Ah ok I see. You can set the Pull Request to "Draft" state as long as you're working on it:

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft

That way I also can't accidentally hit the "Merge" button ;)

floooh:
PS:

> Is it better to open an issue for test before merge?

No, I think it's better to discuss the PR right here.

PR #476, opened 2021-02-13
Enable UTF-8 console output for `/subsystem:windows`

garettbass:
This change attaches stdin/out/err to the console of the parent process so that `/subsystem:windows` applications can still write to the console when executed from a command line.  Additionally, it sets the console output to UTF-8 to enable unicode output at the console.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -6349,6 +6349,12 @@ int WINAPI WinMain(_In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance, _
     _SOKOL_UNUSED(hPrevInstance);
     _SOKOL_UNUSED(lpCmdLine);
     _SOKOL_UNUSED(nCmdShow);
+    if (AttachConsole(ATTACH_PARENT_PROCESS)) {
+        freopen("CON", "r", stdin);
+        freopen("CON", "w", stdout);
+        freopen("CON", "w", stderr);
+    }
+    SetConsoleOutputCP(CP_UTF8);
     int argc_utf8 = 0;
     char** argv_utf8 = _sapp_win32_command_line_to_utf8_argv(GetCommandLineW(), &argc_utf8);
     sapp_desc desc = sokol_main(argc_utf8, argv_utf8);
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -6339,6 +6339,12 @@ _SOKOL_PRIVATE char** _sapp_win32_command_line_to_utf8_argv(LPWSTR w_command_lin
 #if !defined(SOKOL_NO_ENTRY)
 #if defined(SOKOL_WIN32_FORCE_MAIN)
 int main(int argc, char* argv[]) {
+    if (AttachConsole(ATTACH_PARENT_PROCESS)) {
+        freopen("CON", "r", stdin);
+        freopen("CON", "w", stdout);
+        freopen("CON", "w", stderr);
+    }
+    SetConsoleOutputCP(CP_UTF8);
     sapp_desc desc = sokol_main(argc, argv);
     _sapp_win32_run(&desc);
     return 0;


garettbass:
This is causing compiler errors in some situations, still working on it.  Of course windows wants me to use freopen_s() instead of freopen(), and when I run the bindgen scripts clang says freopen() is undeclared :P

floooh:
I'll look into this next.

floooh:
@garettbass, what's the reason for the second commit  where AttachProcess() is also called for SOKOL_WIN32_FORCE_MAIN?

Arent /subsystem:windows and "int main() {}" mutually excluse?

I'll also do two fixes:

- include <stdio.h> for freopen
- and VS warns that freopen is deprecated, I first considered using freopen_s() since the code is "Windows specific" anyway, but I actually don't know if gcc and clang on Windows support freopen_s(), so I guess I'll just supress the warning.

floooh:
I was hoping that the redirected stdout/stderr shows up in Visual Studio's debugger console, but it's silent. So to see log output during debugging one would still need to redefine SOKOL_LOG() as OutputDebugString(). But OutputDebugString() in turn only shows up on the debugging console, not on stdout or stderr.

Hmm...

floooh:
I'm wondering including stdin is useful... because when starting a subsystem:windows application on the terminal, the terminal doesn't block but instead continues working as a regular command line. So I wonder what's would happen if the application would read from stdin, would it get all the random command input from the terminal window?

floooh:
Hmm... the freopen() seems to add about 9 KBytes in Release mode to executables (e.g. clear-sapp goes from 138 KBytes to 147 KBytes). I was actually sorta proud that there are no stdio.h calls in sokol-app and sokol-gfx ;)

floooh:
...ok here's a plan:

- Remove the whole ```AttachConsole()``` block for the SOKOL_WIN32_FORCE_MAIN case, unless there's a good reason for it
- Remove ```freopen(stdin)``` unless there's a good reason for it
- Add a runtime config flag ```sapp_desc.win32_attach_console``` which is false by default. With LTO/LTCG enabled, MSVC "sees" across the function call and doesn't add the stdio code, saving those 9 KBytes if the application doesn't need console output for /subsystem:windows applications.

I just checked this, and surprisingly it works, this:

```c
sapp_desc sokol_main(int argc, char* argv[]) {
    return (sapp_desc){
        ....
        .win32_attach_console = true,
    };
}
```
...produces a 147 KByte exe, while this:

```c
sapp_desc sokol_main(int argc, char* argv[]) {
    return (sapp_desc){
        ....
        .win32_attach_console = false,
    };
}
```
...produces a 136 KByte exe.

Does that plan make sense?

garettbass:
Thanks for the research.  I hadn’t considered the size.  I added the same code in both entry points because I always use FORCE_MAIN on windows to ensure there isn’t an unused main entry point generated by the Nim compiler.  I like the opt-in behavior you added.

garettbass:
I will investigate whether there is a non-stdio equivalent to freopen in kernel32 that could be used instead.

garettbass:
Also, if you would prefer not to add this feature, I can actually do it outside sokol, before main/WinMain, by putting a function pointer in the static initialization section.

garettbass:
I found an freopen_s [example on stackoverflow](https://stackoverflow.com/questions/57210117/unable-to-write-to-allocconsole) in case you prefer that to suppressing the warning.
```c
FILE *fDummy;
freopen_s(&fDummy, "CONIN$", "r", stdin);
freopen_s(&fDummy, "CONOUT$", "w", stderr);
freopen_s(&fDummy, "CONOUT$", "w", stdout);
```

floooh:
> Also, if you would prefer not to add this feature, I can actually do it outside sokol, before main/WinMain, by putting a function pointer in the static initialization section.

I've slept over it and I think it makes sense to add it, because the missing stdout output in Windows apps is an eternal hassle for printf-debugging.

I think I'm going to add a second "mode" which creates a console window instead of attaching to the parent process, this should help when working in the Visual Studio debugger. Or maybe a mode which first tries to attach to the parent console, and if this fails, create an output window. I'll tinker around a bit with this.

Regarding freopen_s(): I had actually tried that but then wasn't sure if this is also supported by non-Visual-Studio clibs, like mingw, so what I did instead is to use freopen() and just suppress the warning.

I'll continue with this stuff in the evening :)

floooh:
Ok, this is how what I ended up with:

Three new bools in sapp_desc:

- **win32_console_attach**: attach console output to parent process (same as your code). This displays console output when starting in a terminal window, but doesn't show anything when double-clicked. This might be useful for release mode to show printf() logging output only when started in a terminal, but look like a regular Windows UI application when double clicked.
- **win32_console_create**: this creates a new console window via AllocConsole() instead of attaching to the parent process. This might be useful to show printf-debugging-output during development, especially when running in a debugger.
- **win32_console_utf8**: set the console codepage to UTF-8

I've also moved that code into a new function ```_sapp_win32_init_console()```, and call that from ```_sapp_win32_run()``` instead of directly from main() or WinMain(). That way it also works with SOKOL_NO_ENTRY (which is especially useful for language bindings).

I'm also restoring the original codepage on exit (without this the terminal remains in UTF-8 mode).

Some screenshots:

win32_console_attach = true, but win32_console_utf8=false:
image

Same with win32_console_utf8=true:
image

win32_console_create = true will always open a new console window, no matter if started from the command line or desktop:

image

If we need "smarter" behaviour, we can still add more config flags, or change the behaviour of the existing flags (I've been thinking about a "smart mode" where console output goes to the terminal when started in a terminal, creates a console window when running in a debugger, and doesn't show output when started otherwise.

Oh, and I also removed stdin for now. If there's a situation where having stdin is really useful for a windowed application we can still add it back.

floooh:
..hm one (unrelated) thing is unexpected. If I change one of the sapp-samples to compile as /subsystem:console and replace the WinMain with a regular main function, stdout output still doesn't appear either in the parent terminal or separate console window. 

It works as expected in the GLFW samples though.

And my sokol-zig examples also open a console window.

I've been thinking that, while I'm at it, I could add a config option to prevent that console window from opening for /subsystem:console programs, but I need to investigate this first (most likely when I'm working on the Zig bindings again).

I'll go ahead and merge the PR now.

floooh:
Hmm how strange, I just merged the PR via the command line workflow, but github didn't detect it this time. Usually github marks the PR as merged when I push the merged version. Didn't we have some similar odd behaviour before maybe related to your github user name or account?

I'm going to close the PR, but as I said, it's been merged :)

floooh:
...or maybe because the PR was from your master branch? I'm not sure...

garettbass:
Thanks for taking this on.  I wish I had some answers to those tricky questions.

Issue #479, opened 2021-02-15
sokol_app.h currently doesn't compile on macOS 10.12


floooh:
Not super high priority, but apparently the function NSPasteboardTypeFileURL wasn't available yet in macOS 10.12.

Since this is the only problematic function it might make sense to find a workaround.

See: https://twitter.com/tumult/status/1361205135376965636

randrew:
It's a constant, not a function :)

Use 1: https://github.com/floooh/sokol/blob/8dd4e4a0ea07cde21308df43d66dedbf388be0dc/sokol_app.h#L3159
Uses 2 and 3: https://github.com/floooh/sokol/blob/8dd4e4a0ea07cde21308df43d66dedbf388be0dc/sokol_app.h#L3174 https://github.com/floooh/sokol/blob/8dd4e4a0ea07cde21308df43d66dedbf388be0dc/sokol_app.h#L3179

I haven't written Cocoa in a long time, but I'm pretty sure there is an equivalent way to do this on older macOS/OS X versions. I'm guessing it will probably marked deprecated for newer macOS versions.

floooh:
Closing this because https://github.com/floooh/sokol/pull/482 has been merged.

PR #482, opened 2021-02-19
Fix sokol_app.h not building on macOS 10.12

randrew:
This addresses #479 enough to get sokol_app.h building and working on macOS 10.12, but it simply disables the drag and drop functionality instead of reimplementing it. It uses a preprocessor guard that checks for 10.13 or greater.


floooh:
Ok merged, thanks! I think it's ok to not have a fallback path for 10.12 and earlier. In case we need this, GLFW seems to have implemented a workaround for the same issue here: https://github.com/glfw/glfw/issues/1377

Issue #415, opened 2020-10-31
Reducing mouse+display lag investigations (mainly for Dear ImGui apps)


floooh:
Sort of a reminder ticket. See here:

https://github.com/floooh/cimgui-sokol-starterkit/issues/2

And this also overlaps with the "only render on input" mode for sokol_app.h (because at least on Windows switching off vsync and rendering from within WM_MOUSEMOVE already makes a dramatic difference).

Also see:

https://github.com/floooh/sokol/issues/341

https://github.com/floooh/sokol/issues/301

https://github.com/floooh/sokol/issues/292

https://github.com/floooh/sokol/issues/248

...these are all more or less related. An idea would be to add a couple different "update modes" to sokol_app.h which would tune the swapchain setup and the place where the frame callback is called for different scenarios. Alternatively instead of modes, add a number of sapp_desc options to tweak unique things instead of grouping them into "modes".

floooh:
...probably not an option but:

https://github.com/nlguillemot/LateLatching

floooh:
...another information tidbit:

https://docs.microsoft.com/en-us/windows/uwp/gaming/reduce-latency-with-dxgi-1-3-swap-chains

slmjkdbtl:
not sure if really related but I fixed #341 by changing the macOS timer based loop to manual input polling loop [like this](https://github.com/slmjkdbtl/dirty/blob/master/ext/inc/sokol/sokol_app.h#L2794) (borrowed code from glfw, all changes are marked by `// CHANGE`), haven't tested if it breaks metal so haven't submitted a PR

cmp102:
I'm struggling with the same issue with linux/opengl, i did this quick experiment to reduce the lag.

This adds some padding time before the frame based on the time wasted on the previous frame. The idea is to keep the en of the logic close to the actual swap of the buffer.

```
Pad -> Logic -> vsync -> Swap buffer -\
 ^                                    |
 |____________________________________|
```

```
APP_PUBLIC int
APP_Run_application_loop(int(*application_frame)(void)) {
    assert(application_frame);
    int loop_result = 0;
    static i64 pad_time = 0;
    for (;;) {
        i64 frame_begin = APP_Time();
        // Sleep based on the padding time
        struct timespec remaining, request = {0, pad_time};
        nanosleep(&request, &remaining);

        APP__Process_events();
        loop_result = application_frame();
        if (loop_result != 0) {break;}
        APP__Clear_events();

        i64 frame_end  = APP_Time();
[...]

Issue #341, opened 2020-07-21
sokol_app.h: macOS with GLCORE33 mouse position update is noticeably laggy


slmjkdbtl:
```c
#define SOKOL_IMPL
#define SOKOL_GLCORE33
// #define SOKOL_METAL

#include "sokol_app.h"

void event(const sapp_event* e) {
	printf("%f, %f\n", e->mouse_x, e->mouse_y);
}

sapp_desc sokol_main(int argc, char* argv[]) {
	return (sapp_desc){
		.event_cb = event,
	};
}
```
toggling between `SOKOL_GLCORE33` and `SOKOL_METAL` can tell the mouse position update with `GLCORE33` is laggier, it's more obvious if drawing something at mouse position

floooh:
I don't think I can do much about that. The mouse event code is the same between Metal and GL in sokol_app.h, it might be that NSOpenGLView has more latency for presenting the framebuffer than MTKView. I'll check for any obvious things (like triple- vs double-buffering), but it's unlikely that it's something as simple as that the case.

slmjkdbtl:
`[NSEvent setMouseCoalescingEnabled:NO]` did make the mouse event reporting more frequent (the printing frequency in the example above is higher), but it still seems a little bit laggier than with `SOKOL_METAL` (there's no obvious improvement if I draw stuff at mouse position). Thanks! I'll try do more research on this topic

slmjkdbtl:
I found that if I remove the `setNeedsDisplay` in `timerFired` impl it seem to report smooth mouse motion in the `mouseMoved` event, is it possible the timer is somehow interfering with the view / event update? (Also I didn't see other windowing libs uses a `NSTimer` and `setNeedsDisplay ` for updating view, is there an alternative way I can try?)

floooh:
Hmm ok, worth a try. IFIR had to add the NSTimer code because I couldn't get the CVDisplayLink method described here to properly work after the macOS Catalina GL regressions:

https://developer.apple.com/library/archive/qa/qa1385/_index.html

floooh:
Closing this with the mouse latency improvements for macOS that had just been merged. It's still not perfect, but very noticeably better.

PR #483, opened 2021-02-19
Improve macOS mouse latency and stale positioning

randrew:
Prior to this patch, `_sapp_macos_update_mouse()` used `[window mouseLocationOutsideOfEventStream]` to get the mouse position. `mouseLocationOutsideOfEventStream` seems to often lag behind the position of the mouse in the application's windows as reported by incoming mouse `NSEvent`s.

Because the old code before this commit only called `_sapp_macos_update_mouse()` right before `_sapp_frame()`, the individual `mouseMoved:`, `mouseDragged:`, etc. methods on the `_sapp_macos_view` class, which are the only entry points for calling `_sapp_macos_mouse_event()` to send the mouse data to the user application, would be sending the mouse position data that was returned by `_sapp_macos_update_mouse()`, not by the event in the incoming `NSEvent`.

Let's review what this looks like:

* `_sapp_macos_update_mouse()` is called right before rendering a frame. `[window mouseLocationOutsideOfEventStream]` is called within, and its return value is used to update the mouse position in the global `_state`. This data is not dispatched to the user application here. Additionally, `mouseLocationOutsideOfEventStream` may have returned a laggy mouse position.

* `_sapp_frame()` is called and a frame is rendered.

* Time passes.

* `NSEvent`s arrive on the `mouseMoved:`, `mouseDown:`, etc. methods. The mouse positions within the events are ignored, and instead, the position stored in the global `_state` from the previous call to `_sapp_macos_update_mouse()` is sent to the user application.

* Time passes.

* `_sapp_macos_update_mouse()` is called, …

This would result in the mouse input being delayed by one or more frames (at 60hz) on top of any additional latency added by vsync timing, video buffering, and display lag.

Another problem: after the user brings their mouse to a stop, the mouse position will often be incorrect — the newest mouse position has yet to be processed by the application. You can reproduce by moving the mouse quickly in a test application that draws a trail of squares along the mouse positions. When moving the mouse quickly and then coming to a stop, the trail of squares will stop short of the mouse cursor. Even if the mouse is left unmoving for seconds or minutes. Then the next time the user moves the mouse by even 1 pixel, a square will finally be drawn underneath the mouse position. (Except it will actually be the previous, old position.)

---

This commit makes several changes to input event handling in macOS:

`_sapp_macos_update_mouse()` has been changed to take an `NSEvent*` argument, instead of no arguments. And instead of using `[window mouseLocationOutsideOfEventStream]` to get the position of the mouse, it uses `event.locationInWindow`. This seems to sometimes eliminate a frame of mouse latency.

`_sapp_macos_update_mouse()` is no longer called immediately before rendering a frame. Instead, it's called at the beginning of each of the `mouseMoved:`, `mouseDragged:`,`mouseDown:`, etc. methods with its associated `NSEvent*`. Immediately after, `_sapp_macos_mouse_event()` is called. This eliminates a full frame of mouse latency. This also fixes situations where the mouse cursor's resting apparent position would lag behind its resting true position after the user stops moving the mouse.

`_sapp_macos_poll_input_events()` has been added, which calls `[NSApp nextEventMatchingMask: ...]` to manually read and dispatch input events from the event queue. `_sapp_macos_poll_input_events()` is called immediately before `_sapp_macos_frame()`. This will sometimes catch and dispatch input events that occurred shortly before the frame is to be rendered, but hadn't yet been dispatched by other mechanisms. This occasionally eliminates a frame of latency, and improves the worst-case bounds. This seems to happen to 5% or fewer mouse events.

`[NSEvent setMouseCoalescingEnabled:NO]`, if it even still does anything, is now called once at during application startup instead of toggled off and on along with mouse lock.

---

The improvements are noticeable in both Metal and OpenGL.

This significantly improves #341 and #415 for Mac.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2917,9 +2917,9 @@ _SOKOL_PRIVATE void _sapp_macos_update_window_title(void) {
     [_sapp.macos.window setTitle: [NSString stringWithUTF8String:_sapp.window_title]];
 }
 
-_SOKOL_PRIVATE void _sapp_macos_update_mouse(void) {
+_SOKOL_PRIVATE void _sapp_macos_update_mouse(NSEvent* event) {
     if (!_sapp.mouse.locked) {
-        const NSPoint mouse_pos = [_sapp.macos.window mouseLocationOutsideOfEventStream];
+        const NSPoint mouse_pos = event.locationInWindow;
         float new_x = mouse_pos.x * _sapp.dpi_scale;
         float new_y = _sapp.framebuffer_height - (mouse_pos.y * _sapp.dpi_scale) - 1;
         /* don't update dx/dy in the very first update */
@@ -2961,19 +2961,16 @@ _SOKOL_PRIVATE void _sapp_macos_lock_mouse(bool lock) {
         stack with calls to sapp_show_mouse()
     */
     if (_sapp.mouse.locked) {
-        [NSEvent setMouseCoalescingEnabled:NO];
         CGAssociateMouseAndMouseCursorPosition(NO);
         CGDisplayHideCursor(kCGDirectMainDisplay);
     }
     else {
         CGDisplayShowCursor(kCGDirectMainDisplay);
         CGAssociateMouseAndMouseCursorPosition(YES);
-        [NSEvent setMouseCoalescingEnabled:YES];
     }
 }
 
 _SOKOL_PRIVATE void _sapp_macos_frame(void) {
-    _sapp_macos_update_mouse();
     _sapp_frame();
     if (_sapp.quit_requested || _sapp.quit_ordered) {
         [_sapp.macos.window performClose:nil];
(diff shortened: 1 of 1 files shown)


randrew:
This significantly improves #341 and #415 for Mac.

randrew:
Additional disclaimer: I have only tested this on 10.12.

floooh:
Looks like a good fix, but I first need to do some git archeology to find out why ```mouseLocationOutsideOfEventStream``` is used in the first place instead of the more straightforward mouse movement events, and I need to do some testing. So it'll take a few days.

I'll start looking into this now though.

floooh:
Ok, merged! Just wanted to mention that this was a really nice PR, very non-intrusive, nicely written and explained. Thanks :)

floooh:
Hmm, I'm still seeing infrequent odd behaviour when pressing the window close button. Sometimes the close button remains pressed, rendering stops, and the application doesn't close.

I'll remove the event polling for a little while and see if that makes any difference. Just FIY.

Issue #400, opened 2020-10-06
sokol_audio: ALSA don't work


xglabs:
In function `_saudio_backend_init()`, `snd_pcm_hw_params_set_rate_near()` return negative result.

Reproduced on two different computers. Both run Arch current (x86_64):

-  Card: HDA Intel, Chip: Analog Devices AD1989B
- Card: HDA Intel PCH, Chip: Conexant CX20751/2

The following patch fixes sound (saudio-app and modplay-app work), but I'm not sure that it doesn't hammer something else.

```patch
diff --git a/modules/sokol/sokol_audio.h b/modules/sokol/sokol_audio.h
index bf7e4a7..ebafca2 100644
--- a/modules/sokol/sokol_audio.h
+++ b/modules/sokol/sokol_audio.h
@@ -1044,7 +1044,6 @@ _SOKOL_PRIVATE bool _saudio_backend_init(void) {
     snd_pcm_hw_params_any(_saudio.backend.device, params);
     snd_pcm_hw_params_set_access(_saudio.backend.device, params, SND_PCM_ACCESS_RW_INTERLEAVED);
     snd_pcm_hw_params_set_channels(_saudio.backend.device, params, _saudio.num_channels);
-    snd_pcm_hw_params_set_buffer_size(_saudio.backend.device, params, _saudio.buffer_frames);
     if (0 > snd_pcm_hw_params_test_format(_saudio.backend.device, params, SND_PCM_FORMAT_FLOAT_LE)) {
         goto error;
     }
@@ -1056,6 +1055,12 @@ _SOKOL_PRIVATE bool _saudio_backend_init(void) {
     if (0 > snd_pcm_hw_params_set_rate_near(_saudio.backend.device, params, &val, &dir)) {
         goto error;
     }
+    snd_pcm_uframes_t periodsize = _saudio.buffer_frames * 2;
+    rc = snd_pcm_hw_params_set_buffer_size_near(_saudio.backend.device, params, &periodsize);
+    if(rc < 0) {
+        goto error;
+    }
+
     if (0 > snd_pcm_hw_params(_saudio.backend.device, params)) {
         goto error;
     }
```

I.e. we set rate first, then try to set buffer size. The code was copied from alsa samples. I'm not sure, if we must change anything else (period_size).

P.S. It would be nice, if samples report sound initializion failure somehow. With red screen for example.

floooh:
[...]

PR #484, opened 2021-02-20
Added AAudio support for Android

oviano:
This can be selected by compiling with SOKOL_AAUDIO defined.

Requires Android API >= 26.

### DIFF
--- a/sokol_audio.h
+++ b/sokol_audio.h
@@ -527,6 +527,8 @@ inline void saudio_setup(const saudio_desc& desc) { return saudio_setup(&desc); 
 #elif (defined(__linux__) || defined(__unix__)) && !defined(__EMSCRIPTEN__) && !defined(__ANDROID__)
     #define ALSA_PCM_NEW_HW_PARAMS_API
     #include <alsa/asoundlib.h>
+#elif defined(__ANDROID__) && defined(SOKOL_AAUDIO)
+    #include "aaudio/AAudio.h"
 #elif defined(__ANDROID__)
     #include "SLES/OpenSLES_Android.h"
 #elif defined(_WIN32)
@@ -626,6 +628,16 @@ typedef struct {
     int buffer_frames;
     pthread_t thread;
     bool thread_stop;
+} _saudio_backend_t;
+
+/*=== AAudio BACKEND DECLARATIONS ==============================================*/
+#elif defined(__ANDROID__) && defined(SOKOL_AAUDIO)
+
+typedef struct {
+	AAudioStreamBuilder* builder;
+    AAudioStream* stream;
+    pthread_t thread;
+    pthread_mutex_t mutex;
 } _saudio_backend_t;
 
 /*=== OpenSLES BACKEND DECLARATIONS ==============================================*/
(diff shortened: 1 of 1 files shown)


floooh:
Thanks!

Do you think it make sense to drop the OpenSLES backend in favour of AAudio?

(e.g. is AAudio supported by the majority of Android devices, or does it make sense to keep OpenSLES as fallback option?)

oviano:
I would say keep them both for now but perhaps if left undefined you could have it automatically select AAudio if available?

API 26 is Android 8, and there are probably a fair number of devices still on Android 7 or below.

Here are some real time stats:

https://gs.statcounter.com/os-version-market-share/android/mobile-tablet/worldwide

sokol_audio.h
```diff
@@ -1660,7 +1672,113 @@ _SOKOL_PRIVATE void _saudio_backend_shutdown(void) {
...
+        _saudio_backend_shutdown();
+        return false;
+    }
+
+    AAudioStreamBuilder_setFormat(_saudio.backend.builder, AAUDIO_FORMAT_PCM_FLOAT);
+    AAudioStreamBuilder_setSampleRate(_saudio.backend.builder, _saudio.sample_rate);
+    AAudioStreamBuilder_setChannelCount(_saudio.backend.builder, _saudio.num_channels);
+    AAudioStreamBuilder_setBufferCapacityInFrames(_saudio.backend.builder, _saudio.buffer_frames * 2);
+    AAudioStreamBuilder_setDataCallback(_saudio.backend.builder, _saudio_aaudio_data_callback, NULL);
+	AAudioStreamBuilder_setFramesPerDataCallback(_saudio.backend.builder, _saudio.buffer_frames);
```
floooh, line 1766: Minor indentation typo here

floooh: ...looks like it's a tab character. Do you have an editor that honors the .editorconfig file?

(see: https://github.com/floooh/sokol/blob/master/.editorconfig)

oviano: Yep I am editing this in Xcode and I use tabs so I manually change them before submitting a PR.

I'll try and figure out a better workflow for next time.

floooh:
FYI, I'll first try to get Android CI builds working again, then I'll try to bump the default platform version in fips to something more recent, then I'll look into the PR. So it might be a few days.

I also have some sokol_audio.h changes queued up in the back of my head to address this topic:

https://github.com/floooh/sokol/issues/400

...I'll probably batch all sokol_audio.h related stuff. Just as a heads up :)

oviano:
> FYI, I'll first try to get Android CI builds working again, then I'll try to bump the default platform version in fips to something more recent, then I'll look into the PR. So it might be a few days.
> 
> I also have some sokol_audio.h changes queued up in the back of my head to address this topic:
> 
> #400
> 
> ...I'll probably batch all sokol_audio.h related stuff. Just as a heads up :)

Yeah that's fine

> FYI, I'll first try to get Android CI builds working again, then I'll try to bump the default platform version in fips to something more recent, then I'll look into the PR. So it might be a few days.
> 
> I also have some sokol_audio.h changes queued up in the back of my head to address this topic:
> 
> #400
> 
> ...I'll probably batch all sokol_audio.h related stuff. Just as a heads up :)

For reference, I used these resources for the implementation:

https://developer.android.com/ndk/guides/audio/aaudio/aaudio
https://developer.android.com/codelabs/making-waves-1-synth#4

oviano:
So just on this, be aware there is a problem with AAudio on Android 8.0, apparently fixed in 8.1 so if you end up making this automatically selected then best aim for API 27, not 26.

https://github.com/google/oboe/blob/master/docs/AndroidAudioHistory.md

floooh:
Wow, time flies :/

I'll try to look into this soon-ish. I'll take care of a couple other smaller 'drive-by' PRs, but then I'll try to test and merge this PR. It's amazing how little code is needed with AAudio vs OpenSLES, would be nice if we could get rid of OpenSLES alltogether, but I guess the responsible thing to do is to implement a fallback (I'll take care of this when merging).

oviano:
Nice one. I actually migrated away from sokol_audio because I needed hardware decoding in addition to raw PCM playback, but the PR was working perfectly fine when I stopped using it and perhaps someone will find it useful in future!

oviano:
As for getting rid of OpenSLES....what Android API level do you support for your other libraries, does it vary, or have you got some kind of consistent baseline?

Seems like API 26 and below is an ever-decreasing % of Android devices...

floooh:
Out of interest, what library are you using now, miniaudio maybe? Reason I'm asking is because I'd like to keep the sokol_audio.h feature set small, but want to recommend people something useful if they 'outgrow' the sokol-audio features :)

oviano:
> Out of interest, what library are you using now, miniaudio maybe? Reason I'm asking is because I'd like to keep the sokol_audio.h feature set small, but want to recommend people something useful if they 'outgrow' the sokol-audio features :)

No I ended up just writing my own code directly using the APIs for the various platforms.

For Apple platforms CoreAudio also allows passing packets to be decoded by configuring things slightly differently. Android requires some JNI interface code to get access to AudioTrack, which is quite similar to CoreAudio, and Windows I haven't properly figured out yet so I'm using decoding in software and accessing WASAPI much like sokol_audio does.

Since my project is AV-related I wanted to be able to do things like pass EAC3 stream data so that it can be directly decoded by the TV.

I was actually using miniaudio before I stumbled across your libraries, and since at that point I was only dealing with software decoding and PCM playback I found sokol_audio a more compact/simpler fit.

floooh:
Starting to tinker with this now. I'll probably kick out OpenSLES support alltogether while I'm at it...

floooh:
FYI, I'm closing the PR instead of merging. I had started a cleanup branch already and it was a bit of a hassle to cleanly merge the PR into it, so I copied your code over and applied the structural 'cleanup changes'. I'll make sure to mention you in the changelog though (hope that's ok).

oviano:
> FYI, I'm closing the PR instead of merging. I had started a cleanup branch already and it was a bit of a hassle to cleanly merge the PR into it, so I copied your code over and applied the structural 'cleanup changes'. I'll make sure to mention you in the changelog though (hope that's ok).

Perfect - nice one!

PR #487, opened 2021-02-26
ADDING: bindgen/gen_nim.py and related changes

garettbass:


floooh:
Ok, merged! I'll play around with this a bit now.

Btw, I just remembered that there's also a separate sokol-nim-examples repository: https://github.com/floooh/sokol-nim-samples

...but I think we should add the Nim examples to the sokol-nim project, like I did in the sokol-zig project.

floooh:
Oh and btw, I notice you're still on the master branch ;) It's really useful to put PRs in dedicated branches. That way old changes like this:

https://github.com/floooh/sokol/pull/487/commits/e4177472c8209d553135b7e61e2ccc87ec629449

...don't show up in the PR's changelist and also can't accidentally slip into the main branch when merging.

floooh:
Do you have some example code to get started with? It seems the sokol_app.h integration works a bit differently than I expected :)

E.g. I was basically expecting this:

```nim
proc myInitFunc(): void = 
    sg.setup(...)

proc myFrameFunc(): void = 
    ....

proc myCleanupFunc(): void =
    ....

sapp.run(sapp.Desc(
    initCb = myInitFunc,
    frameCb = myFrameFunc,
    cleanupCb = myCleanupFunc,
    width = 800,
    height = 600 
))
```

Also, wouldn't it make sense to automatically convert those "value-to-pointer" function wrappers here?

https://github.com/floooh/sokol-nim/blob/2bdc2d2be0d798527c832c96c1b882f3c69a1421/src/sokol/ext/gfx.nim#L110

...that's what I do in Zig too, the Zig functions take value arguments, and there are wrapper functions which call the exported C function with a pointer:

https://github.com/floooh/sokol-zig/blob/c0b820ede7978f3ec248c4ca301e3f6f2d84ec43/src/sokol/gfx.zig#L654-L656

...this is taken care of here where function arguments are treated differently depending on their type (struct pointers, char pointer, or "other"):

https://github.com/floooh/sokol/blob/4c56a2ee15d81cac91b738bb7e740c964dc6bd56/bindgen/gen_zig.py#L456-L466

Cheers :)

garettbass:
I have 3 samples converted to work with this: clear, cube, and blend.  I can put them in the sokol-nim repo following the pattern in sokol-zig.  I have removed some sapp freedom in favor of convenience and correct Nim behavior, but I think you have persuaded me to back off that a bit.  Also warming to the idea of generating the value->ptr functions.

PR #488, opened 2021-02-26
Change macOS OpenGL timing to use CVDisplayLink

randrew:
When building with `SOKOL_GLCORE33` on Mac, this commit adds the CoreVideo to the list of required frameworks: `-framework CoreVideo`

This replaces the `NSTimer` with `CVDisplayLink` for driving the timing of rendering frames with OpenGL on macOS. `CVDisplayLink` can be used to receive a callback on a separate high-priority thread synchronized with the timing of a display device.

Because sokol_app.h expects the rendering code to run on the main thread, we use: `NSObject performSelectorOnMainThread:` to trigger our code on the main thread.

Without vertical sync enabled, this will cause tearing. Maybe because both the rendering time of a frame can't be guaranteed to meet the deadline, and maybe also because of the time spent going from one thread to another.

However, it does give a consistent timer, and seems to be an improvement over `NSTimer`. In my measurements, CPU time waste from the high-frequency `NSTimer` was reduced.

---

I believe @floooh has found some issue with resizing the window with this patch. I haven't been able to reproduce that in 10.12, but I also haven't tried very hard.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3102,14 +3102,6 @@ _SOKOL_PRIVATE void _sapp_macos_frame(void) {
 
         _sapp.macos.window.contentView = _sapp.macos.view;
         [_sapp.macos.window makeFirstResponder:_sapp.macos.view];
-
-        NSTimer* timer_obj = [NSTimer timerWithTimeInterval:0.001
-            target:_sapp.macos.view
-            selector:@selector(timerFired:)
-            userInfo:nil
-            repeats:YES];
-        [[NSRunLoop currentRunLoop] addTimer:timer_obj forMode:NSDefaultRunLoopMode];
-        timer_obj = nil;
     #endif
     _sapp.valid = true;
     if (_sapp.fullscreen) {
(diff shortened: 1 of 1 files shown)


follower:
A short drive-by FYI:

As a result of research related to a `CVDisplayLink` issue in Godot (quite some time ago), I discovered many implementations miss an important detail which can result in crashes and/or program hangs (connected with system sleep and/or Fast User Switching):

> "_In macOS 10.4 and later, the **display link thread is automatically stopped if the user employs Fast User Switching**. The display link **is restarted when switching back** to the original user._"
>
> -- <https://developer.apple.com/documentation/corevideo/1457281-cvdisplaylinkstop?language=objc>

This means:

 1. The display link may already have been stopped. (Leading to an error on exit.)
 2. The display link may already have been stopped _and a *different* link may have been created_. (Leading to a hang during normal program execution because the originally created link disappeared.)

This may not be an issue with your use case but thought I'd at least mention it. :)

randrew:
Is that true? I slept and and woke my laptop several times, but the thread never died. Apple's documentation seems to indicate that the thread is stopped, but not killed, when switching to another user, and that it is resumed when switching back to the original user.

floooh:
Closing this, because the multi-window-stuff I'm currently working on in a branch (https://github.com/floooh/sokol/tree/sapp-multiwindow) will use CVDisplayLink both for driving the MTKView and NSOpenGLView (this was the only clean solution I found when rendering in multiple windows with a single sokol-app frame callback). Don't know yet when this will arrive in the master branch, but eventually it will arrive :)

PR #489, opened 2021-03-04
fix: wrong disable of msvc warning based on comment and commit details

ikrima:
C4202 => nonstandard extension used : '...': prototype parameter in name list illegal
  [https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4202?view=msvc-160]
C4204 => VS2015: nonstandard extension used: non-constant aggregate initializer
  [https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4204?view=msvc-160]

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -644,7 +644,7 @@ typedef struct sg_range {
 // disabling this for every includer isn't great, but the warnings are also quite pointless
 #if defined(_MSC_VER)
 #pragma warning(disable:4221)   /* /W4 only: nonstandard extension used: 'x': cannot be initialized using address of automatic variable 'y' */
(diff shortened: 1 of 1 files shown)


floooh:
Good catch. I was confused for a moment because somehow that old wrong warning number seems to have unbroken the VS2015 build on AppVeyor, but I think this is a side effect of including multiple header implementations into the same source file, and sokol_gfx.h actually using the correct pragma from sokol_app.h here:

https://github.com/floooh/sokol/blob/4c56a2ee15d81cac91b738bb7e740c964dc6bd56/sokol_app.h#L1486

I should better separate the header implementation in the compilation tests.

Thanks for the PR!

PR #496, opened 2021-03-22
Add Beef lang bindings generator

kochol:


floooh:
Thanks for the PR :) I'll try to give it a whirl soon-ish (I haven't tinkered with beef yet so I first will need to read up on this a bit).

One thing: The submodule should be removed, and instead for now the scripts should expect that the bindings git repo has been checked out manually at the right place. I don't know yet if directory structure is final, but in any case the sokol git repository shouldn't contain submodules.

kochol:
https://www.beeflang.org/ is a new programming language similar to C# but without GC and based on LLVM with many awesome features.

I also removed submodule as you requested.

floooh:
I'm currently looking into Beef a bit, just tinkering with the Beef IDE and example project. Really impressive stuff :)

Is the PR ready for merging or still WIP? As soon as you give the go ahead I can look into it. It would probably also be good to port a few examples from the sokol-samples repository.

kochol:
The below video has some info about language.
https://www.youtube.com/watch?v=L6w4y6_ENdU

For using sokol in Beef you have to create ether a static library or dynamic library to use them.
I try to port some samples so for now I make this PR WIP.

floooh:
Thanks for the link! If possible we should aim for a "batteries included" bindings repository which contains a snapshot of the sokol headers which match the generated bindings, and a simple way to compile those into a library (I would prefer static, but I don't know what's the preference is in the Beef world).

As example, check out how it looks in the Zig and Nim bindings (note the "c" subdirectory):

https://github.com/floooh/sokol-zig/tree/master/src/sokol

https://github.com/floooh/sokol-nim/tree/master/src/sokol

Both Zig and Nim can directly compile C source files though, don't know how this is handled with Beef in a cross-platform way (I guess if the Beef "build system" doesn't support compiling C libraries, then we should add a simple cmake file to the bindings repository which compiles the sokol headers into a library and could look like this snippet:

```cmake
if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
    add_library(sokol STATIC sokol/sokol.m ${SOKOL_HEADERS})
    target_link_libraries(sokol PUBLIC
        "-framework QuartzCore"
        "-framework Cocoa"
        "-framework MetalKit"
        "-framework Metal")
else()
    add_library(sokol STATIC sokol/sokol.c ${SOKOL_HEADERS})
    if (CMAKE_SYSTEM_NAME STREQUAL Linux)
        target_link_libraries(sokol INTERFACE X11 Xi Xcursor GL dl m)
        target_link_libraries(sokol PUBLIC Threads::Threads)
    endif()
endif()
```

(taken from: https://github.com/floooh/cimgui-sokol-starterkit/blob/main/CMakeLists.txt)

floooh:
PS: for porting the samples, it's a good idea to copy-paste the GLSL, HLSL and Metal shader code from the following platform-specific samples, instead of trying to extract the code-generated samples from the ```sapp``` samples (it might make sense to add Beef-output to sokol-shdc later):

- for HLSL: https://github.com/floooh/sokol-samples/tree/master/d3d11
- for GLSL: https://github.com/floooh/sokol-samples/tree/master/glfw
- for MetalSL: https://github.com/floooh/sokol-samples/tree/master/metal

kochol:
Hi @floooh 
I am learning project management from you. thank you

Beef can not compile c files at this time. I don't know if it will support later or not but can work with both static and dynamic libraries so I add a cmake script to compile sokol into the .lib file.

I also write a sokol-clear example in Beef and every thing works just fine.

below is my example code in Beef
```
using System;
using sokol;

namespace test_sokol
{
	class Program
	{
		static Gfx.Desc _desc = .();
		static Gfx.PassAction _pass = .();

		static void OnInit()
		{
			_desc.context = Glue.context();
			Gfx.setup(&_desc);
			_pass.colors[0] = .{ action = .CLEAR, value = .{ r=1, g=1, b=0, a=1 } };
		}

		static void OnFrame()
		{
			let g = _pass.colors[0].value.g + 0.01f;
			_pass.colors[0].value.g = g > 1.0f ? 0.0f : g;
			Gfx.beginDefaultPass(&_pass, App.width(), App.height());
			Gfx.endPass();
			Gfx.commit();
		}

		static void OnEvent(App.Event* _event)
		{

		}

		static void OnCleanUp()
		{
			Gfx.shutdown();
		}

		static void OnFail(char8* msg)
		{
			Console.WriteLine(msg);
		}

		public static int Main(String[] args)
		{
			App.Desc desc = .();
			desc.init_cb = => OnInit;
			desc.frame_cb = => OnFrame;
			desc.event_cb = => OnEvent;
			desc.cleanup_cb = => OnCleanUp;
			desc.fail_cb = => OnFail;
			desc.window_title = "Beef-Clear";
			App.run(&desc);

			return 0;
		}
	}
}
```

farism:
Is there any interest in trying to continue this?

floooh:
I'm happy to merge another attempt as long as the 'maintenance overhead' from my side is low, e.g. similar to how it was done for the sokol-jai, sokol-d or sokol-c3 contribution.

That means not only the bindings-generator, but I would create a `floooh/sokol-beef` repo which the contributor would get full push access, and would need to populate that repo with examples and gitlab CI integration for building the bindings and examples on CI.

PR #498, opened 2021-03-27
Updated gen_nim.py and gen_zig.py

garettbass:
* Updated gen_nim.py and gen_zig.py to use sokol_gfx.c and sokol_app.c instead of sokol_app_gfx.c
* Fixed gen_nim.py conversion of `const char*` to `cstring` for function arguments

### DIFF
--- a/bindgen/gen_nim.py
+++ b/bindgen/gen_nim.py
@@ -266,7 +266,7 @@ def as_extern_c_arg_type(arg_type, prefix):
     elif is_const_void_ptr(arg_type):
         return "pointer"
     elif is_string_ptr(arg_type):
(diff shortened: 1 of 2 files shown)


floooh:
I just stumbled over the same thing :)

I'm thinking about reorganizing the script structure a bit... basically getting rid of gen_all.py, and instead run gen_zig.py, gen_nim.py etc... directly. And probably move gen_ir.py back into a standalone script which writes a .json file (and gen_zig, gen_nim etc... consuming that json file). That's how it was in the beginning, and TBH I don't know why I though it's a good idea merging all this stuff :)

PR #505, opened 2021-04-05
sokol_imgui.h: fix compile if built with SOKOL_DUMMY_BACKEND

code-disaster:
Just a minor fix - global variables were renamed a while ago, but I guess nobody was crazy enough to try compiling with SOKOL_DUMMY_BACKEND. ;)


floooh:
Good catch thanks! I'll try to add a compilation test for the dummy-back to the sokol-samples test in the evening, so that this is caught in the future.

floooh:
Ok, found quite a few related problems:

https://github.com/floooh/sokol/commit/0d955d75cd909d778d15e040fe7b8cbf857e6cc0

There's now a compilation test here:

https://github.com/floooh/sokol-samples/blob/dc0055d4389e5a0c6f0ee3eb8c53cd4d8ff810fc/tests/compile/CMakeLists.txt#L86-L94

(note that sokol_app.h and sokol_fetch.h don't have dummy backends though).

PR #507, opened 2021-04-08
macOS: mouseDragged and rightMouseDragged mousebutton left and right

rcqls:
Currently, compared with `otherMouseDragged` activating `SAPP_MOUSEBUTTON_MIDDLE`, `mouseDragged` and `rightMouseDragged` activate `SAPP_MOUSEBUTTON_INVALID` whereas I expect `SAPP_MOUSEBUTTON_LEFT` and `SAPP_MOUSEBUTTON_RIGHT` respectively. The PR simply correct that. Is there some reason `mouseDragged` and `rightMouseDragged` behave differently from `otherMouseDragged` and do not reflect the actual mouse button activation?

BTW, sokol is really amazing!

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3026,7 +3026,7 @@ _SOKOL_PRIVATE void _sapp_macos_poll_input_events() {
         _sapp.mouse.dx = [event deltaX];
         _sapp.mouse.dy = [event deltaY];
     }
-    _sapp_macos_mouse_event(SAPP_EVENTTYPE_MOUSE_MOVE, SAPP_MOUSEBUTTON_INVALID , _sapp_macos_mod(event.modifierFlags));
+    _sapp_macos_mouse_event(SAPP_EVENTTYPE_MOUSE_MOVE, SAPP_MOUSEBUTTON_INVALID, _sapp_macos_mod(event.modifierFlags));
 }
 - (void)mouseDragged:(NSEvent*)event {
     _sapp_macos_update_mouse(event);
@@ -3034,7 +3034,7 @@ _SOKOL_PRIVATE void _sapp_macos_poll_input_events() {
         _sapp.mouse.dx = [event deltaX];
         _sapp.mouse.dy = [event deltaY];
     }
-    _sapp_macos_mouse_event(SAPP_EVENTTYPE_MOUSE_MOVE, SAPP_MOUSEBUTTON_INVALID , _sapp_macos_mod(event.modifierFlags));
+    _sapp_macos_mouse_event(SAPP_EVENTTYPE_MOUSE_MOVE, SAPP_MOUSEBUTTON_LEFT, _sapp_macos_mod(event.modifierFlags));
 }
 - (void)rightMouseDragged:(NSEvent*)event {
     _sapp_macos_update_mouse(event);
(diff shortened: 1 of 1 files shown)


floooh:
Hmmm... I'm afraid that the code in otherMouseDragged is an oversight, because that's the only place where a mouse button is set for mouseMoved events...

otherMouseDragged was implemented much later than the other two dragged callbacks, and having those mouseDragged handlers in the first place was a workaround for macOS not providing mouse-moved events outside the window area (regular mouse moved events also don't have the currently pressed mouse button set).

The original idea was basically that the application needs to check for mouse button events independently from mouse moved events.

If mouse moved events should also contain the currently pressed mouse buttons, then this should be done consistentely for all mouse moved events on all platforms, which is a bit of a bigger change. But I think this makes sense, because why "waste" the mouse button field in mouse moved events.
 
I'll keep this PR open for now as a reminder to make up my mind ;) I just don't want to merge a fix now which would let the Mac backend behave much differently from the other backends.

floooh:
PS: One very icky issue with the idea to have the mouse button field set in regular move events (and not just mouse button up/down events) is that in this case, the mouse button field needs to be a bit mask, because when "polling" button state like this, there might be multiple mouse buttons pressed (which is not a problem for the up/down events, because it's always just one button per up/down event which changes state).

For polling mouse buttons in other events, it might actually more sense to add mouse button modifiers here:

https://github.com/floooh/sokol/blob/e1058e11ec1dfb799740bf822fb91274adadd16c/sokol_app.h#L1086-L1091

...those would then be set as a bit mask in sapp_events here:

https://github.com/floooh/sokol/blob/e1058e11ec1dfb799740bf822fb91274adadd16c/sokol_app.h#L1099

I'll write a reminder ticket.  And I'll probably remove the middle mouse button from the otherMouseDragged to avoid any confusion in the future ;)

rcqls:
Thanks for your quick answer. Currently I am not really interested in otherMouseDragged event but more in mouseDragged and rightMouseDragged events that seem in my opinion the only way to catch the mouse button state when at the same time moving the mouse on the macOS backend. As you mentioned it, mouseMoved event does not seem to catch the mouse button state. So My PR seems to be the only way to get the mouse button (left or right and possibly the middle) state when dragging on the macOS side. With other OS,, at lest linux, I guess there is not this issue. 
What I understand with your last comment about SAPP_MODIFIER_ is to add mouse button (left, right and middle) state as additional  bits mask. Why not? But mouseDragged and rightMouseDragged event would still be  the only way to fulfill this SAPP_MODIFIER_ bit mask.

floooh:
> So My PR seems to be the only way to get the mouse button

The 'recommended way' currently is to catch mousebutton down/up events and store them yourself, similar to how I'm doing it here in the sokol_imgui.h header:

https://github.com/floooh/sokol/blob/e1058e11ec1dfb799740bf822fb91274adadd16c/util/sokol_imgui.h#L2063-L2076

...then if you need the currently pressed mouse buttons during mouse move:

```c
    case SAPP_EVENTTYPE_MOUSE_MOVE:
        if (_simgui.btn_up[SAPP_MOUSEBUTTON_LEFT]) {
            // left mouse button is currently pressed
        }
        if (_simgui.btn_up[SAPP_MOUSEBUTTON_RIGHT) {
            // right mouse button is currently pressed
        } 
        ...
```
...note how this allows for multiple mouse button to be pressed at the same time, which wouldn't work if the mouse button is stored in sapp_event.mouse_button.

With my proposed 'modifier fix', the mouse button state doesn't need to be tracked, instead:

```c
    case SAPP_EVENTTYPE_MOUSE_MOVE:
        if (event.modifiers & SAPP_MODIFIER_MOUSEBUTTON_LEFT) {
            // left mouse button is currently pressed
        }
        if (event.modifiers & SAPP_MOFIFIER_MOUSEBUTTON_RIGHT) {
            // right mouse button is currently pressed
        } 
        ...
```
...which also works when multiple mouse buttons are pressed at the same time.

How this is implemented in the platform-backends doesn't matter much, for instance on macOS, every input event also carries the current mouse button state in the ```pressedMouseButtons``` property:

https://developer.apple.com/documentation/appkit/nsevent/1527943-pressedmousebuttons?language=objc

rcqls:
Thank you very much for correcting me and for sharing these nice tips. I let you then close the PR since I understand that these parts of code would disappear soon.

floooh:
@rcqls this is now implemented, as described here:

https://github.com/floooh/sokol/issues/508

Example code is in the events-sapp.cc sample, e.g. you'd check the ```sapp_event.modifiers``` whether SAPP_MODIFIER_LMB/RMB/MMB is set:

https://github.com/floooh/sokol-samples/blob/73af0563ae7e0d5865e1be78d87cbd8f940bb9e6/sapp/events-sapp.cc#L302-L310

PR #513, opened 2021-04-17
Minor QoL changes

code-disaster:
- sokol_time : adds 100 Hz to list of common refresh rates
- sokol_app : adds function to query IDXGISwapChain object


floooh:
Looks good! Thanks!

PR #362, opened 2020-08-20
Allow macOS apps to use an optional main menu.

trojanfoe:
If a MainMenu.xib file is in the app bundle then it will be loaded and used as the main menu for the app.  This also allows Cmd-Q to be used to quit the app.

PR #515, opened 2021-04-20
MacOS main menu bar activation fix.

stevinz:
On MacOS 10.14.6, after implementing a quit menu item (seen in pull request #362) and subsequently additional menu items, I found that upon initial App launch the Apple menu and the App's menus in the main menu bar do not respond to mouse clicks. Switching away from the App and then back again resolves the issue.

I found others having the same issue (MacOS 10.15, using Swift) on stack overflow along with a fix: (https://stackoverflow.com/questions/62739862/why-doesnt-activateignoringotherapps-enable-the-menu-bar). Moving the App activation into the function applicationDidFinishLaunching() seems to work great, after this change, the menus are clickable from first launch.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2824,10 +2824,8 @@ _SOKOL_PRIVATE void _sapp_macos_run(const sapp_desc* desc) {
     // set the application dock icon as early as possible, otherwise
     // the dummy icon will be visible for a short time
     sapp_set_icon(&_sapp.desc.icon);
-    NSApp.activationPolicy = NSApplicationActivationPolicyRegular;
     _sapp.macos.app_dlg = [[_sapp_macos_app_delegate alloc] init];
     NSApp.delegate = _sapp.macos.app_dlg;
-    [NSApp activateIgnoringOtherApps:YES];
     [NSApp run];
     // NOTE: [NSApp run] never returns, instead cleanup code
     // must be put into applicationWillTerminate
@@ -3170,6 +3168,8 @@ _SOKOL_PRIVATE void _sapp_macos_frame(void) {
     else {
         [_sapp.macos.window center];
     }
+    NSApp.activationPolicy = NSApplicationActivationPolicyRegular;
+    [NSApp activateIgnoringOtherApps:YES];
     [_sapp.macos.window makeKeyAndOrderFront:nil];
     _sapp_macos_update_dimensions();
     [NSEvent setMouseCoalescingEnabled:NO];


meshula:
Good find! I thought I had an error in my application and was pulling my hair trying to track it down.

floooh:
Oh interesting... that's surprisingly late in the initialization process. Not the first time I've seen on Cocoa where changing the order of initialization calls has surprising effects though :)

I'll give it a whirl in the evening and then merge to master (and also integrate the change into the multi-window branch I'm currently working on, which will change the initialization code quite a bit again (but mostly the window-related stuff which is moved out into separate functions).

stevinz:
Awesome, thanks for your time. And thanks for the quick reply. 

For thoroughness, I found a few other posts around the web about the same issue. Here's another on stack overflow: (https://stackoverflow.com/questions/33345686/cocoa-application-menu-bar-not-clickable). And in this blog post (https://ar.al/2018/09/17/workaround-for-unclickable-app-menu-bug-with-window.makekeyandorderfront-and-nsapp.activate-on-macos/) the author goes to the trouble of programmatically toggling focus away from the app and back. It sounds like the bug is inconsistent across versions of MacOS, I haven't yet come across any official discussion from Apple.

> ... into the multi-window branch I'm currently working on...

Exciting stuff!!

floooh:
Ok merged! I don't have samples with a "proper" macOS menu to test, but it doesn't behave weird either, so I guess it's ok ;)

floooh:
Hmm, one question: does it matter if the two lines:

```
    NSApp.activationPolicy = NSApplicationActivationPolicyRegular;
    [NSApp activateIgnoringOtherApps:YES];
```

...are moved after the call to:
```
    [_sapp.macos.window makeKeyAndOrderFront:nil];
```
??

Because in my multiwindow-branch, the makeKeyAndOrderFront call is inside the new platform-create-window function, and without hacks I'd need to move the 'activation stuff' after the call to makeKeyAndOrderFront (otherwise, if the order matter I'll come up with a workaround).

floooh:
NVM it seems to work. At least I can open the Apple menu when moving the activation code after makeKeyAndOrderFront (I actually never questioned why I can't open the Apple menu lol, but now it works both with your fix on the master branch, and on my wip multiwindow-branch by moving the two activation lines after makeKeyAndOrderFront (or rather: after the platform-create-window call).

stevinz:
Seems to work that way fine for me too... Looks great, thanks!

PR #527, opened 2021-06-01
sg_color utilities

nyalloc:
sokol_color.h defines preset colors based on the [X11 color names](https://en.wikipedia.org/wiki/X11_color_names).

This is useful when you want to reference a familiar color, but don't want to write it out by hand. The palette should be familiar to web developers and those familiar with XNA / MonoGame.

The header is generated from gen_sokol_color.py. Users can use this script to extend the palette locally if they feel like adding more color presets to their project.

The palette is defined using `static const` (or `static constexpr` if you are using a C++ compiler) objects. These objects use lowercase names:

```c
static SOKOL_COLOR_CONSTEXPR sg_color sg_red = SG_RED;
static SOKOL_COLOR_CONSTEXPR sg_color sg_green = SG_GREEN;
static SOKOL_COLOR_CONSTEXPR sg_color sg_blue = SG_BLUE;
```

An `sg_color` preset object like `sg_red` can be used to initialize an `sg_pass_action`:

```c
sg_pass_action pass_action = {
    .colors[0] = { .action=SG_ACTION_CLEAR, .value = sg_red }
};
```

Initializing an object with static storage duration is more complicated because of C language rules. Technically, a `static const` object is not a compile-time constant in C. To work around this, the palette is also defined as a series of brace-enclosed list macro definitions. These definitions use uppercase names:

```c
#define SG_RED { 1.0f, 0.0f, 0.0f, 1.0f }
#define SG_GREEN { 0.0f, 1.0f, 0.0f, 1.0f }
#define SG_BLUE { 0.0f, 0.0f, 1.0f, 1.0f }
```

A preset macro like `SG_RED` can be used to initialize objects with static storage duration:

```c
static struct {
    sg_pass_action pass_action;
} state = {
    .pass_action = {
        .colors[0] = { .action = SG_ACTION_CLEAR, .value = SG_RED }
    }
};
```

A second set of macro definitions exists for colors packed as 32 bit integer values. These definitions are also uppercase, but use the `_RGBA32` suffix:

```c
#define SG_RED_RGBA32 0xFF0000FF
#define SG_GREEN_RGBA32 0x00FF00FF
#define SG_BLUE_RGBA32 0x0000FFFF
```

This is useful if your code makes use of packed colors, as sokol_gl.h does for its internal vertex format:

```c
sgl_begin_triangles();
sgl_v2f_c1i( 0.0f,  0.5f, SG_RED_RGBA32);
sgl_v2f_c1i( 0.5f, -0.5f, SG_GREEN_RGBA32);
sgl_v2f_c1i(-0.5f, -0.5f, SG_BLUE_RGBA32);
sgl_end();
```


nyalloc:
Implemented a few utility functions for creating and modifying colors, documentation still needs to be updated to reflect this.

nyalloc:
> Implemented a few utility functions for creating and modifying colors, documentation still needs to be updated to reflect this.

Updated the documentation.

floooh:
And merged! Thanks for the nice documentation, very helpful :) I'll try to add some tests to the sokol-samples repository, and start thinking about a little sample.

PR #537, opened 2021-07-15
sokol_fetch docs fix

geekynils:
The TL;DR instructions should also mention that you need to call sfetch_dowork().


floooh:
Oops, yep totally makes sense. Thanks for the PR!

PR #539, opened 2021-07-20
Update gen_zig to support 0.8.0 and master

leecannon:
These changes are required for zig master and are also valid for the latest stable version 0.8.0


floooh:
Ok, many thanks! I haven't tested with the zig master branch yet, but 0.8.0 works fine. I also updated the sokol-zig repo with the fixed bindings.

Issue #460, opened 2021-01-20
[Feature Request] - Support accented characters (non-us keyboard) on MacOS


eduardodoria:
It's missing support for accented characters in event SAPP_EVENTTYPE_CHAR on MacOS (using dead keys) for non-us keyboards. Like this: á, ã, ù, etc.

I checked and maybe necessary to use UCKeyTranslate, like this example: https://stackoverflow.com/questions/22566665/how-to-capture-unicode-from-key-events-without-an-nstextview

eduardodoria:
I found a more ellegant solution in GLFW, using NSTextInputClient protocol in NSView, like this example: https://github.com/glfw/glfw/blob/3.3.2/src/cocoa_window.m

It works perfectly!

eduardodoria:
The same behavior seems to appear in Linux (Ubuntu 20.04.3 LTS). The sokol_app uses `XLookupString` inside KeyPress to capture char, but it is not working with accented characters.

Using `xev` command at terminal I get this result when I press "´" and "a":

```
KeyPress event, serial 37, synthetic NO, window 0x2400001,
    root 0x17d, subw 0x0, time 11120406, (153,-22), root:(259,92),
    state 0x10, keycode 34 (keysym 0xfe51, dead_acute), same_screen YES,
    XLookupString gives 2 bytes: (c2 b4) "´"
    XmbLookupString gives 0 bytes: 
    XFilterEvent returns: True

KeyRelease event, serial 37, synthetic NO, window 0x2400001,
    root 0x17d, subw 0x0, time 11120469, (153,-22), root:(259,92),
    state 0x10, keycode 34 (keysym 0xfe51, dead_acute), same_screen YES,
    XLookupString gives 2 bytes: (c2 b4) "´"
    XFilterEvent returns: False

KeyPress event, serial 37, synthetic NO, window 0x2400001,
    root 0x17d, subw 0x0, time 11123685, (153,-22), root:(259,92),
    state 0x10, keycode 38 (keysym 0x61, a), same_screen YES,
    XLookupString gives 1 bytes: (61) "a"
    XmbLookupString gives 1 bytes: (61) "a"
    XFilterEvent returns: True

KeyPress event, serial 37, synthetic NO, window 0x2400001,
    root 0x17d, subw 0x0, time 11123685, (153,-22), root:(259,92),
    state 0x10, keycode 0 (keysym 0xe1, aacute), same_screen YES,
    XLookupString gives 0 bytes: 
[...]

PR #542, opened 2021-08-01
impl NSTextInputClient

Hperigo:
the PR implements the NSTextInputClient ( #460 ) for `MTKView` and `NSOpenGLView`, this will allow `_sapp_macos_app` to receive key code for characters like `á, é, í, ó, ú` 

a couple of notes: 

1. the `_sapp.event.char_code` is set inside `insertText` method. `_sapp.event.modifiers` and `_sapp.event.key_repeat` are still inside `keyDown` I'm not sure if this is going to break something, but it looks correct. 
2. As suggested in the original feature request ( #460 ) this impl. is similar to the GLFW, the main difference is that `firstRectForCharacterRange` is returning `return NSMakeRect(0.0, 0.0, 0.0, 0.0);`. the original impl uses the window bounds, I'm not sure what is the difference here.. 

This is my first PR here, so feel free to point any corrections, etc.. =)

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1781,6 +1781,9 @@ inline void sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
 
 /*== MACOS DECLARATIONS ======================================================*/
 #if defined(_SAPP_MACOS)
+
+static const NSRange kEmptyRange = { NSNotFound, 0 };
+
 @interface _sapp_macos_app_delegate : NSObject<NSApplicationDelegate>
 @end
 @interface _sapp_macos_window : NSWindow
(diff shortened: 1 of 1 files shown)


floooh:
I need to read up on this to provide better feedback, but one dangerous thing I noticed while looking through the changes is that your PR undoes the following bug fix:

https://github.com/floooh/sokol/commit/b326f332163d16d9210b7d0790bd4772e30137f5

Not sure how this can happen TBH.

floooh:
(I changed the PR to a draft so that I don't accidentally click the merge button)

Hperigo:
I'll double check the `b326f33` that you pointed, I think that's was a mistake on my end.. 

please feel free to provide any other feedback =)

eduardodoria:
I changed https://github.com/floooh/sokol/issues/460 because it is not working on Linux (Ubuntu 20.04.3 LTS) too.

PR #543, opened 2021-08-02
CMake init

SamuelMarks:
Related https://github.com/floooh/sokol/pull/496#issuecomment-806516897

Not sure how you want to handle this? - Maybe configurable packages where you can use `-D` to select a different backend for your target?

(assuming you even want CMake integration)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,63 @@
+cmake_minimum_required(VERSION 3.19)
+
+# set the project name and version
+project(sokol VERSION 0.0.1 LANGUAGES C)
+string(TOLOWER "${PROJECT_NAME}" PROJECT_LOWER_NAME)
+
+set(CMAKE_C_STANDARD 90)
+set(CMAKE_VERBOSE_MAKEFILE ON)
+
+add_library("${PROJECT_LOWER_NAME}_compiler_flags" INTERFACE)
+target_compile_features("${PROJECT_LOWER_NAME}_compiler_flags" INTERFACE "c_std_90")
+
+# add compiler warning flags just when building this project via
+# the BUILD_INTERFACE genex
+set(gcc_like "$<COMPILE_LANG_AND_ID:C,CXX,ARMClang,AppleClang,Clang,GNU>")
+set(msvc "$<COMPILE_LANG_AND_ID:C,CXX,MSVC>")
+target_compile_options(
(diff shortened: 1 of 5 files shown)


floooh:
Thanks for the PR, but I'd like to keep any build system files out of the sokol repository. The comment was aimed specifically at the beef bindings, because beef doesn't seem to come with a way to build C libraries without external tooling support.

SamuelMarks:
Okay but how about for integration with package managers like [`vcpkg`](https://github.com/microsoft/vcpkg)?

There are plenty of complexities to consider, and I'm trying to derive a simplification for use by package managers.

So on starting a new project:

  0. Specify target OS (e.g., iOS, Android, WASM, macOS, Linux, and Windows)
  1. Add additional dependencies (e.g., libcurl)
  2. Configure
  3. Build
  4. Test
  5. Package

Now the question is how can this workflow be simplified?

Ideas for sokol:

  - Integration with `vcpkg` (or similar, like `conan` or `clibs`)
  - Integration directly with CMake (or similar), e.g., using [`file(DOWNLOAD`](https://cmake.org/cmake/help/latest/command/file.html#download) with the specific headers from this repo
  - Integration with OS-specific package managers (apt, yum, dnf, MSI-based ones, brew, &etc.)

Being able to choose which backend is important, e.g., imgui or nuklear.

What are your thoughts?

floooh:
Support for specific build systems and package managers should be handled outside the main repository in separate repositories similar to the langauge bindings (e.g. https://github.com/floooh/sokol-zig). Those separate repositories can either contain snapshots of the sokol headers (like here: https://github.com/floooh/sokol-zig/tree/master/src/sokol/c), or if it makes more sense, can include the sokol repository as a git submodule (whatever makes more sense for the specific build system or package manager).

I've done similar "out-of-source" build system repositories here for various projects and fips+cmake:

- fips + Dear ImGui: https://github.com/fips-libs/fips-imgui
- fips + Nuklear: https://github.com/fips-libs/fips-nuklear
- fps + libmodplug: https://github.com/fips-libs/fips-libmodplug

...etc... that way I didn't have to pitch fips to the library maintainers for instance.

SamuelMarks:
fips looks pretty cool (I hadn't heard of this project of yours).

Will give it a shot. Perhaps it solves the missing 'quick-start' for sokol…

PR #564, opened 2021-09-22
Fix `sapp_mouse_shown` function signature

cedric-h:
I discovered this while experimenting with adding a new language to the bindgen


floooh:
Oh no... I will immediately return my C programming license. Many thanks for noticing :)

PR #583, opened 2021-11-21
Check D3D11 calls failures

edubart:
Currently, some D3D11 APIs that create resources like a texture may fail in the following scenarios:

* GPU device has been removed
* GPU driver has been updated
* GPU is out of memory

The Sokol GFX code does not check if these functions fail, instead `SOKOL_ASSERT` is used for example in the result of `_sg_d3d11_CreateTexture2D`, however it's better to be able to detect the fail to be able to recover from such situations. For example, it is not uncommon to have computers where the GPU device is removed shortly due to change of power state of a laptop like battery energy saving (where the laptop switches from dedicated GPU card to integrated GPU card) or a GPU driver being updated. In this case failure proof applications will recover from that error by checking at swap phase if D3D11 `Present` fails with device removed error and then recreates the entire device context, but currently until the `Present` call any resource creation call like `CreateTexture2D` will fail and then crash, thus the current design of Sokol GFX does not allow applications to fully recover from those situations.

There is also another case when the GPU runs out of memory, currently this will give a crash in D3D11, and with this PR `sg_make_image` will return `SG_RESOURCESTATE_FAILED`, enabling the user to decide how to act in the failure.

In summary this PR add means to recover such scenarios, by returning `SG_RESOURCESTATE_FAILED` in many D3D11 resource creation calls (like `_sg_d3d11_CreateTexture2D` and `_sg_d3d11_CreateTexture3D`) instead of using just `SOKOL_ASSERT`, and also add checks in `_sg_d3d11_Map` to avoid crashes when the device is an bad state and it fails.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -8256,8 +8256,10 @@ _SOKOL_PRIVATE sg_resource_state _sg_d3d11_create_buffer(_sg_buffer_t* buf, cons
             init_data_ptr = &init_data;
         }
         HRESULT hr = _sg_d3d11_CreateBuffer(_sg.d3d11.dev, &d3d11_desc, init_data_ptr, &buf->d3d11.buf);
-        _SOKOL_UNUSED(hr);
-        SOKOL_ASSERT(SUCCEEDED(hr) && buf->d3d11.buf);
+        if (!(SUCCEEDED(hr) && buf->d3d11.buf)) {
+            SOKOL_LOG("failed to create D3D11 buffer\n");
+            return SG_RESOURCESTATE_FAILED;
+        }
     }
     return SG_RESOURCESTATE_VALID;
 }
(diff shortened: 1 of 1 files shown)


floooh:
Thanks for the PR! At first glance those changes look good. Eventually we'll need better error reporting, but this is a good start.

One minor nitpick:

Please fix ```if(...)``` to ```if (...)``` (space between if and the opening brace).

It'll will probably take a bit (2..3 weeks maybe) until I can focus on sokol stuff again though.

edubart:
> Please fix `if(...)` to `if (...)` (space between if and the opening brace).

Fixed.

> It'll will probably take a bit (2..3 weeks maybe) until I can focus on sokol stuff again though.

No problem, thanks for Sokol and for reviewing this PR.

iryont:
@floooh It would be lovely if you could review and merge it.

I have been testing this change for some time and it works perfectly. My application can fully recover from GPU failures.

floooh:
Ok merged! I found a comment that needs to be moved but I'll just do that in a separate commit.

Thanks!

PR #587, opened 2021-12-01
add point size attribute to sokol_fontstash.h

jasonliang-dev:
When using sokol_gl.h + sokol_fontstash.h with GLES2 or D3D11, the version of sokol_gl.h with point size support creates the validation error: "missing vertex attribute name/semantics in shader".

I added the vertex attribute info required for GLES2 and D3D11 and `psize` to the shader code.


floooh:
Ah oops, that was an oversight on my side. Many thanks!

Issue #576, opened 2021-10-28
Reading from depth textures


PeterHiber:
Hi again! Sorry in advance, we know that this is a duplicate of https://github.com/floooh/sokol/issues/165. But it's been some time, so basically we are creating this issue to see if anything has changed. Please feel free to close it if you wish :)

Being able to read from the depth texture is very useful. Its most obvious usage (shadow maps) can easily be worked around by writing the depth as a color in the pixel shader. But for post process effects (e.g. SSAO, depth of field, fog, etc) this is not really feasible, as you would have to render all geometry twice (in WebGL 1 without MRT). At least for us, our geometry pass is really expensive.

Looking around a bit, `WEBGL_depth_texture` seems well supported on basically everything as far as I can tell. But it seems the problem is more with SPIRV-Cross and sokol_shdc? If it helps anything, we don't necessarily need the whole "sample depth texture, compare it with reference value and return 0 or 1 depending on comparison" thing. We just want to be able to sample the raw depth with nearest filtering.

Is there anything we can do to help to make this happen? :)

floooh:
Haven't looked into this yet, sorry. I think the best way forward would be to gate the issue behind a new feature flag (sg_query_features) so that support can be added gradually to the backends instead of having to touch all backends at once.

For GL, these parts would need to be touched:

- GLES2/WebGL needs an extension check here: https://github.com/floooh/sokol/blob/8fcd271f7d4cb19c77aa9f0bbfe8cb7f319dbffc/sokol_gfx.h#L5659-L5699

- in _sg_gl_create_texture() the special case block for depth-stencil-textures should be skipped if WEBGL_depth_texture is supported: https://github.com/floooh/sokol/blob/8fcd271f7d4cb19c77aa9f0bbfe8cb7f319dbffc/sokol_gfx.h#L6237-L6251

- and this part in _sg_gl_create_pass() needs to be fixed accordingly: https://github.com/floooh/sokol/blob/8fcd271f7d4cb19c77aa9f0bbfe8cb7f319dbffc/sokol_gfx.h#L6706-L6714

- IMHO the old render buffer code should still remain as fallback for WebGL1 (but hopefully we can ditch WebGL1 support in the not-too-distant future)

(also note the special case handling for MSAA framebuffers, this needs to be handled the same for depth-textures)

If you want to create such a PR limited to the GL backend I'm happy to merge :)

(I think support for the other backends will be much easier)

floooh:
PS:

[...]

PR #591, opened 2021-12-09
Experimental depth read support

PeterHiber:
This PR contains experimental support for reading the depth buffer from shaders, see: https://github.com/floooh/sokol/issues/576

What works:
* Reading depth buffer in D3D11
* Reading depth buffer in WebGL 2

What doesn't work:
* Metal
* Issues reading depth buffer in WebGL 1 with depth_texture extension, not sure why
* Not super well-tested with depth_stencil formats, likely broken in some variations.

At this point we are a bit stuck mainly at the Metal implementation. In our testing environment we don't get any errors of any kind when attempting to read the depth buffer, just a black screen. (Our testing environment could be bad, and we readily admit that we are fairly new to Metal compared to other GFX API's) We suspect that it might be related to the difference between `depth2d` and `texture2d` types in MSL shaders.

Our time on this has unfortunately run out a bit, and it will take a while until we have more time to look at it. If anyone wants to take a look at and fix issues with this PR it would be greatly appreciated. :)

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -3474,6 +3474,7 @@ typedef struct {
     _sg_gl_state_cache_t cache;
     bool ext_anisotropic;
     GLint max_anisotropy;
+    bool ext_depth_read;
     GLint max_combined_texture_image_units;
     #if _SOKOL_USE_WIN32_GL_LOADER
     HINSTANCE opengl32_dll;
@@ -5563,6 +5564,7 @@ _SOKOL_PRIVATE void _sg_gl_init_caps_glcore33(void) {
     _sg.features.image_clamp_to_border = true;
     _sg.features.mrt_independent_blend_state = false;
     _sg.features.mrt_independent_write_mask = true;
+    _sg.gl.ext_depth_read = true;
 
     /* scan extensions */
     bool has_s3tc = false;  /* BC1..BC3 */
@@ -5640,6 +5642,7 @@ _SOKOL_PRIVATE void _sg_gl_init_caps_gles3(void) {
     _sg.features.image_clamp_to_border = false;
     _sg.features.mrt_independent_blend_state = false;
     _sg.features.mrt_independent_write_mask = false;
+    _sg.gl.ext_depth_read = true;
 
     bool has_s3tc = false;  /* BC1..BC3 */
     bool has_rgtc = false;  /* BC4 and BC5 */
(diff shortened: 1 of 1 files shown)


pseregiet:
Hi. This is required to create a render pass with a cubemap depth buffer attachment

```
diff --git a/3rdparty/sokol/sokol_gfx.h b/3rdparty/sokol/sokol_gfx.h
index ec893f7..bd4be4f 100644
--- a/3rdparty/sokol/sokol_gfx.h
+++ b/3rdparty/sokol/sokol_gfx.h
@@ -6964,8 +6964,18 @@ _SOKOL_PRIVATE sg_resource_state _sg_gl_create_pass(_sg_pass_t* pass, _sg_image_
         if (_sg.gl.ext_depth_read) {
             const GLuint gl_depth_tex = pass->gl.ds_att.image->gl.tex[0];
             SOKOL_ASSERT(gl_depth_tex);
-            glFramebufferTexture2D(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D, gl_depth_tex, 0);
-             if (_sg_is_depth_stencil_format(pass->gl.ds_att.image->cmn.pixel_format)) {
+            switch (pass->gl.ds_att.image->cmn.type) {
+                case SG_IMAGETYPE_2D:
+                    glFramebufferTexture2D(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D, gl_depth_tex, 0);
+                    break;
+                case SG_IMAGETYPE_CUBE:
+                    const GLenum slice = _sg_gl_cubeface_target(pass->cmn.ds_att.slice);
+                    glFramebufferTexture2D(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, slice, gl_depth_tex, 0);
+                    break;
+                default:
+                    break;
+            }
+            if (_sg_is_depth_stencil_format(pass->gl.ds_att.image->cmn.pixel_format)) {
                 glFramebufferTexture2D(GL_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D, gl_depth_tex, 0);
             }
         }
```

With this change you still need to attach a dummy RGB texture even if you don't use it in the shader. Do you have any plans on fixing that ? I tried but I think it's a bit too complicated for me.

floooh:
Some updates... I was hoping to allow depth texture access in shaders with https://github.com/floooh/sokol/issues/816 but ran into various problems in the other 3D backends (specifically D3D11). I'll make another attempt after finalizing the WebGPIU backend.

Rendering without color attachments is also on the list to fix (this will be easier I think).

Basically: first finalize the WebGPU backend, then re-evaluate how much the following things can be implemented, either across all backends, or behind a runtime feature flag:

- accessing MSAA textures from shaders (e.g. loading individual samples)
- accessing depth textures from shaders (as 'raw values', not the special 'depth compare' sampling in GLSL)
- depth-only rendering (no color attachments)
- rendering without vertex/index buffer (still won't allow 'proper' vertex pulling from buffers though because at least WebGL2 lacks SSBOs)

PR #600, opened 2021-12-21
Fix SAPP_EVENTYTPE_KEY_UP typo

SanderMertens:


floooh:
Ah good catch, thanks!

PR #605, opened 2021-12-30
fixed sokol_debugtext.h for SOKOL_WGPU backend

mchiasson:
Just a minor typo and some code that probably didn't get updated during a past refactoring session.


floooh:
Many thanks! I'll merge the PR, but please keep in mind that the sokol_gfx.h WebGPU backend hasn't been kept uptodate with the latest WebGPU API changes and probably won't currently work. My plan is to delay this until the API has stabilized (so probably "pretty soon now").

PR #607, opened 2022-01-07
sokol_imgui.h: use ImDrawCmd::IdxOffset

0x1100:
sokol_imgui.h wasn't using ```ImDrawCmd::IdxOffset```, causing visual gliches with Dear ImGui v1.86.
Several backends had the same bug. The release notes explain the bug and its fix:
https://github.com/ocornut/imgui/releases/tag/v1.86

I cast the value to ```int``` to be consistent with the rest of the code.


floooh:
Yep, much better :)

Screenshot 2022-01-07 at 18 32 45

PR #610, opened 2022-01-13
sokol_app.h win32: implements optional per monitor DPI awareness

tjachmann:
For win32 adds per monitor DPI arewareness, so the window's DPI gets updated when it is moved to a monitor with a different DPI. 
 - Reverts to system DPI, if an older windows then Windows 10 Creators Update is used
 - Although I think this behaviour should be the default, I made it optional for now.  See flag in backend-specific options in app_desc

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1366,6 +1366,8 @@ typedef struct sapp_desc {
     /* backend-specific options */
     bool gl_force_gles2;                // if true, setup GLES2/WebGL even if GLES3/WebGL2 is available
     bool win32_console_utf8;            // if true, set the output console codepage to UTF-8
+    bool win32_default_size;            // if true, set the output console codepage to UTF-8
+    bool win32_per_monitor_dpi_v2;      // if true and high_dpi is also enabled, enables support for per monitor DPI for Windows 10 (since Creators Update)
     bool win32_console_create;          // if true, attach stdout/stderr to a new console window
     bool win32_console_attach;          // if true, attach stdout/stderr to parent process
     const char* html5_canvas_name;      // the name (id) of the HTML5 canvas element, default is "canvas"
@@ -6227,11 +6229,12 @@ _SOKOL_PRIVATE void _sapp_win32_lock_mouse(bool lock) {
         SOKOL_ASSERT(res); _SOKOL_UNUSED(res);
     }
 }
-
 /* updates current window and framebuffer size from the window's client rect, returns true if size has changed */
 _SOKOL_PRIVATE bool _sapp_win32_update_dimensions(void) {
     RECT rect;
     if (GetClientRect(_sapp.win32.hwnd, &rect)) {
+
+
         _sapp.window_width = (int)((float)(rect.right - rect.left) / _sapp.win32.dpi.window_scale);
         _sapp.window_height = (int)((float)(rect.bottom - rect.top) / _sapp.win32.dpi.window_scale);
         int fb_width = (int)((float)_sapp.window_width * _sapp.win32.dpi.content_scale);
@@ -6410,6 +6413,7 @@ _SOKOL_PRIVATE LRESULT CALLBACK _sapp_win32_wndproc(HWND hWnd, UINT uMsg, WPARAM
                     PostQuitMessage(0);
                 }
                 return 0;
+
             case WM_SYSCOMMAND:
                 switch (wParam & 0xFFF0) {
                     case SC_SCREENSAVE:
@@ -6458,6 +6462,35 @@ _SOKOL_PRIVATE LRESULT CALLBACK _sapp_win32_wndproc(HWND hWnd, UINT uMsg, WPARAM
                     }
                 }
                 break;
+
+
+            // Update window's DPI and size if its moved to another monitor with a different DPI
+            // Only send by windows, if PROCESS_PER_MONITOR_DPI_AWARE:
+            case WM_DPICHANGED:
+            {
+                HINSTANCE user32 = LoadLibraryA("user32.dll");
+
+                if (user32) {
+                    typedef UINT(WINAPI * GETDPIFORWINDOW_T)(HWND hwnd);
+                    GETDPIFORWINDOW_T fn_getdpiforwindow = (GETDPIFORWINDOW_T)(void*)GetProcAddress(user32, "GetDpiForWindow");
+
+                    if (fn_getdpiforwindow) {
+
+                        UINT dpix = fn_getdpiforwindow(_sapp.win32.hwnd);
+                        _sapp.win32.dpi.window_scale = (float)dpix / 96.0f;
+                        _sapp.win32.dpi.content_scale = _sapp.win32.dpi.window_scale;
+                        _sapp.dpi_scale = _sapp.win32.dpi.window_scale;
+
+                        LPRECT lpRect = (LPRECT)lParam;
+                        SetWindowPos(hWnd, nullptr, lpRect->left, lpRect->top, lpRect->right - lpRect->left, lpRect->bottom - lpRect->top, SWP_NOZORDER | SWP_NOACTIVATE);
+                    }
+
+                    FreeLibrary(user32);
+                }
+
+                break;
+            }
+
             case WM_LBUTTONDOWN:
                 _sapp_win32_mouse_event(SAPP_EVENTTYPE_MOUSE_DOWN, SAPP_MOUSEBUTTON_LEFT);
                 _sapp_win32_capture_mouse(1<<SAPP_MOUSEBUTTON_LEFT);
@@ -6572,6 +6605,7 @@ _SOKOL_PRIVATE LRESULT CALLBACK _sapp_win32_wndproc(HWND hWnd, UINT uMsg, WPARAM
             case WM_EXITSIZEMOVE:
                 KillTimer(_sapp.win32.hwnd, 1);
                 break;
+
             case WM_TIMER:
                 _sapp_win32_timing_measure();
                 _sapp_frame();
@@ -6612,6 +6646,7 @@ _SOKOL_PRIVATE void _sapp_win32_create_window(void) {
     wndclassw.hIcon = LoadIcon(NULL, IDI_WINLOGO);
     wndclassw.lpszClassName = L"SOKOLAPP";
     RegisterClassW(&wndclassw);
+    
 
     DWORD win_style;
     const DWORD win_ex_style = WS_EX_APPWINDOW | WS_EX_WINDOWEDGE;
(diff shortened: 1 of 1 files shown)


sokol_app.h
```diff
@@ -6458,6 +6462,35 @@ _SOKOL_PRIVATE LRESULT CALLBACK _sapp_win32_wndproc(HWND hWnd, UINT uMsg, WPARAM
                     }
                 }
                 break;
+
```
floooh, line 6465: please remove 1 line of whitespace

sokol_app.h
```diff
@@ -6458,6 +6462,35 @@ _SOKOL_PRIVATE LRESULT CALLBACK _sapp_win32_wndproc(HWND hWnd, UINT uMsg, WPARAM
...
+                        _sapp.win32.dpi.window_scale = (float)dpix / 96.0f;
+                        _sapp.win32.dpi.content_scale = _sapp.win32.dpi.window_scale;
+                        _sapp.dpi_scale = _sapp.win32.dpi.window_scale;
+
+                        LPRECT lpRect = (LPRECT)lParam;
+                        SetWindowPos(hWnd, nullptr, lpRect->left, lpRect->top, lpRect->right - lpRect->left, lpRect->bottom - lpRect->top, SWP_NOZORDER | SWP_NOACTIVATE);
+                    }
+
+                    FreeLibrary(user32);
+                }
```
floooh, line 6489: I think it would be better to wrap this code into a small helper function _sapp_win32_dpi_changed().

sokol_app.h
```diff
@@ -6698,37 +6736,49 @@ _SOKOL_PRIVATE void _sapp_win32_restore_console(void) {
 }
 
 _SOKOL_PRIVATE void _sapp_win32_init_dpi(void) {
-
+    
+    DECLARE_HANDLE(DPI_AWARENESS_CONTEXT_T);
```
floooh, line 6740: Should this be moved up into:

https://github.com/floooh/sokol/blob/b2f60754841c0bbfad47da86974e43e494f11bef/sokol_app.h#L2114-L2127

sokol_app.h
```diff
@@ -6698,37 +6736,49 @@ _SOKOL_PRIVATE void _sapp_win32_restore_console(void) {
...
+            if (!(_sapp.desc.win32_per_monitor_dpi_v2 && fn_setprocessdpiawarenesscontext&&fn_setprocessdpiawarenesscontext((DPI_AWARENESS_CONTEXT_T)-4)))
+                fn_setprocessdpiawareness(process_dpi_awareness); // fallback system dpi aware
         }
-        fn_setprocessdpiawareness(process_dpi_awareness);
     }
     else if (fn_setprocessdpiaware) {
         fn_setprocessdpiaware();
         _sapp.win32.dpi.aware = true;
     }
+
```
floooh, line 6781: whitespace

sokol_app.h
```diff
@@ -6738,6 +6788,7 @@ _SOKOL_PRIVATE void _sapp_win32_init_dpi(void) {
         _SOKOL_UNUSED(hr);
         SOKOL_ASSERT(SUCCEEDED(hr));
         /* clamp window scale to an integer factor */
+
```
floooh, line 6791: whitespace

sokol_app.h
```diff
@@ -6752,6 +6803,7 @@ _SOKOL_PRIVATE void _sapp_win32_init_dpi(void) {
         _sapp.win32.dpi.mouse_scale = 1.0f / _sapp.win32.dpi.window_scale;
     }
     _sapp.dpi_scale = _sapp.win32.dpi.content_scale;
+
```
floooh, line 6806: whitespace

floooh (changes requested):

floooh:
Looks good!

Sorry about all the "remove whitespace" requests.

The new ```win32_default_size``` config setting is unrelated to the DPI changes though right?

floooh:
...I've been sleeping over the default window size thing, and I think it's better to turn this into a cross-platform-thing by simply providing zero-initialized width and height  in sapp_desc. I'll check if the other desktop platforms also provide some sort of default-window-size feature (on web and mobile, the size is ignored anyway).

You don't need to change the PR for this, I'll take care of that during merging.

I'll first take care of https://github.com/floooh/sokol/pull/611 and then start integrating your PR (with the default-size changes).

floooh:
Looks like you are compiling as C++? There's a ```nullptr``` in the call to SetWindowPos ;)

I'll clean up that stuff in a merge branch, I'll probably also move the LoadLibrary() and GetProcAddress() call out of the dpi-changed code path into the initialization code.

Also I didn't notice the -4 constant at first here:

```
fn_setprocessdpiawarenesscontext((DPI_AWARENESS_CONTEXT_T)-4)
```

Is there a human readable name for that constant?

floooh:
Ok, a question:

The per-monitor DPI code-path is currently enabled with ```sapp_desc.win32_per_monitor_dpi_v2```. Is there a situation where one wouldn't want to set this to true? If not I would remove this config setting.

floooh:
Here's my attempt to somewhat simplify the DPI setup code in _sapp_win32_dpi_init(), does that look alright to you?

```c
    /*
        NOTE on SetProcessDpiAware() vs SetProcessDpiAwareness() vs SetProcessDpiAwarenessContext():

        These are different attempts to get DPI handling on Windows right, from oldest
        to newest. SetProcessDpiAwarenessContext() is required for the new
        DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2 method.
    */
    if (fn_setprocessdpiawareness) {
        if (_sapp.desc.high_dpi) {
            /* app requests HighDPI rendering, first try the Win10 Creator Update per-monitor-dpi awareness,
               if that fails, fall back to system-dpi-awareness
            */
            _sapp.win32.dpi.aware = true;
            DPI_AWARENESS_CONTEXT_T per_monitor_aware_v2 = (DPI_AWARENESS_CONTEXT_T)-4;
            if (!(fn_setprocessdpiawarenesscontext && fn_setprocessdpiawarenesscontext(per_monitor_aware_v2))) {
                // fallback to system-dpi-aware
                fn_setprocessdpiawareness(PROCESS_SYSTEM_DPI_AWARE);
            }
        }
        else {
            /* if the app didn't request HighDPI rendering, let Windows do the upscaling */
            _sapp.win32.dpi.aware = false;
            fn_setprocessdpiawareness(PROCESS_DPI_UNAWARE);
        }
    }
    else if (fn_setprocessdpiaware) {
        // fallback for Windows 7
        _sapp.win32.dpi.aware = true;
        fn_setprocessdpiaware();
    }
```

floooh:
FYI: this is the temporary merge branch I'm currently working on:

https://github.com/floooh/sokol/compare/tjachmann-master

tjachmann:
> Looks like you are compiling as C++? There's a `nullptr` in the call to SetWindowPos ;)

 Yep and ops :) 

> fn_setprocessdpiawarenesscontext((DPI_AWARENESS_CONTEXT_T)-4)

You could alternativly replace the beginning of the _sapp_win32_init_dpi like this

#ifndef _DPI_AWARENESS_CONTEXTS_
    DECLARE_HANDLE(DPI_AWARENESS_CONTEXT);
#endif

#ifndef DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2
    #define DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2 (DPI_AWARENESS_CONTEXT)-4
#endif

typedef BOOL(WINAPI * SETPROCESSDPIAWARE_T)(void);
typedef bool (WINAPI * SETPROCESSDPIAWARENESSCONTEXT_T)(DPI_AWARENESS_CONTEXT); // since Windows 10, version 1703`


And then use the DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2 instead of -4. A bit longer, but somehow cleaner.

tjachmann:
> Ok, a question:
> 
> The per-monitor DPI code-path is currently enabled with `sapp_desc.win32_per_monitor_dpi_v2`. Is there a situation where one wouldn't want to set this to true? If not I would remove this config setting.

I also think it should become the default. The only issue I could currently imagine is that when you pass in an concrete window size on creation, then the initial window creating does not respect per-monitor DPI.  More exactly if windows decides to create the window on another monitor the where you query the DPI.

floooh:
I just did this, I think that's alright. It's just the -4 without context as function arg I had a problem with.

```c
DPI_AWARENESS_CONTEXT_T per_monitor_aware_v2 = (DPI_AWARENESS_CONTEXT_T)-4;
```

floooh:
> More exactly if windows decides to create the window on another monitor the where you query the DPI.

Yep, but that's not supported currently anyway, and for this the code which initially queries the DPI would need to be changed anyway. I just removed that config item in the merge branch and I think it's working ok. So far I have only tested on a single-monitor setup by changing the DPI in the system settings, but next I'll check on my HighDPI laptop with a 1080p monitor attached...

tjachmann:
> Here's my attempt to somewhat simplify the DPI setup code in _sapp_win32_dpi_init(), does that look alright to you?
> 
> ```c
>     /*
>         NOTE on SetProcessDpiAware() vs SetProcessDpiAwareness() vs SetProcessDpiAwarenessContext():
> 
>         These are different attempts to get DPI handling on Windows right, from oldest
>         to newest. SetProcessDpiAwarenessContext() is required for the new
>         DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2 method.
>     */
>     if (fn_setprocessdpiawareness) {
>         if (_sapp.desc.high_dpi) {
>             /* app requests HighDPI rendering, first try the Win10 Creator Update per-monitor-dpi awareness,
>                if that fails, fall back to system-dpi-awareness
>             */
>             _sapp.win32.dpi.aware = true;
>             DPI_AWARENESS_CONTEXT_T per_monitor_aware_v2 = (DPI_AWARENESS_CONTEXT_T)-4;
>             if (!(fn_setprocessdpiawarenesscontext && fn_setprocessdpiawarenesscontext(per_monitor_aware_v2))) {
>                 // fallback to system-dpi-aware
>                 fn_setprocessdpiawareness(PROCESS_SYSTEM_DPI_AWARE);
>             }
>         }
>         else {
>             /* if the app didn't request HighDPI rendering, let Windows do the upscaling */
>             _sapp.win32.dpi.aware = false;
>             fn_setprocessdpiawareness(PROCESS_DPI_UNAWARE);
>         }
>     }
>     else if (fn_setprocessdpiaware) {
>         // fallback for Windows 7
>         _sapp.win32.dpi.aware = true;
>         fn_setprocessdpiaware();
>     }
> ```

Looks good

floooh:
Hmm, I need to fix sokol_imgui.h somehow, the scaling is now too big when moving from a HighDPI to a LowDPI monitor. I guess there are some places where the DPI scale factor isn't expected to change.

Interestingly, with the old SetProcessDPIAwareness() it worked as expected. I'll continue working on this stuff for the next few days I think.

floooh:
...looks like image upload is currently borked... I can't even fix the previous comment hmm...

floooh:
...while at it I also implemented per-monitor DPI for macOS (together with a generally more robust dpi_scale handling code).

Also reminder to self: check if on Windows an SAPP_EVENTTYPE_RESIZED is issued when the DPI changes (it should, as a "delayed side effect" of the SetWindowPos() call.

tjachmann:
Ah to compile with legacy SDKs (WINVER < 0x0601), need to check if the DPI message is defined...

#ifndef WM_DPICHANGED
    #define WM_DPICHANGED                   0x02E0
#endif

floooh:
> Ah to compile with legacy SDKs (WINVER < 0x0601), need to check if the DPI message is defined...

Good point, I'll add that. The merge will most likely happen tomorrow btw. Got a bit tangled up with fixing the new timing code when moving the window between displays with a different refresh rate.

floooh:
Ok merged. Changelog entry (18-Jan-2022): https://github.com/floooh/sokol/blob/master/CHANGELOG.md

floooh:
PS: as I mentioned above, I haven't included the default-size handling in this PR. I will come up with a different solution (after looking into the sokol_audio.h sample format problem).

floooh:
@tjachmann FYI, the default-window-size changes are now implemented (see: https://github.com/floooh/sokol/pull/623)

PR #611, opened 2022-01-14
Array uniforms should respect SG_UNIFORMLAYOUT_NATIVE

nmr8acme:
I had a little trouble recently upgrading from sokol_gfx @ bfc41f3f13710f7fc50989215fb844a24466d9bb to aaf2c33278d047404ca434d38804a9e84bb9eda2, as the UB layouts introduced no longer match the expected size and layout for a native GLES2 shader when a float array is used (or any other sub-16-byte primitive array.) IIUC there is just an unhandled case, so I tried to handle it.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -4304,7 +4304,7 @@ _SOKOL_PRIVATE uint32_t _sg_uniform_alignment(sg_uniform_type type, int array_co
     }
 }
 
-_SOKOL_PRIVATE uint32_t _sg_uniform_size(sg_uniform_type type, int array_count) {
+_SOKOL_PRIVATE uint32_t _sg_uniform_size(sg_uniform_type type, int array_count, sg_uniform_layout ub_layout) {
     SOKOL_ASSERT(array_count > 0);
     if (array_count == 1) {
         switch (type) {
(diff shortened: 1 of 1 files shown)


floooh:
Ok merged, many thanks!

I've written a new GLFW sample/test (not committed yet) which tests float, vec2 and vec3 uniform arrays (not pushed yet). Before your PR this would already fail in the sokol-gfx validation layer (wrong uniform block size).

Screenshot 2022-01-15 at 15 17 32

floooh:
PS: I merged the PR with the "command line" procedure, not by pressing the PR Merge button. 

Github seems to be a bit confused (maybe because your git handle "nathan" is different from the Github account "nmr8acme") and doesn't list you in the Contributors tab: https://github.com/floooh/sokol/graphs/contributors

Your commit is properly showing up in the history though (but note how your name isn't clickable):

https://github.com/floooh/sokol/commits/master

Just in case you're wondering what's up with that ;)

floooh:
Link to sample: https://github.com/floooh/sokol-samples/blob/master/glfw/uniformarrays-glfw.c

floooh:
And Changelog update (15-Jan-2022): https://github.com/floooh/sokol/blob/master/CHANGELOG.md

nmr8acme:
Thanks :)

PR #624, opened 2022-01-31
Fix -wmissing-braces warning on Clang

mlabbe:
If building using clang on Windows (as new as 13.0.0) and `-wmissing-braces` is enabled, this pull request prevents three warnings.


floooh:
Thanks! There's also a couple of similar definitions in sokol_audio.h, I'll fix those after merging your PR.

floooh:
No wait, the IIDs in sokol_audio.h already have the braces.

PR #627, opened 2022-02-08
sokol_imgui.h: Updated imgui keypad enter constant

iboB:
The constant in ImGui has been updated in https://github.com/ocornut/imgui/pull/2625

The old one `ImGuiKey_KeyPadEnter` is still present for backwards compatibility but if one uses `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` it is not defined.


floooh:
Thanks! But this will be fixed in a few minutes when I merge the sokol_imgui.h changes for Dear ImGui v1.87, I've updated the whole input code for the new event-based IO and extended virtual key codes.

Issue #301, opened 2020-05-24
Reminder: pause/continue in sokol_app.h


floooh:
See here: https://twitter.com/andsve/status/1264513943051153408

Basically:

```sapp_pause()```s skip calling the frame callback and buffer flip until ```sapp_continue()``` is called. This would be useful (amonst others) for event-driven application which only need to render on input events.

martincohen:
I'm not sure what non-desktop platforms do, but I'd like to point out a few things. (And I'm also sure you already considered these)

Problem I see with this is that you'd lose periodic per-frame `update` call, which might be used to do app-specific optimizations or tasks:
- Syncing threads.
- Running UI animations.
- Detecting whether app is in use (being on front, and last time input events were received).
- Slowing down updates with `sleep` if nothing is happening for a long time. (On Windows you can even [sleep in alertable mode](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleepex))
- Detecting whether the drawing data is different from previous frame, and either doing no draw, partial per-tile draw, or full redraw.

So what I'd propose is a `redraw` flag (perhaps via a function similar to `sapp_consume_event`):
- If the flag is `true`, then `sapp` does buffer swap and sets it to `false`.
- If the flag is `false` then `sapp` does not swap buffers.
- In either case `frame` callback is called.

This approach also makes you not having to worry about `pairing` pause/continue calls.

floooh:
Some more brain-dumping after more feedback from that twitter thread:

- the important part isn't actually to pause/continue the entire frame callback, but to signal to sokol-app whether a frame-swap is needed

- even in "paused state" it would be important to get a regular "tick", and not freeze the application completely (apart form "irregular" events)

For the two reasons above it's probably better to not ever stop calling the frame callback at regular intervals, but instead let the application decide whether it wants to do any rendering (or other expensive stuff) inside the frame callback, and instead tell sokol-app whether a frame-swap is needed.

Some (discarded) alternatives to not calling the frame callback:

- have a "new frame" event which is called each frame before any other events (this is probably still a good idea for other reasons: https://github.com/floooh/sokol/issues/293 and https://github.com/floooh/sokol/issues/283)

- have a separate "paused" callback (not a good idea)

[...]

PR #640, opened 2022-03-08
add sapp_input_wait() to fix high cpu usage while idle

rofl0r:
a sokol_app produces constant cpu load because it keeps repainting
the screen whether something has changed or not. we can fix this
easily by simply calling XNextEvent(), which blocks if there are no
pending events. this causes the app to idle until the user interacts
with the program.

a new function sapp_input_wait() was added to control this behaviour,
which causes the sokol app to block until the next input is received.
afterwards, the input_wait flag is cleared and the user has to set
it again if so desired.

this PR currently only addresses linux, and uses a "dirty" goto, which can be avoided by factoring out the XNextEvent block into a separate function.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1476,6 +1476,8 @@ SOKOL_APP_API_DECL void sapp_request_quit(void);
 SOKOL_APP_API_DECL void sapp_cancel_quit(void);
 /* initiate a "hard quit" (quit application without sending SAPP_EVENTTYPE_QUIT_REQUSTED) */
 SOKOL_APP_API_DECL void sapp_quit(void);
+/* this causes the app to block until input is received. the flag will be cleared after that, so you will have to call it again if necessary. */
+SOKOL_APP_API_DECL void sapp_input_wait(void);
 /* call from inside event callback to consume the current event (don't forward to platform) */
 SOKOL_APP_API_DECL void sapp_consume_event(void);
 /* get the current frame counter (for comparison with sapp_event.frame_count) */
@@ -2517,6 +2519,7 @@ typedef struct {
     bool event_consumed;
     bool html5_ask_leave_site;
     bool onscreen_keyboard_shown;
+    bool input_wait;
     int window_width;
     int window_height;
     int framebuffer_width;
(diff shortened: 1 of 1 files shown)


Manuzor:
Related: https://github.com/floooh/sokol/issues/301

rofl0r:
interesting. the suggested solution there (just skipping draw) doesn't fix the problem entirely - the app will still be executing stuff over and over while "polling" for input. i tried this approach before and it resulted in roughly 4% constant cpu usage even without redrawing and some usleep()s built in.
the only way to get 0% cpu usage is to actually block waiting for events.
related: https://github.com/libsdl-org/SDL/issues/3976

floooh:
Hmm, I could've sworn I had written a detailed reply to this PR already, oh well... ;)

The problem with the PR is that it doesn't fit into frame-callback based sokol_app.h backends (any backend which is not Windows or X11 basically), but I think that's already discussed in the issue thread linked by @Manuzor

rofl0r:
> The problem with the PR is that it doesn't fit into the frame-callback-based sokol_app.h backends 

maybe you're misunderstanding, the function added doesn't block itself, it merely sets a flag which causes the event/frame dispatch loop in the header to block upon next execution.

a typical use case would look like:

```c
// timestamp in microsecs of last user activity
static uint64_t user_activity = -1ULL;

static uint64_t get_ts(void) {
    struct timespec nowtime;
    clock_gettime(CLOCK_REALTIME, &nowtime);
    return nowtime.tv_sec * 1000000ul + nowtime.tv_nsec/1000000ul;
}   
static int need_render(void) {
    if (user_activity == -1ULL || !sim_get_paused()) return 1;
    uint64_t ts = get_ts();
    return (ts - user_activity < 200);
}           
static void app_frame(void) {
    if (!need_render()) {
       sapp_input_wait();
       return;  
    }
    gfx_new_frame(sapp_widthf(), sapp_heightf());
    ui_frame();
 ....
}
static void app_input(const sapp_event* ev) {
    input_handle_event(ev);
    user_activity = get_ts();
}       
sapp_desc sokol_main(int argc, char* argv[]) {
    sargs_setup(&(sargs_desc){ .argc = argc, .argv = argv });
    return (sapp_desc){
        .init_cb = app_init,
        .frame_cb = app_frame,
        .event_cb = app_input,
...
}
```
(from https://github.com/rofl0r/visualz80/blob/nofips/src/main.c )

PR #664, opened 2022-05-16
Add option to select opengl version

### LINKED ISSUE #663: sokol_app: Allow to create OpenGL contexts X.Y
I use sokol_app a lot, and needed to do a OpenGL 4.6 demo. I was able to load my GL functions through glad, no problems there, and then tried to use some GL4.6 functions, but failed. Looking at the context creation the version 3.3 is forced in sokol_app, I changed that and it started working again with no problems.

Would it make sense to allow an extra gl setting in sapp_desc to allow the user override the context? I think I can even do it myself, but I was wondering first if this could be a good/bad idea.

pplux:
Fixes #663 

I tried to make it work on win/linux/mac. But worth double checking.

On the original code for windows/linux it was requestion an opengl 3.3 core context, instead of a 3.2 (which is what from the constants looks it was supposed to do). Not sure what is the proper choice there, but it can be revertec back to 3.3 (or now, let the user select...)

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1433,6 +1433,8 @@ typedef struct sapp_desc {
 
     /* backend-specific options */
     bool gl_force_gles2;                // if true, setup GLES2/WebGL even if GLES3/WebGL2 is available
+    int gl_major_version;               // set GL major version (default: 3)
+    int gl_minor_version;               // set GL minor version (default: 2)
     bool win32_console_utf8;            // if true, set the output console codepage to UTF-8
     bool win32_console_create;          // if true, attach stdout/stderr to a new console window
     bool win32_console_attach;          // if true, attach stdout/stderr to parent process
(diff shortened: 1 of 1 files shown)


floooh:
Thanks! I just started to move tests from the sokol-samples to the sokol directory, which is quite a bit more work than I anticipated (because I need to work with "raw" cmake, not fips).
I'll look into the PR once I'm done with the tests stuff, might take a week or two.

pplux:
sure! no problem :)

floooh:
Found a little flaw:

If version 4.0 is desired, it would instead use version 4.2 :D

I've merged the PR into my own temporary branch, so I'll just fix it there. I'll just look only at the major version to decide if the default 3.2 should be used, and if gl_major_version is != 0 I'll just take the minor version as it is.

I have also 'centralized' the various places where the default is 'patched in' into the function ```_sapp_desc_defaults()```.

floooh:
...and merged... I somehow messed up the git history a bit (the later commits from master which I first merged into my own branch pulled from the PR are now duplicated in the history - but ¯\_(ツ)_/¯).

This is the change I talked about in the earlier comment:

https://github.com/floooh/sokol/blob/8cdce3018e28554830dc72bacebf22209201f89a/sokol_app.h#L2748-L2755

PR #665, opened 2022-05-17
Update Nim bindgen (May 2022)

gustavolsson:
* Follow nim code style
* Avoid keyword clashes
* Remove redundant functions, refactor
* Use nim bitfields for some consts
* Generate bindings for sokol_glue.h (I _think_ this makes it possible to use gfx without app)

Corresponding PR for sokol-nim: https://github.com/floooh/sokol-nim/pull/12

### DIFF
--- a/bindgen/gen_nim.py
+++ b/bindgen/gen_nim.py
@@ -2,8 +2,8 @@
 #   Read output of gen_json.py and generate Zig language bindings.
 #
 #   Nim coding style:
-#   - types and constants are PascalCase
-#   - functions, parameters, and fields are camelCase
+#   - type identifiers are PascalCase, everything else is camelCase
+#   - reference: https://nim-lang.org/docs/nep1.html
 #-------------------------------------------------------------------------------
 import gen_ir
 import json, re, os, shutil
@@ -11,6 +11,7 @@ import json, re, os, shutil
 module_names = {
     'sg_':      'gfx',
     'sapp_':    'app',
+    'sapp_sg':  'glue',
     'stm_':     'time',
     'saudio_':  'audio',
     'sgl_':     'gl',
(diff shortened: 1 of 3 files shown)


bindgen/gen_nim.py
```diff
@@ -252,239 +298,141 @@ def extract_ptr_type(s):
...
         for value in sorted(item_names_by_value):
             name = item_names_by_value[value]
-            l(f"  {name} = {value},")
+            l(f"    {name} = {value},")
     else:
         for name in item_names_by_value.values():
-            l(f"  {name},")
+            l(f"    {name},")
+    if bitfield is not None:
+        l(f"  {enum_name_nim}s = set[{enum_name_nim}]")
```
gustavolsson, line 428: Hmm.. this doesn't seem to work. I thought Nim bitfields were interoperable with C..

gustavolsson, line 428: Hmm.. this doesn't seem to work. I thought Nim bitfields were interoperable with C..

floooh:
Random thought: I wonder if it makes sense to get rid of the glm dependency? In sokol-zig I just wrote a simple math module with just the stuff that's needed (mostly ripped from HandmadeMath.h):

https://github.com/floooh/sokol-zig/blob/master/src/examples/math.zig

garettbass:
I’ve switched to Zig.  No objections from me :)

garettbass:
Have you seen Zig’s `@Vector()`?

floooh:
@garettbass I had tinkered a bit with @Vector in my pacman.zig project, but replaced it with a regular struct because there were some miscompilation issues at the time (see: https://github.com/floooh/pacman.zig/pull/9)

gustavolsson:
> @gustavolsson I'm planning to look into this next, is the PR ready or do you plan any more big changes?
> 
> Also, @garettbass any objections? Do you currently have anything important based on your last changes to the nim bindings?

Nope, all done :)

I wanted to use Nim bitfields for flag enums, bit they don't seem to be compiled to plain integers so it didn't work out. I will create a separate PR if I figure it out!

floooh:
Just a heads up, I will spend a bit more time with the Nim bindings and language bindings in general in the next few days to weeks:

- update the language bindings automatically on commits/merges to the main branch via GH Actions (wip here: https://github.com/floooh/sokol/blob/master/.github/workflows/gen_bindings.yml)
- add Nim output to the sokol-shdc tool (similar to the Zig output mode: https://github.com/floooh/sokol-tools/blob/master/src/shdc/sokolzig.cc)
- ...and I will experiment a bit to restructure the "application model" when using the Nim bindings, I want to find out if it makes more sense to use sapp.run() instead of a sokol_main() function, similar to how it works in Zig (https://github.com/floooh/sokol-zig/blob/3fcc60c1c8be8fee1d5e51f9d76d5b1886725056/src/examples/clear.zig#L31-L43), if this works it would also get rid of the extra ```--noMain``` flag

gustavolsson:
Sounds great! :)

PR #670, opened 2022-05-30
Change GH badges to use relative URLs

cwoffenden:
This is a really minor one-liner but it points the status badge and links for branches and forks to the desired place (and not back to the main repo). In the screenshot here you'll see the URL is going to my fork from the relative URL:

url

(I'm not sure this is documented anywhere, BTW)


cwoffenden:
I found it through trial and error for a work project with multiple forks.

PR #676, opened 2022-06-14
fix(imgui): clipboard paste didn't work on Windows

Namek:
**Bug:** clipboard pasting doesn't work on Windows
**Fix:** change order of event simulation by releasing the V key BEFORE releasing the modifier (CTRL/Cmd) key.

Tested on:
1. Windows 10 21H2 Build 19044.1706
2. MacOS 12.4 Monterey


floooh:
Thanks! I'll look into the PR later today on all platforms.

This was most likely a regression when rewriting the sokol_imgui.h input code to use the new evented IO system in Dear ImGui.

floooh:
Ok, I could reproduce the issue on macOS with the imgui-sapp.cc sample (I'll enable clipboard in the imgui-samples, so that this will be a bit easier to test in the future - at least on the non-web platforms)

PR #678, opened 2022-06-22
sokol_app.h: support for settings x11 cursor

tomc1998:
Add a new api call, sapp_set_cursor

Add some standard cursors, same as GLFW, e.g. `sapp_set_cursor(&SAPP_IBEAM_CURSOR)`

Only implemented on x11, tested on my own (zig) project locally

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1470,6 +1470,21 @@ typedef struct sapp_html5_fetch_request {
     void* user_data;                        /* optional userdata pointer */
 } sapp_html5_fetch_request;
 
+
+typedef struct sapp_cursor sapp_cursor_t;
+
+/* Same set of standard cursors as GLFW */
+extern sapp_cursor_t SAPP_ARROW_CURSOR;
+extern sapp_cursor_t SAPP_IBEAM_CURSOR;
+extern sapp_cursor_t SAPP_CROSSHAIR_CURSOR;
+extern sapp_cursor_t SAPP_POINTING_HAND_CURSOR;
+extern sapp_cursor_t SAPP_RESIZE_EW_CURSOR;
+extern sapp_cursor_t SAPP_RESIZE_NS_CURSOR;
+extern sapp_cursor_t SAPP_RESIZE_NWSE_CURSOR;
+extern sapp_cursor_t SAPP_RESIZE_NESW_CURSOR;
+extern sapp_cursor_t SAPP_RESIZE_ALL_CURSOR;
+extern sapp_cursor_t SAPP_NOT_ALLOWED_CURSOR;
+
 /* user-provided functions */
 extern sapp_desc sokol_main(int argc, char* argv[]);
 
@@ -1509,6 +1524,8 @@ SOKOL_APP_API_DECL bool sapp_mouse_shown(void);
 SOKOL_APP_API_DECL void sapp_lock_mouse(bool lock);
 /* return true if in mouse-pointer-lock mode (this may toggle a few frames later) */
 SOKOL_APP_API_DECL bool sapp_mouse_locked(void);
+  /* set the OS cursor, multiple standard cursors exist already - see e.g. SAPP_ARROW_CURSOR */
+SOKOL_APP_API_DECL void sapp_set_cursor(const sapp_cursor_t *cursor);
 /* return the userdata pointer optionally provided in sapp_desc */
 SOKOL_APP_API_DECL void* sapp_userdata(void);
 /* return a copy of the sapp_desc structure */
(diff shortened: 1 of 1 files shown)


floooh:
FIY I'll start working on this PR now in a branch created from the PR.

One thing I noticed and which I'm going to change is that I'll turn the sapp_cursor_t struct into a simple sapp_cursor enum.

Next I'll look around how to do the Windows and macOS implementation.

Finally I'll change the imgui-samples to use the cursor images.

tomc1998:
@floooh sapp_cursor_t contains the actual platform-independent cursor data/handle, that needs to be a struct not an enum.

`_sapp_x11_create_standard_cursor` is the equivalent to GLFW's `glfwCreateStandardCursor`, but I chose not to expose that to the user & instead just create all the cursors at startup & put them in global vars. Maybe this doesn't suit sokol, in which case remove the globals & expose an `sapp_create_standard_cursor` function like glfw, which switches on an enum value

The 10 'standard' cursors I initialize at startup are taken from the defines at `glfw3.h:1135`, possibly this is more up to date than the docs? limiting sokol's support to the same as GLFW's official docs is probably reasonable though. You can see these are used in `glfw/tests/cursor.c`

floooh:
...ah ok, I've been looking at GLFW 3.3, not 3.4. Ok, I'll use the GLFW 3.4 cursor image types.

I'll still change the public cursor type to an enum, wrapping the platform specific cursor handle in a public type isn't really needed (note that in GLFW, the cursor 'handles' are also just defines: https://github.com/glfw/glfw/blob/da6713cd096a40a4512f468b34c189017e73f987/include/GLFW/glfw3.h#L1161)

I'm still ending up rewriting most of the PR unfortunately, because of the complex relationship between sapp_show/hide_mouse() and the currently set cursor type (on X11, hiding and showing the cursor would always end up calling XUndefineCursor() which I think would switch back to the system default cursor.

The functions _sapp_x11_show_mouse() and _sapp_x11_set_cursor() will basically be merged into a single _sapp_x11_update_cursor() function.

floooh:
You can track my changes here:

https://github.com/floooh/sokol/compare/master...tomc1998-master

I have also added the mouse cursor update to sokol_imgui.h, on Linux this is working fine. Over the next few days I'll try to look into Windows, macOS and HTML5 (I think updating the cursor should also be possible on the web).

This is branched from your PR, so after merging your original contribution should still be properly attributed in the history :)

tomc1998:
Ah go ahead, i don't understand how sokol does other cursor stuff so my code is likely wrong :+1: 

> note that in GLFW, the cursor 'handles' are also just defines

That's what i was trying to explain, these aren't handles! they're just parameters that you pass into glfwCreateStandardCursor to *get* a handle, which contains the platform specific stuff. sapp_cursor_t is the equivalent of GLFWcursor. That's the difference with my impl, I already call 'create_standard_cursor' for all the standard cursors at startup & then expose them as global vars - glfw doesn't do this, and lets you call 'create_standard_cursor' in your own time.

floooh:
Ah ok, I prefer the enum approach though, less private information leaking into the user side of the API. In a later version we could add another function to associate a custom cursor image with a cursor image, a bit similar to the ```sapp_set_icon()``` function (something like 

```
sapp_set_mouse_cursor_image(sapp_mouse_cursor cursor, const sapp_mouse_cursor_desc* desc);
```

floooh:
PS: another nice effect of this PR is that we can hopefully remove the somewhat weird "user cursor" feature (sapp_desc.user_cursor and SAPP_EVENTTYPE_UPDATE_CURSOR), I never particularly liked this, but it was a necessary workaround for switching the cursor image on Windows.

floooh:
Ok, macOS implemented: https://github.com/floooh/sokol/commit/8c1bc368b0b8b13c471d7bdab609e7cd3947908c

I guess the reason why GLFW didn't have the 'diagonal' cursor types was because macOS doesn't expose those in the public API (but everybody is using the private methods anyway)

floooh:
Windows implemented: https://github.com/floooh/sokol/commit/454972323d5902aa38373735283d1b71e5c4bc3e

floooh:
Reminders:
- [ ] HTML5
- [ ] minimal UWP hack (just fix the compilation issue)
- [ ] can remove SAPP_EVENT_
- [ ] update header documentation
- [ ] update changelog

floooh:
HTML 5: https://github.com/floooh/sokol/commit/56d09103606208a84f2496a55dc7c26034997dc6

floooh:
UWP: https://github.com/floooh/sokol/commit/dfc01d45ce79479ced4be413f345adb34e30741e

floooh:
Ok, my branch which I created from your PR has been merged, but this time Github didn't detect the association with the original PR and didn't mark the PR as merged. Your commits are still recorded in the history though.

So I gonna need to "Close" this PR, even though it has been actually merged.

floooh:
...I'll update the changelog next and see how the new enum looks in the Zig and Nim bindings, I priobably need to manually tweak the name mapping.

Issue #681, opened 2022-06-23
sokol_gfx : sg_append_would_overflow proposal


RandyGaul:
Would you be interested in a PR for the function `sg_append_would_overflow`? I'm dynamically streaming transient data to the GPU and could definitely do the bookkeeping logic on my end, but it might be a simple and good function to add to sokol_gfx.

```c
SOKOL_API_IMPL bool sg_append_would_overflow(sg_buffer buf_id, size_t size) {
    SOKOL_ASSERT(_sg.valid);
    _sg_buffer_t* buf = _sg_lookup_buffer(&_sg.pools, buf_id.id);
    bool result = false;
    if (buf) {
        int append_pos = buf->cmn.append_pos;
        /* rewind append cursor in a new frame */
        if (buf->cmn.append_frame_index != _sg.frame_index) {
            append_pos = 0;
        }
        if ((append_pos + _sg_roundup((int)size, 4)) > buf->cmn.size) {
            result = true;
        }
    }
    _SG_TRACE_ARGS(append_would_overflow, buf_id, size, result);
    return result;
}
```

I can make a PR if you think this is a good idea. Any thoughts?

RandyGaul:
Also, I do have a question. I'm doing triple buffering, and sometimes all three buffers can fill up with data in a single frame. Do you know of a simple way to block until the buffers are free for usage? It looks like the default behavior is to silently drop further data appends with sg_append_buffer, and assert on `sg_update_buffer`. Though, I'd like to block and wait, even in debug mode.

floooh:
It's a good idea, but going with the sg_query_* pattern for getting information out of the API, and the existing function ```sg_query_buffer_overflow()``` I think a better name would be ```sg_query_will_buffer_overflow(buf, size)```.

I think the function doesn't need a trace hook (all the other query functions also don't have one).

A PR would be welcome :)

As for the triple buffers filling up question, I don't quite understand the scenario, especially the part about "all three buffers filling up in a single frame", shouldn't only one buffer be used per frame? Also be aware that sokol_gfx.h does buffer renaming under the hood anyway (on GL and Metal, but not D3D11, where D3D does this on its own).

I think the best buffer data updating scenario is to write your per-frame updates into your own memory buffer with enough capacity for the worst case, and on the side record vertex ranges which will be needed later for rendering.

[...]

PR #682, opened 2022-06-24
Added sg_query_buffer_will_overflow

RandyGaul:
As discussed here https://github.com/floooh/sokol/issues/681

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -247,6 +247,11 @@
 
             bool sg_query_buffer_overflow(sg_buffer buf)
 
+        You can manually check to see if an overflow would occur before adding
+        any data to a buffer by calling
+
+            bool sg_query_will_buffer_overflow(sg_buffer buf, size_t size)
+
         NOTE: Due to restrictions in underlying 3D-APIs, appended chunks of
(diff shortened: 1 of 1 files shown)


floooh:
Thanks! I'll spend a few more days with the Nim bindings stuff, and then will look into merging a couple of pending smaller PRs, like this one.

floooh:
FYI I'll start working on the PR now via a local branch, I'll see if I can move some of the code into a common helper function used by both the will-overflow and append-buffer functions.

floooh:
...actually would "sg_query_buffer_will_overflow()" be a better name, because it matches the existing "sg_query_buffer_overflow()" better? (maybe better for searchability / discoverability), I'll change the name in my branch to see how it feels...

floooh:
...ok... trying to merge common code into a helper function isn't really worth it :) I'll keep the name change though!

floooh:
Ok, merged:

- I have renamed the function to sg_query_buffer_will_overflow()
- added a test: https://github.com/floooh/sokol/blob/7f2b86f5bb7145b1d6b2b8d81c81b8356475c1e7/tests/functional/sokol_gfx_test.c#L824-L838
- updated CHANGELOG.md:

PR #687, opened 2022-07-08
Write alpha for ImGui draw ops.

luigi-rosso:
This allows ImGui draw ops to write their alpha values to the framebuffer.

This is nice for when you draw Imgui into one buffer that you later want to blit over the rest of your screen, or similarly if you're drawing with two separate MtkViews with Metal and have one transparently layered over the other.

I ran into a case where I was clearing the frame buffer to fully transparent, drawing imgui onto it and then blitting that framebuffer elsewhere. It was showing up fully transparent (no imgui content visible) because imgui is drawing with no alpha write mask.

### DIFF
--- a/util/sokol_imgui.h
+++ b/util/sokol_imgui.h
@@ -1882,10 +1882,12 @@ SOKOL_API_IMPL void simgui_setup(const simgui_desc_t* desc) {
     pip_desc.sample_count = _simgui.desc.sample_count;
     pip_desc.depth.pixel_format = _simgui.desc.depth_format;
     pip_desc.colors[0].pixel_format = _simgui.desc.color_format;
-    pip_desc.colors[0].write_mask = SG_COLORMASK_RGB;
+    pip_desc.colors[0].write_mask = SG_COLORMASK_RGBA;
     pip_desc.colors[0].blend.enabled = true;
     pip_desc.colors[0].blend.src_factor_rgb = SG_BLENDFACTOR_SRC_ALPHA;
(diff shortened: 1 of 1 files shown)


floooh:
In general a good idea, but it shouldn't be enabled by default because it may have undesired behaviour on some platforms (for instance on the web where the WebGL canvas is always alpha-blended with the background, this was the reason why I limted rendering to RGB in the first place). Please add a new config boolean around here: https://github.com/floooh/sokol/blob/fdd9a15ec28101db3341089af2c7c30b8cd024ff/util/sokol_imgui.h#L301

(I think ```sapp_desc.write_alpha_channel``` would be a good name, and the default behaviour should be that alpha is not written).

Also I think the alpha-blend factors don't do the right thing if you want to compose an ImGui-rendered texture with something else. I would think that the alpha-values during Imgui-rendering need to be added on top of each other (which would be src=one/dst=one).

luigi-rosso:
Great, thanks for the pointers on how to do it! Yeah I'm not sure about the alpha values, but in my simple examples both adding (one/one and srcOver worked), but I didn't try complex multi-blend pixels on the framebuffer...

floooh:
Look good thanks! I'll merge this first and then rebase the mouse-cursor WIP branch on top and merge that afterwards (there will be collisions because I also worked on sokol_imgui.h, but I'll handle those in my branch).

floooh:
And merged, thanks!

floooh:
PS: I'll update the changelog later today together with the mouse cursor changes.

PR #694, opened 2022-07-20
sokol_app.h win32: Calculate hidpi framebuffer size more precisely

ilidemi:
Minor change, switching from int cast to `roundf` and reusing the float intermediate instead of roundtripping through int.

For a window with 850 logical width and 1.75 dpi scale, client rect is 1487px wide. Dividing it back results in window_width of 849.71.
Current logic truncates it to 849 and sets fb_width to 1485px, making it misaligned with the client rect.
After this change window_width is rounded to 850 and fb_width is back to 1487px.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1824,6 +1824,7 @@ inline void sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
     #endif
     #include <stdio.h>  /* freopen_s() */
     #include <wchar.h>  /* wcslen() */
+    #include <math.h>   /* roundf() */
 
     #pragma comment (lib, "kernel32")
     #pragma comment (lib, "user32")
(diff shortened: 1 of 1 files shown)


floooh:
Thanks for the PR! I think the change definitely makes sense. I'll try to look into this some time next week (don't have access to a Windows machine currently), I will probably also look into a similar solution for the other platforms.

ilidemi:
Fixed the header, good catch! Must've been included elsewhere in my app.

floooh:
I will start looking into this now.

floooh:
Ok merged, I also added roundf() to the window and framebuffer size computation on macOS, iOS and Emscripten (although at least on macOS and iOS it should make a difference because the content scale factor isn't fractional).

Thanks for the PR!

PR #657, opened 2022-04-28
[WIP] Update WebGPU backend

floooh:
Open issues:

- [ ] sokol_imgui.h and sokol_nuklear.h need to create and switch between different shaders and pipelines for rendering image widgets, depending on the texture sample type of the image to be displayed.
- [ ] the validation layer currently doesn't seem to catch a sample type mismatch between the current pipeline and bound images?
- [ ] documentation needs to be updated for the new texture sample types
- [ ] sokol_shdc needs a big documentation update (new @texture tag etc)
- [ ] fix remaining deprecation warnings (may require to wait for Emscripten SDK updates)

PR #707, opened 2022-08-23
Add RGB9_E5 texture format

allcreater:
Support for the SG_PIXELFORMAT_RGB9E5 texture format is added.
It is widely used to store HDR textures in a memory efficient manner

The format support has not been added to the WebGPU backend since the WGPUTextureFormat_RGB9E5Ufloat constant is not available in the webgpu headers used in the current WebGPU backend implementation (the feature was introduced a bit later, gpuweb/gpuweb#975). The support can be added right after #657 is completed.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -959,6 +959,8 @@ typedef enum sg_pixel_format {
     SG_PIXELFORMAT_ETC2_RG11,
     SG_PIXELFORMAT_ETC2_RG11SN,
 
+    SG_PIXELFORMAT_RGB9E5,
+
     _SG_PIXELFORMAT_NUM,
     _SG_PIXELFORMAT_FORCE_U32 = 0x7FFFFFFF
 } sg_pixel_format;
@@ -2982,6 +2984,8 @@ inline int sg_append_buffer(sg_buffer buf_id, const sg_range& data) { return sg_
         #define GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS 0x8B4D
         #define GL_R11F_G11F_B10F 0x8C3A
         #define GL_UNSIGNED_INT_10F_11F_11F_REV 0x8C3B
+        #define GL_RGB9_E5 0x8C3D
+        #define GL_UNSIGNED_INT_5_9_9_9_REV 0x8C3E
         #define GL_RGBA32UI 0x8D70
         #define GL_RGB32UI 0x8D71
         #define GL_RGBA16UI 0x8D76
@@ -4552,6 +4556,7 @@ _SOKOL_PRIVATE int _sg_pixelformat_bytesize(sg_pixel_format fmt) {
         case SG_PIXELFORMAT_BGRA8:
         case SG_PIXELFORMAT_RGB10A2:
         case SG_PIXELFORMAT_RG11B10F:
+        case SG_PIXELFORMAT_RGB9E5:
             return 4;
 
         case SG_PIXELFORMAT_RG32UI:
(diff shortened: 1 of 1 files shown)


allcreater:
Hello @flooh!
It would be great to hear your opinion on this improvement.
This format complements other compressed texture formats and simplifies HDR rendering on broad spectrum of devices.
Is there anything else that can be done here?

floooh:
I'm in favour of the PR but need to find some time to tinker with it first :) Currently I need to push through with the new sokol_spine.h header which is a lot more work than I had originally planned/expected. After that I'll take some time to look into PRs.

allcreater:
@floooh First of all, thank you very much for your attention!

Unfortunately, I don't have test images (on the other hand, as far as I know, there are not many standard image formats that support such a specific pixel format, with the possible exception of KTX or DDS), but I wrote (or rather, just ported it [from GLM library](https://github.com/g-truc/glm/blob/cc98465e3508535ba8c7f6208df934c156a018dc/glm/gtc/packing.inl#L615)) this code for converting from RGB32F to RGB9E5, it seems to work on my test images, I hope it will be useful:

```C
float max_float(float a, float b)
{
    return a > b ? a : b;
}

float clamp_float(float x, float min, float max)
{
    return x > max ? max : (x < min ? min : x);
}

bool equal_float(float a, float b)
{
    return abs(a - b) <= 1e-5;
}

struct float3
{
    float x, y, z;
};

uint32_t float3_to_rgb9e5(float3 v)
{
    float const SharedExpMax = (pow(2.0f, 9.0f - 1.0f) / pow(2.0f, 9.0f)) * pow(2.0f, 31.f - 15.f);
    
    float const ColorR = clamp_float(v.x, 0.0f, SharedExpMax);
    float const ColorG = clamp_float(v.y, 0.0f, SharedExpMax);
    float const ColorB = clamp_float(v.z, 0.0f, SharedExpMax);
    
    float const MaxColor = max_float(ColorR, max_float(ColorG, ColorB));

    float const ExpSharedP = max_float(-15.f - 1.f, floor(log2(MaxColor))) + 1.0f + 15.f;
    float const MaxShared = floor(MaxColor / pow(2.0f, (ExpSharedP - 15.f - 9.f)) + 0.5f);
    float const ExpShared = equal_float(MaxShared, pow(2.0f, 9.0f)) ? ExpSharedP + 1.0f : ExpSharedP;

    union u9u9u9e5
    {
        struct
        {
            unsigned int x : 9;
            unsigned int y : 9;
            unsigned int z : 9;
            unsigned int w : 5;
        } data;
        uint32_t pack;
    };
    
    u9u9u9e5 Unpack;
    Unpack.data.x = floor(ColorR / pow(2.f, (ExpShared - 15.f - 9.f)) + 0.5f);
    Unpack.data.y = floor(ColorG / pow(2.f, (ExpShared - 15.f - 9.f)) + 0.5f);
    Unpack.data.z = floor(ColorB / pow(2.f, (ExpShared - 15.f - 9.f)) + 0.5f);
    Unpack.data.w = uint(ExpShared);
    return Unpack.pack;
}

uint32_t half2float(uint32_t h)
{
    return ((h & 0x8000) << 16) | ((( h & 0x7c00) + 0x1C000) << 13) | ((h & 0x03FF) << 13);
}
```

floooh:
Tested on:

- [ ] macOS Metal
- [ ] macOS GL
- [ ] iOS Metal
- [ ] iOS GLES2
- [ ] iOS GLES3
- [ ] Windows D3D11
- [ ] Windows GL
- [ ] Linux GL
- [ ] WebGL
- [ ] WebGL2
- [ ] Android GLES2
- [ ] Android GLES3

floooh:
...interestingly, WebGL2 supports the format (I used the texcube sample and forced the pixel format to RGB9E5)
Screenshot 2022-10-20 at 19 40 13

floooh:
...I actually found a bug in the soko_app.h Android backend when testing, when sokol_app.h was initialized with ```sapp_desc.gl_force_gles2 = true```, a GLES2 context was created (which is correct), but the function ```sapp_gles2()``` still returned false (which was a bug and caused sokol_gfx.h to use the GLES3 code path instead of GLES2).

floooh:
Ok merged, thanks!

I'll update the CHANGELOG.txt next.

PR #710, opened 2022-08-28
gen_zig.py: add compile time branches for self-hosted support

kcbanner:
This adds compile-time switches to have the generated code be compatible with both stage1 and self-hosted.

The `asRange` change has to do with [the self-hosted compiler not passing structs/arrays by reference in all cases](https://github.com/ziglang/zig/issues/12638#issuecomment-1228824773). 

In debugging why the triangle example wasn't rendering, the GL state showed that the gpu-side buffer was filled with bad data (ie. not the verts specified in the source). Upon further inspection, even though the buffer data was being read from the text segment, it was being copied into an array on the stack inside `asRange`, and the ptr to that array was then being returned in the `Range`, and then the data on the stack was getting stomped on the way to the call to upload it to the GPU. 

This does seem like a bug in the compiler at the moment. However, it also sounds like the language spec says that the compiler can either pass by value or by reference depending on what is optimal. So, even though it would likely be optimal to always pass by reference to `asRange`, to comply with the spec I think a pointer to the struct/array has to be passed to `asRange`.  https://github.com/ziglang/zig/wiki/Self-Hosted-Compiler-Upgrade-Guide#escaped-pointer-to-parameter seems to say that passing by pointer is the intended way.

This is a breaking change (I've updated the examples: https://github.com/floooh/sokol-zig/pull/26).

### DIFF
--- a/bindgen/README.md
+++ b/bindgen/README.md
@@ -22,6 +22,7 @@ To update the Zig bindings:
 > cd sokol/bindgen
 > git clone https://github.com/floooh/sokol-zig
 > git clone https://github.com/floooh/sokol-nim
+> git clone https://github.com/floooh/sokol-odin
 > python3 gen_all.py
 ```
 
--- a/bindgen/gen_zig.py
+++ b/bindgen/gen_zig.py
@@ -377,7 +377,8 @@ def gen_struct(decl, prefix):
         elif is_const_prim_ptr(field_type):
             l(f"    {field_name}: ?[*]const {as_zig_prim_type(extract_ptr_type(field_type))} = null,")
         elif is_func_ptr(field_type):
-            l(f"    {field_name}: ?fn({funcptr_args_c(field_type, prefix)}) callconv(.C) {funcptr_result_c(field_type)} = null,")
+            sig = f"fn({funcptr_args_c(field_type, prefix)}) callconv(.C) {funcptr_result_c(field_type)}"
+            l(f"    {field_name}: (switch (builtin.zig_backend) {{ .stage1 => ?{sig}, else => ?*const {sig} }}) = null,")
         elif is_1d_array_type(field_type):
             array_type = extract_array_type(field_type)
             array_sizes = extract_array_sizes(field_type)
@@ -481,6 +482,7 @@ def pre_parse(inp):
                 enum_items[enum_name].append(as_enum_item_name(item['name']))
 
 def gen_imports(inp, dep_prefixes):
+    l('const builtin = @import("builtin");')
     for dep_prefix in dep_prefixes:
         dep_module_name = module_names[dep_prefix]
         l(f'const {dep_prefix[:-1]} = @import("{dep_module_name}.zig");')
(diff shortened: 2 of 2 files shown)


floooh:
Thanks for all the work :)

So the meta.FnPtr() makes it compatible both with the new self-hosted syntax and the old 9.x compiler? If so very cool, because that would mean that there doesn't need to be a separate branch of the bindings for the self-hosted compiler until 10.0 is released).

I guess/hope the pass-by-value vs -reference won't be a problem in other places, like:

```zig
pub fn makeBuffer(desc: BufferDesc) Buffer {
    return sg_make_buffer(&desc);
}
```
...because that doesn't return any pointers to stack data(?).

I hope taking the address of a function argument will be allowed in the future though, and that the compiler instead produces an error when trying to return a pointer to stack data (this is also a dangerous trap in much simpler situations currently).

I'll need to tinker with the PR a bit before merging (especially try out the asRange() change). I'll do the merge probably some time next week.

Cheers!

kcbanner:
> Thanks for all the work :)
> 
> So the meta.FnPtr() makes it compatible both with the new self-hosted syntax and the old 9.x compiler? If so very cool, because that would mean that there doesn't need to be a separate branch of the bindings for the self-hosted compiler until 10.0 is released).

No problem, thanks for sokol and these bindings! 

Yes, the idea behind `meta.FnPtr` is just to add on the `*const ` to the definition to comply with the new syntax rules. However, I didn't realize it hadn't made it to the release zips yet - I was working from zig master looking into the instancing stack overflow issue.

> I guess/hope the pass-by-value vs -reference won't be a problem in other places, like:
> 
> ```zig
> pub fn makeBuffer(desc: BufferDesc) Buffer {
>     return sg_make_buffer(&desc);
> }
> ```
> 
> ...because that doesn't return any pointers to stack data(?).

Yea, in this case it's fine because desc lives (either internally as a copy or a reference) until the end of `makeBuffer`, and `sg_make_buffer` doesn't save that pointer to the desc it was passed.

> I hope taking the address of a function argument will be allowed in the future though (for instance in Odin this isn't allowed, which is a bit awkward because one needs to create an explicit copy of the function arg first), and that the compiler instead produces an error when trying to return a pointer to stack data (this is also a dangerous trap in much simpler situations currently).

I think it will always be allowed to take the address, but it seems there are plans to do escape analysis to make sure you don't return that address and create essentially a use-after-free. It's mentioned here: https://github.com/ziglang/zig/issues/3180#issuecomment-680259699. I believe it will also always be a const ptr when you take the address of a parameter.

> I'll need to tinker with the PR a bit before merging (especially try out the asRange() change). I'll do the merge probably some time next week.
> 
> Cheers!
> 
> PS: ah ok, looks like it still needs a recent compiler for the meta.Fn() to work:
> 
> https://github.com/floooh/sokol/runs/8056635032?check_suite_focus=true
> 
> I need to think a bit about how to handle this, so that the sokol-zig main branch is compatible with 9.x until 10.x is released, and the code-gen changes are instead placed into a zig-10.x branch in the meantime... I'll probably need to keep your changes in a separate branch and update the bindings manually, otherwise the automatic bindings generation (via GH Actions) would update the sokol-zig repo with a version that doesn't work in the current stable zig version.

Yea this might be tricky since the mechanisms for switching on `stage1` aren't there in 0.9.x. Makes sense to wait to go to the main branch. The examples do compile/run with `-fstage1` on 0.10.x against the new bindings, conveniently passing arrays/structs as a pointer uses the previously existing pointer branch so that works fine. However, that still requires `meta.FnPtr` to exist. 

Cheers!

floooh:
...I'll look into merging this PR into a zig-0.10.0 branch now, and update the sokol-zig bindings manually (also in a separate zig-0.10.0 branch).

floooh:
Ok, the content of this PR is now sitting in a branch ```zig-0.10.0```, waiting until zig 0.10.0 officially releases. Once that happens I'll merge the ```zig-0.10.0``` branch into master and hopefully this PR will then resolve automatically too. I have converted the PR to 'draft' so that I don't accidentially hit the merge button :)

Thanks for the PR!

floooh:
Ok, and this one is merged too, just waiting for the bindings CI pipeline to verify that everything works as expected.

Issue #708, opened 2022-08-26
On Android, make the native activity available before sokol_main is called


neslib:
I am writing a Sokol language binding for Delphi and it works great on Windows, macOS and iOS.

However, for Android, I need to know the native activity at the moment `sokol_main` is called, but `ANativeActivity_onCreate` sets the native activity *after* `sokol_main` returns. I see 2 solutions for this:

1. Set the native activity *before* `sokol_main` is called, so I can use `sapp_android_get_native_activity` to retrieve it. However, the application is not valid yet at this point, so `sapp_android_get_native_activity` currently raises an assertion.

2. Pass the native activity as an extra parameter to the `sokol_main` function. However, this breaks compatibility with existing code.

floooh:
Hi, apologies for late reply. I'm actually not all that familiar with the Android backend, but I think it should be possible to move the line:

```
_sapp.android.activity = activity;
```
...before the call to sokol_main() and remove the assert from sapp_android_get_native_activity().

I agree that it would be better to pass some sort of struct into sokol_main() which has additional information (not just the cmdline args), but it's not worth it yet to break compatibility for this. I'll try to keep it in mind for the next compatibility breaking update.

Would you like to provide a PR with the 'simple' fix?

neslib:
Done. And thanks for this amazing library!
I'm almost done with the Delphi language binding and this was the only issue standing in the way.

floooh:
Closing this as the PR is merged.

PR #712, opened 2022-09-02
Made Android native activity available before calling sokol_main (#708)

neslib:
Simple fix for issue #708

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -9194,11 +9194,14 @@ void ANativeActivity_onCreate(ANativeActivity* activity, void* saved_state, size
     _SOKOL_UNUSED(saved_state_size);
     SOKOL_LOG("NativeActivity onCreate()");
 
+    // Set activity *before* calling sokol_main.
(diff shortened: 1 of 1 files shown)


floooh:
One thing that I just noticed now is that the call to ```_sapp_init_state()``` here:

https://github.com/floooh/sokol/blob/01eeb35658fee9b6fe2e31145925fbd5e22faf47/sokol_app.h#L9198

...clears the entire _sapp struct (here: https://github.com/floooh/sokol/blob/01eeb35658fee9b6fe2e31145925fbd5e22faf47/sokol_app.h#L2821), which deletes the ```_sapp.android.activity``` pointer again. I'll think of a solution when I work on the merge (worst case is that I'll just write the activity pointer again).

I'm sure this clearing was also the main reason why the activity pointer was written so late.

floooh:
Ok merged, I added a couple of comments, and set the native activitiy pointer before *and* after _sapp_init_state().

...also tested on my Pixel 4.

PR #717, opened 2022-09-18
sokol_app.h: add a `tool_type` field to the `sapp_touchpoint` struct (valid on Android)

Wertzui123:
This makes it possible to differentiate between different types of touch input.
See: https://developer.android.com/ndk/reference/group/input#amotionevent_gettooltype for the documentation of this function
I've tested it and it successfully showed 1 for finger input and 2 for S-Pen input.

I'm unsure if adding a new field to the event struct is the way to go; maybe we should provide the user directly with the `AMotionEvent` and the `pointer_index` (`e` and `idx` variables) instead so that they can then use arbitrary NDK functions in their callbacks.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1336,6 +1336,7 @@ typedef struct sapp_event {
     float mouse_dy;                     // relative vertical mouse movement since last frame, always valid
     float scroll_x;                     // horizontal mouse wheel scroll distance, valid in MOUSE_SCROLL events
     float scroll_y;                     // vertical mouse wheel scroll distance, valid in MOUSE_SCROLL events
+    int touch_tool_type;                // the type of tool (e.g. finger, pen, mouse) that caused the touch event, valid in TOUCHES_BEGAN, TOUCHES_MOVED, TOUCHES_ENDED
     int num_touches;                    // number of valid items in the touches[] array
     sapp_touchpoint touches[SAPP_MAX_TOUCHPOINTS];  // current touch points, valid in TOUCHES_BEGIN, TOUCHES_MOVED, TOUCHES_ENDED
     int window_width;                   // current window- and framebuffer sizes in pixels, always valid
@@ -8854,6 +8855,7 @@ _SOKOL_PRIVATE bool _sapp_android_touch_event(const AInputEvent* e) {
             dst->changed = true;
         }
     }
+    _sapp.event.touch_tool_type = AMotionEvent_getToolType(e, (size_t)idx);
     _sapp_call_event(&_sapp.event);
     return true;
 }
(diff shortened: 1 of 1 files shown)


billzez:
I think it might make more sense to emit mouse events so that Android matches other platforms, and we don't have to use two different APIs to handle mouse input.

Wertzui123:
No, I don't think so. Afaik, mouses on Android already emit a mouse event, but touch should remain touch.

Wertzui123:
I've noticed my previous implementation (a field in the event struct) didn't make much sense in the multi-touch architecture of `sokol_app.h`. 
Instead, I have changed it to a field of the `sapp_touchpoint` struct now, which should also be a smaller bc break for wrappers.

billzez:
Actually, android has a single API for both mouse and touch, and calls them "pointers".  Currently sokol creates touch events, but not mouse events for android.

Wertzui123:
Oh, okay, I didn't know that. Well, on one hand, it makes sense, but on the other hand, you normally have only 1 mouse, but multiple fingers, so it maybe makes sense to separate it.

floooh:
I'll start looking into this PR now.

@billzez: I'm not a big fan of unified mouse and and touch events, for instance UWP also only has this sort of "pointer events", and it's quite a mess. It's better if the application decides to explicitely support touch vs mouse input, and if it wants to unify those two.

floooh:
@Wertzui123: Do you know if there's a documentation somewhere what the integer return value from AMotionEvent_getToolType() actually means? I have a hard time finding any useful documentation.

Also, I'll rename the field to ```android_tool_type```, to make clear that it is only populated on Android.

Wertzui123:
> Do you know if there's a documentation somewhere what the integer return value from AMotionEvent_getToolType() actually means? I have a hard time finding any useful documentation.

I think you are looking for this: https://developer.android.com/ndk/reference/group/input#anonymous-enum-38

> Also, I'll rename the field to android_tooltype, to make clear that it is only populated on Android.

I used a generic name since we may want to add iOS support in the future too (if that's possible), but I'm fine with adding an `android` prefix for now.

> I'll add an `sapp_android_tooltype` enum with those values, and replace the plain integer with the enum.

Good idea.

floooh:
Please ignore the Github Actions error, this seems to be a regression in Nim.

floooh:
> since we may want to add iOS support in the future too

I did some quick googling too but haven't found anything (neither iOS nor web APIs). If iOS provides that sort of information then we should go back from the ```android``` prefixes to generic names, but as long as it's not a multi-platform feature it's better to keep the prefix.

floooh:
Ok, merged, I did the following changes on top:

https://github.com/floooh/sokol/commit/2a1efce3557c489cf958e7412c9344518d1b2365

(argh, extra whitespace line, I'll remove that too...)

Wertzui123:
Thanks for merging!

PR #718, opened 2022-09-20
Add macro to reduce repeated code.

wtholliday:
Just a little cleanup.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -13171,372 +13171,111 @@ _SOKOL_PRIVATE void _sg_wgpu_update_image(_sg_image_t* img, const sg_image_data*
 #endif
 
 /*== BACKEND API WRAPPERS ====================================================*/
+
+#if defined(_SOKOL_ANY_GL)
+#define _SOKOL_CHOOSE_BACKEND(name)  _sg_gl_##name
+#elif defined(SOKOL_METAL)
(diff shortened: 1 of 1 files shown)


floooh:
While it reduces the line count (which usually a good thing) I'm not a fan of macro magic like this, for instance when step-debugging, the debugger just sits on the opaque macro instead of the specific function call when stepping through the code, same with line numbers in compile errors.

Many thanks for the PR, but I'm going to close it :)

wtholliday:
Ok, cool. I get the consideration about step debugging, but I don't get the part about error line numbers. Were you concerned about where that #error is?

floooh:
It's not really relevant for this specific case, but compile errors in macros can be a bit hard to understand because the error happens in the unexpanded code (some compilers provide a hint though that the errors happened during the expansion of a macro and provide some additional info for context).

wtholliday:
Ok, so for this it's just about step debugging. I guess if you're unfamiliar with the code, when you land on the macro line, you wonder what the macro does, so you look at it and see it just expands to a single call depending on the backend, then you step into it. You would know not to step over it. Maybe you don't look at the macro and step into, which would take you to the backend-specific function, and that could be confusing? I dunno, really seems pretty minor on the macro-magic end of things compared to other stuff I've seen. Anyway, your call :)

PR #719, opened 2022-09-27
Remove wrapper procs for Odin bindings. Use c.int type

thePHTest:
With Odin's new support for `#by_ptr` to represent const references for bindings, we can now generate the bindings without wrapper procs by also using `link_prefix` for the foreign block. I also made the changes to use `c.int` where sokol uses `int` in the C headers to be more idiomatic.

A few of the example programs in `sokol-odin` will need to be updated to remove casts to `int` once this is merged.

### DIFF
--- a/bindgen/gen_odin.py
+++ b/bindgen/gen_odin.py
@@ -330,17 +330,6 @@ def funcdecl_args_c(decl, prefix):
             s += f"{param_name}: {map_type(param_type, prefix, 'c_arg')}"
     return s
 
-def funcdecl_args_odin(decl, prefix):
-    s = ''
-    func_name = decl['name']
-    for param_decl in decl['params']:
-        if s != '':
-            s += ', '
-        param_name = param_decl['name']
-        param_type = check_override(f'{func_name}.{param_name}', default=param_decl['type'])
-        s += f"{param_name}: {map_type(param_type, prefix, 'odin_arg')}"
-    return s
-
 def funcptr_args_c(field_type, prefix):
     tokens = field_type[field_type.index('(*)')+4:-1].split(',')
     s = ''
@@ -366,12 +355,6 @@ def funcdecl_result_c(decl, prefix):
     decl_type = decl['type']
     res_c_type = decl_type[:decl_type.index('(')].strip()
     return map_type(check_override(f'{func_name}.RESULT', default=res_c_type), prefix, 'c_arg')
-
-def funcdecl_result_odin(decl, prefix):
-    func_name = decl['name']
-    decl_type = decl['type']
-    res_c_type = decl_type[:decl_type.index('(')].strip()
-    return map_type(check_override(f'{func_name}.RESULT', default=res_c_type), prefix, 'odin_arg')
 
 def get_system_libs(module, platform, backend):
     if module in system_libs:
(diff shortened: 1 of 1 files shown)


bindgen/gen_odin.py
```diff
@@ -417,15 +405,25 @@ def gen_c_imports(inp, prefix):
...
     prefix = inp['prefix']
     for decl in inp['decls']:
         if decl['kind'] == 'func' and not decl['is_dep'] and not check_ignore(decl['name']):
             args = funcdecl_args_c(decl, prefix)
             res_type = funcdecl_result_c(decl, prefix)
             res_str = '' if res_type == '' else f'-> {res_type}'
-            l(f"    {decl['name']} :: proc({args}) {res_str} ---")
+            # Need to special case sapp_sg to avoid Odin's context keyword
+            if c_prefix == "sapp_sg":
+                l(f'    @(link_name="{decl["name"]}")')
```
thePHTest, line 423: Another option here would be to remove using `link_prefix` and then always use `@(link_name=...)` with `check_override(as_snake_case(...))` . This way there wouldn't need to be a special case for `sapp_sg`.

thePHTest:
With the latest Odin release now out, #by_ptr will be available with the github actions setup for sokol-odin

floooh:
I'll look into this PR next, might take couple of days though.

floooh:
> A few of the example programs in sokol-odin will need to be updated to remove casts to int once this is merged.

Hmm, it seems the other way around, without the separate wrapper functions I now need a cast from ```int``` to ```i32``` (i32 seems to be equivalent with c.int):

```
...debugtext/main.odin(48:15) Cannot assign value 'font_index' of type 'int' to 'i32' in argument
```

IIRC this was actually the reason why I used extra wrapper functions, so that the 'user code' can use "regular" types like ```int``` instead of specific types like ```i32```.

Thoughts?

thePHTest:
That may have been one of the reason for the wrappers. The other reason I saw was to pass things by value with the wrapper procs, e.g.:
```
build_plane :: proc(buf: Buffer, params: Plane) -> Buffer {
    _buf := buf
    _params := params
    return sshape_build_plane(&_buf, &_params)
}
```

For those compiler errors of the debugtext example, I would just change the definition of `print_font` to take an `i32` as the first parameter. And then the generated `printf` will require the cast from int to i32. (I can update the PR for that if you'd like)

So for Odin, my understanding is that `int` is register sized so the size may change depending on the target platform (see this issue for example https://github.com/odin-lang/Odin/issues/2091). And `c.int` is currently defined in `odin/core/c/c.odin` as `int :: builtin.i32` and is the idiomatic way to bind to C's `int` type

floooh:
> The other reason I saw was to pass things by value with the wrapper procs

Ah yeah, right, that was the main reason. I'll tinker around a bit with the samples tomorrow, I guess replacing int with i32 isn't such a big deal, and using i32 instead of int is required in structs anyway (so that they are layout-compatible).

thePHTest:
Ah, I just remembered that there is the `#any_int` directive which could be helpful.
Seems to be a common pattern in the core library for length parameters.
e.g.
```
core/runtime/core_builtin.odin:make_slice :: proc($T: typeid/[]$E, #any_int len: int, allocator := context.allocator, loc := #caller_location) -> (T, Allocator_Error) #optional_allocator_error
```

floooh:
PS: I also found another thing which must be fixed in the code generation, an injected helper function needs a cast to i32 for the ```len(fstr)``` call (like this):

```python
def gen_helpers(inp):
    if inp['prefix'] == 'sdtx_':
        l('import "core:fmt"')
        l('import "core:strings"')
        l('printf :: proc(s: string, args: ..any) {')
        l('    fstr := fmt.tprintf(s, ..args)')
        l('    putr(strings.unsafe_string_to_cstring(fstr), i32(len(fstr)))')
        l('}')
```

(PS: this ```#any_int``` looks indeed very useful)

floooh:
...yeah we really need to find a better solution for int vs i32, for instance here where the loop variable gets an implicit int type, a cast is needed for calling ```sdtx.font(i)``` (this is from the debugtext-print sample):

```odin
    for i in 0..<NUM_FONTS {
        color := state.palette[i]
        sdtx.font(i)
        ...
```
Could you check if this ```#any_type``` thingy helps with this, and update the PR?

floooh:
...hm I can't seem to build my own version of Odin on an M1 Mac which produces executables for the sokol samples that don't produce a segfault, so bisecting is a bit tricky :/

thePHTest:
I'll update the PR with `#any_int` usage where appropriate. Maybe I'll just look for parameter names with `len` or `index` .
As for M1 mac, I am on Windows and can't really help there. Searching through the Odin discord it seems like there were some bugs with that platform fixed towards the end of July. I'm sure Bill and the folks on the discord server would have more insight and be able to get to the root of the issue.

thePHTest:
Seems to work fine on Windows with the updated bindings fwiw:
image

floooh:
> I'll update the PR with #any_int usage where appropriate. Maybe I'll just look for parameter names with len or index .

Would it make sense to use #any_int for any integer function parameter? The parameter names are not supposed to carry any meaning :)

As for the M1 bug, it's most likely some sort of ABI problem when Odin is calling across into C code. I've seen this sort of stuff quite often on Zig too, mostly after LLVM had been updated.

thePHTest:
Ah that makes sense that the specific sizes are used when needed. I just updated the PR to use `#any_int` along with `c.int` parameters. Looks like now all of the examples work without updates besides `examples/mrt` . That one will just need an update to it's internal proc of `create_offscreen_pass` to use `#any_int` .

Looks like Bill just made a commit to fix an ABI bug with AMD64 SysV fyi: <https://github.com/odin-lang/Odin/commit/419eab50590ded8777f55f5b29306201a85162c5> relating to this bug <https://github.com/odin-lang/Odin/issues/2121> So you may be able to build with the latest now.

floooh:
Ok merged, thanks! I had to do some small fixes in the mrt sample (actually removing some casts, so that's a good thing).

PR #721, opened 2022-10-03
Replace SOKOL_LOG with runtime callbacks.

### LINKED ISSUE #720: Replacing SOKOL_LOG with a runtime callback
Following up on [sokol-zig issue 27](https://github.com/floooh/sokol-zig/issues/27) about replacing `SOKOL_LOG(s)` with a runtime callback. I wasn't sure if you'd be willing to accept a PR for this or if you wanted to do this yourself at some point. Either way, I wanted to see how hard it would be so I did it anyway on sokol_gfx and sokol_app and it seems pretty straight forward with some minor caveats.

Here's a rough sketch of what I did:

* Add new field `logger` of type `sg_logger` to `sg_desc`.
* Create new private function `_sg_log(s)` that forwards the message to `_sg.desc.logger...` if the callback is valid, otherwise it calls `puts(s)`.
* Replace `SOKOL_LOG(s)` with `SG_LOG(s)`, which calls `_sg_log` in non-debug code, and is a no-op otherwise.
* Add deprecation error `#ifdef SOKOL_LOG\n#error ...` to give existing code a chance to notice the API change.
* Do the same thing in sokol_app (just with `s/SG_/SAPP_/g` and `s/sg_/sapp_/g`).

So far so good. There are some open questions to this, however.

* Should there be a way to disable logging altogether as a build-option? Something like `#define SOKOL_NO_LOGGING`?
* The `SOKOL_LOG` implementation of sokol_app.h calls `__android_log_write` on Android targets, but the other headers haven't adopted this, as far as I can tell (so as of right now, the order in which you include the sokol headers on android determine whether log messages go to stdout or the android log...). Should this be adopted with this callback approach and be part of the fallback case of the `_sg_log` implementation? So instead of calling `puts`, on Android we'd call `__android_log_write`. Or do we push the responsibility for this back to the user and have them implement it in their log callback? If we go with the former, we could also add `OutputDebugStringA` on Windows which would get rid of all the `#define SOKOL_LOG(s) OutputDebugStringA(s)` in the samples and the sokol-zig version of `sokol.c`. (By the way, `OutputDebugStringA` does *not* append a newline, so you'd actually have to do `#define SOKOL_LOG(s) { OutputDebugStringA(s); OutputDebugStringA("\n") }` for properly formatted output.)
* In sokol_app, there's `sapp_desc.fail_cb` and `sapp_desc.fail_userdata_cb` as well as `_sapp_fail(s)`. Should these be replaced with the new log callbacks? Or should failure stuff remain separate?

Manuzor:
Removes SOKOL_LOG from sokol_fontstash.h and sokol_args.h entirely because it's not used there.

Please note that merging this as-is will break the binding repos for windows because they do `#define SOKOL_LOG(msg) OutputDebugStringA(msg)`. Here it is for sokol-zig:

https://github.com/floooh/sokol-zig/blob/2bd3b4f44ae137858dbfd5038db40c40563aa86d/src/sokol/c/sokol_defines.h#L5

This should either be removed, or replaced with something like this:

```c
#define SG_LOG_DEFAULT(msg) { OutputDebugStringA("sg: "); OutputDebugStringA(msg); OutputDebugStringA("\n"); }
#define SAPP_LOG_DEFAULT(msg) { OutputDebugStringA("sapp: "); OutputDebugStringA(msg); OutputDebugStringA("\n"); }
#define SAUDIO_LOG_DEFAULT(msg) { OutputDebugStringA("saudio: "); OutputDebugStringA(msg); OutputDebugStringA("\n"); }
#define SGL_LOG_DEFAULT(msg) { OutputDebugStringA("sgl: "); OutputDebugStringA(msg); OutputDebugStringA("\n"); }
#define SDTX_LOG_DEFAULT(msg) { OutputDebugStringA("sdtx: "); OutputDebugStringA(msg); OutputDebugStringA("\n"); }
#define SFETCH_LOG_DEFAULT(msg) { OutputDebugStringA("sfetch: "); OutputDebugStringA(msg); OutputDebugStringA("\n"); }
```

Closes #720

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -28,7 +28,6 @@
     Optionally provide the following defines with your own implementations:
 
         SOKOL_ASSERT(c)     - your own assert macro (default: assert(c))
-        SOKOL_LOG(msg)      - your own logging function (default: puts(msg))
         SOKOL_UNREACHABLE() - a guard macro for unreachable code (default: assert(false))
         SOKOL_ABORT()       - called after an unrecoverable error (default: abort())
         SOKOL_WIN32_FORCE_MAIN  - define this on Win32 to use a main() entry point instead of WinMain
@@ -204,7 +203,7 @@
             The fail callback is called when a fatal error is encountered
             during start which doesn't allow the program to continue.
             Providing a callback here gives you a chance to show an error message
-            to the user. The default behaviour is SOKOL_LOG(msg)
+            to the user. The default behaviour is SAPP_LOG(msg)
 
         As you can see, those 'standard callbacks' don't have a user_data
         argument, so any data that needs to be preserved between callbacks
@@ -1029,7 +1028,7 @@
                 .allocator = {
                     .alloc = my_alloc,
                     .free = my_free,
-                    .user_data = ...;
+                    .user_data = ...,
                 }
             };
         }
@@ -1038,6 +1037,29 @@
 
     This only affects memory allocation calls done by sokol_app.h
     itself though, not any allocations in OS libraries.
+
+
+    LOG FUNCTION OVERRIDE
+    =====================
+    You can override the log function at initialization time like this:
+
+        void my_log(const char* message, void* user_data) {
+            printf("sapp says: \s\n", message);
+        }
+
+        sapp_desc sokol_main(int argc, char* argv[]) {
+            return (sapp_desc){
+                // ...
+                .logger = {
+                    .log_cb = my_log,
+                    .user_data = ...,
+                }
+            };
+        }
+
+    If no overrides are provided, puts will be used on most platforms.
+    On Android, __android_log_write will be used instead.
+
 
     TEMP NOTE DUMP
     ==============
@@ -1424,6 +1446,17 @@ typedef struct sapp_allocator {
     void* user_data;
 } sapp_allocator;
 
+/*
+    sapp_logger
+
+    Used in sapp_desc to provide custom log callbacks to sokol_app.h.
+    Default behavior is SAPP_LOG_DEFAULT(message).
+*/
+typedef struct sapp_logger {
+    void (*log_cb)(const char* message, void* user_data);
+    void* user_data;
+} sapp_logger;
+
 typedef struct sapp_desc {
     void (*init_cb)(void);                  // these are the user-provided callbacks without user data
     void (*frame_cb)(void);
@@ -1453,6 +1486,7 @@ typedef struct sapp_desc {
     int max_dropped_file_path_length;   // max length in bytes of a dropped UTF-8 file path (default: 2048)
     sapp_icon_desc icon;                // the initial window icon to set
     sapp_allocator allocator;           // optional memory allocation overrides (default: malloc/free)
+    sapp_logger logger;                 // optional log callback overrides (default: SAPP_LOG(message))
 
     /* backend-specific options */
     bool gl_force_gles2;                // if true, setup GLES2/WebGL even if GLES3/WebGL2 is available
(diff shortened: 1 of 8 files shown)


floooh:
I only had a quick glance, but looking good. I'll try to take care of smaller PRs again in a couple of days.

Incidentally I started thinking about a different unified error/warning mechanism for the sokol headers, I'll try this out in sokol_spine.h first, for other headers it would be a bit more work (because a lot of error checking code that currently aborts or results in 'undefined behaviour' when asserts are compiled out needs to be rewritten).

Basic idea being that the sokol headers should (almost) never abort/panic on errors, but record any warnings and errors in an internal buffer and then try to carry on (essentially skipping any API calls once a 'fatal' error was recorded). If the API user is interested in error information they would drain the error buffer at at their own place and frequency (e.g. it's not required to call a 'get error' function after each API call that could go wrong). An error would at least contain an error code and a line number, and in debug mode als a human-readable error message, function name and file name (so that it's possible to output 'clickable' error messages).

But as I said, that's a bigger topic to get right across all headers, so for the time being it makes more sense to merge your PR.

floooh:
Just a little followup to my previous comment, and after a little walk to think about error handling: I think instead of an internal error buffer I will just call a more 'structured' logging callback which instead of just a string message gets the following information:

- a 'tag' string which identifies the header (e.g. the current prefixes, ```sg```, ```sapp```, ```sgl``` etc...
- a 'level' (panic, error, warning - where panic is so serious that the error callback should terminate the program because the code can't continue in any way - this should absolutely be the last resort though)
- a unique error code
- the line number (__LINE__) where the error occured (or rather the error logging function was called)
- an optional human readable error message (debug mode only)
- the optional header's filename (__FILE__) (also debug mode only, to construct 'clickable' error messages)

The actual logging output can then be provided by a separate sokol header which plugs directly into the logging callback.

As I said, I'll tinker with this idea in the new sokol_spine.h header, and until this new system is implemented I'll merge your logging callback PR :)

floooh:
I was actually thinking about passing a struct to the logging callback, but the downside of this is that each header needs to define its own struct, and that makes it impossible (without dangerous casting) to use the same logging callback across all sokol headers. With distinct args all headers can define the same callback function signature (without any header-specific types) and it's possible to use the same callback function across all headers.

It's better if you keep the PR as is, I will try to merge this 'soon-ish', and it will be quite a while before I get around to add the 'new logging' to all headers anyway.

floooh:
@Manuzor I'll try to get this PR merged some time this week. I've been thinking a bit more about it though, and think we should make it compatible with SOKOL_LOG somehow. Not sure what the 'least intrusive' is in your PR though, any ideas?

Would it make sense to just replace the new ```*_LOG_DEFAULT``` with the old SOKOL_LOG?

Also, concerning Windows, since this special case already exists for Android, it probably makes sense to add another one for Windows:

```c
        #if defined(__ANDROID__)
            #include <android/log.h>
            #define SAUDIO_LOG_DEFAULT(s) __android_log_write(ANDROID_LOG_INFO, "SOKOL_AUDIO", s)
        #else
            #include <stdio.h>
            #define SAUDIO_LOG_DEFAULT(s) puts(s)
        #endif
```
...e.g. something like this:

```c
        #if defined(__ANDROID__)
            #include <android/log.h>
            #define SAUDIO_LOG_DEFAULT(s) __android_log_write(ANDROID_LOG_INFO, "SOKOL_AUDIO", s)
        #elif defined(_WIN32)
            #define SAUDIO_LOG_DEFAULT(s) OutputDebugStringA(msg "\n");
        #else
            #include <stdio.h>
            #define SAUDIO_LOG_DEFAULT(s) puts(s)
        #endif
```

Thoughts? Also if you don't have the time to put those changes into the PR, I can do this during the merge.

Manuzor:
Yeah, I think you can just replace *_LOG_DEFAULT with the old SOKOL_LOG and it should work. So if people define SOKOL_LOG as empty, and don't provide a callback, no logging will happen. Not sure if you still want to provide a compile-time switch to turn all logging into complete no-ops. Might not be worth it at this point.

The Windows case is slightly tricky because it requires Windows.h to be included or a proper forward decl of OutputDebugStringA. But I think that would be really helpful. Although it would make sense to also log to stdout on Windows. Ideally we would log to stdout for console applications and use OutputDebugStringA for Windows applications. Not sure if there's an easy way to determine that.

And about the implementation you suggested: I think you'll have to call OutputDebugStringA twice (once for the newline) since the passed message is not (always) a string literal.

I'm not on a PC right now because I'm a bit sick so I won't be able to work on any of this properly until at least the weekend I expect. Sorry!

floooh:
> And about the implementation you suggested: I think you'll have to call OutputDebugStringA twice (once for the newline) since the passed message is not (always) a string literal.

Ah right, my bad :)

I'll work on this PR first: https://github.com/floooh/sokol/pull/707 and then I can look into the required SOKOL_LOG changes, no worries.

floooh:
Ok, I'll start working on this now.

floooh:
I have simplified the macro setup a bit:
- replace the ```*_DEFAULT_LOG()``` macro with ```SOKOL_LOG()``` (to make the macros backward compatible)
- drop the 'internal' NO_LOG macros, instead check SOKOL_DEBUG directly

I decided against adding a Windows ```OutputDebugStringA()``` ifdef for now... I'm currently thinking that with the next change to replace the simple string-pointer logging callback with a more 'structured' callback, I will also add an easy to plug sokol_log.h header which handles logging for Android and Windows, while the minimal 'fallback logging' will only use puts.

I'll do a bit more testing and then merge to master later today or tomorrow.

floooh:
Ok merged, thanks for the PR!

PR #745, opened 2022-11-15
sokol_gfx.h: add support for half-float vertex attributes

darkuranium:
I had a need for half-float vertex attributes to save on memory, and figured this would be a useful addition upstream. Feel free to ignore this pull request if you disagree.

From my own research, it should be available on *nearly* all GLES2 devices (94.5% coverage according to http://opengles.gpuinfo.org/listextensions.php), and all WebGL2/GLES3 devices.

Note: A useful addition would be a `uint16_t sg_half_from_float(float)` / `float sg_float_from_half(uint16_t)` utility function pair, to help with float16&hArr;float32 conversion.
It is not part of this pull request (mainly since it's not *strictly speaking* part of sokol_gfx's core purpose), but I do have code available, if needed.

The `SG_VERTEXFORMAT_HALF{2,4}` enums are positioned at the end of the existing enum list, for integer-backwards-compatibility reasons. Feel free to reshuffle these if desired.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -638,9 +638,11 @@
         - SG_VERTEXFORMAT_SHORT2
         - SG_VERTEXFORMAT_SHORT4
 
-    - WebGL/GLES2 cannot use integer vertex shader inputs (int or ivecn)
+    - WebGL/GLES2 cannot use integer vertex shader inputs (int or ivecn) or the following:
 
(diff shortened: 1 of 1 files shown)


floooh:
Checking this out now. I think the helper functions to convert between float16 and float32 are out of the scope of sokol-gfx though. If there ever would be a sokol_math.h header, it would better fit in there ;)

PR #747, opened 2022-11-19
sokol_audio.h: check more OpenSLES calls during Android audio init

dfandrich:
This likely caused an application crash on a system with a buggy audio
system (see https://github.com/mgerdes/Open-Golf/issues/27).
Note that this change is completely untested!


floooh:
Looking good, I'll give it a whirl on my Pixel 4 and then merge.

floooh:
Ok , merged. Many thanks! I had to change the SOKOL_LOG() calls to the newer SAUDIO_LOG() (I think this changed in the meantime when a log callback was added).

dfandrich:
I couldn't verify that this solved the problem, since the issue was intermittent and had gone away by the time I came up with the patch. Thanks for merging!

PR #753, opened 2022-11-26
Typo in docs on pipeline stencil

RandyGaul:
Noticed a small typo here, updated the docs for you based on what the default enums are


floooh:
Thanks!

Issue #562, opened 2021-09-20
[Feature request]: sRGB support in sokol_gfx.h


floooh:
sokol_gfx.h should expose the hardware sRGB features to perform sRGB-to-linear conversion when sampling textures and back to sRGB in framebuffers.

See: https://twitter.com/tumult/status/1439897518297411591

...probably needs to be behind feature-flags because of WebGL/GLES2.

OkanArikan:
Are you planning to maintain WebGL/GLES2 support? With the latest Safari, WebGL2 seems widely available now.

floooh:
PS: some more digging regarding SRGB framebuffers:

https://github.com/floooh/sokol/pull/758#issuecomment-1405377648

floooh:
Fixed via https://github.com/floooh/sokol/issues/1520 (only WebGL2 doesn't allow to create an SRGB framebuffer).

PR #758, opened 2022-12-13
basic sRGB sampling with SG_PIXELFORMAT_SRGB8A8 format

allcreater:
SG_PIXELFORMAT_SRGB8A8 texture format is introduced

This is the first separate part towards full sRGB support, as specified in #562.

The SRGB8A8 format is a non-linear RGB format with linear alpha. This format is the most universal and the most supported sRGB format, with the exception of the compressed DXT family. DXT family is out of scope of this pull request but could be implemented separately. The proposed implementation is supposed to work on all backends except WebGL1 and WGPU

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -262,9 +262,9 @@
             where sg_pixel_format is expected). Possible values are:
 
                 23 == SG_PIXELFORMAT_RGBA8
-                27 == SG_PIXELFORMAT_BGRA8
-                41 == SG_PIXELFORMAT_DEPTH
-                42 == SG_PIXELFORMAT_DEPTH_STENCIL
+                28 == SG_PIXELFORMAT_BGRA8
+                42 == SG_PIXELFORMAT_DEPTH
+                43 == SG_PIXELFORMAT_DEPTH_STENCIL
 
         int sapp_sample_count(void)
             Return the MSAA sample count of the default framebuffer.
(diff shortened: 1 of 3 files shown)


floooh:
Thanks for the PR! I'm currently on vacation and do some emulator coding (only do some drive by PR merges here and there), but this one I want to test properly :) I'll try to take care of this when I'm back from vacation (January). If nothing has happened mid January though, please 'at' me :)

allcreater:
@floooh so sorry to bother you, but I wanted to remind you of the request, looks like mid-January is already over :)

Also during this time, [another](https://github.com/floooh/sokol/pull/766) of my requests appeared, it would be great if you can take a look at it too :)

floooh:
Yep, delayed but not forgotten ;)

floooh:
Going to look into this now...

floooh:
Hmm, interesting. I've added the SRGBA8 pixel format to the pixelformats-sapp sample, and with the GL backend I can see a clear difference, while with the Metal backend, RGBA8 and SRGBA8 look absolutely identical 🤔

GL:
Screenshot 2023-01-25 at 19 18 21

Metal:
Screenshot 2023-01-25 at 19 19 05

...I'll check the other platforms too...

floooh:
Ok, I can get the same behaviour in GL as in D3D11 and Metal by enabling GL_FRAMEBUFFER_SRGB for offscreen rendertargets, but disabling it for the default framebuffer (I guess that NSOpenGLView creates an SRGB frambuffer, while the D3D11 and Metal probably create a non-SRGB framebuffer).

I'll enforce that behaviour for in the sokol-gfx GL backend, to properly fix this, an update in sokol_app.h would be needed later (basically an init param for sokol_app.h whether the default framebuffer should be in SRGB format or not).

I'll add a fix to the sokol-gfx GL backend in my merge branch, test on Linux and Windows, and then merge.

floooh:
Ok merged! Many thanks for the PR!

PR #760, opened 2022-12-13
Correct order of mouse/tap events to ImGui for correct touch event handling

Xadiant:
The mouse worked fine, because mouse movement events were updated along the way, but touch events do not do that, so imgui was processing button pushes at the previous tap's location resulting in several issues including getting stuck int he dragging window state like this:

https://user-images.githubusercontent.com/55607276/207410919-c7d61bb6-7760-4d10-9fc1-571d8350ef32.mp4

This PR fixes that issue, and likely fixes issues with selecting buttons rapidly with a mouse (for example a macro that teleports the mouse and doesnt generate move events... I don't know if that is likely to happen or not).

### DIFF
--- a/util/sokol_imgui.h
+++ b/util/sokol_imgui.h
@@ -2344,13 +2344,13 @@ SOKOL_API_IMPL bool simgui_handle_event(const sapp_event* ev) {
             _simgui_add_focus_event(io, false);
             break;
         case SAPP_EVENTTYPE_MOUSE_DOWN:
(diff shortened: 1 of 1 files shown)


floooh:
Many thanks! I wasn't aware that the calling order matters so much with the new evented input API. The PR looks simple enough, I'll just merge it 'blindly', and then update the samples webpage.

floooh:
Ok, the samples webpage has been updated, and the UI works great on my Android phone!

PR #761, opened 2022-12-14
sokol_imgui.h: Fix obsolete key modifier enum usages

legends2k:
ImGui 1.89 marked some key modifiers as obsolete; replace
them with newer ones.


floooh:
Thanks for the PR! It looks like those identifiers are new in 1.89, I'm still using 1.88 everywhere though, so I'll first need to update some places to 1.89 before the merge (I'm currently back in emulator land, but I'll try to do it in the next couple of days, I'll need to touch the UI anyway when working on the emulators, so that's a good opportunity for updating).

floooh:
Ok, merged, and the following wrapper repos have been updated to 1.89.2:

- https://github.com/fips-libs/fips-imgui
- https://github.com/fips-libs/fips-imgui-dock
- https://github.com/fips-libs/fips-cimgui

PR #765, opened 2022-12-23
Add missing XdndFinished notification on XdndDrop

tomas:
Currently this is causing other apps like Firefox to behave erratically, because they expect to receive the XdndFinished when they originate a DND event.

A quick way of seeing the problem is launching a Sokol app with DND enabled and drag a link from Firefox. Since Sokol doesn't reply with a 'XdndFinished' message, Firefox will not allow dragging anything else from the window, because it expects to receive that reply first.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -11507,6 +11507,18 @@ _SOKOL_PRIVATE void _sapp_x11_process_event(XEvent* event) {
                                       _sapp.x11.xdnd.XdndSelection,
                                       _sapp.x11.window,
                                       time);
+
+                    // notify originator source that the dnd operation has finished.
+                    Window source = XGetSelectionOwner(_sapp.x11.display, _sapp.x11.xdnd.XdndSelection);
+                    XEvent reply;
+                    _sapp_clear(&reply, sizeof(reply));
+                    reply.type = ClientMessage;
+                    reply.xclient.window = source;
+                    reply.xclient.message_type = _sapp.x11.xdnd.XdndFinished;
+                    reply.xclient.format = 32;
+                    reply.xclient.data.l[0] = (long)_sapp.x11.window;
+                    XSendEvent(_sapp.x11.display, event->xclient.data.l[0], 0, 0, &reply);
+                    XFlush(_sapp.x11.display);
                 }
                 else if (_sapp.x11.xdnd.version >= 2) {
                     XEvent reply;


floooh:
Thanks for the PR! There's a CI error (just a missing signed/unsigned cast - I have enabled very picky warnings):

```
/home/runner/work/sokol/sokol/tests/compile/../../sokol_app.h:11520:72: error: conversion to ‘Window’ {aka ‘long unsigned int’} from ‘long int’ may change the sign of the result [-Werror=sign-conversion]
11520 |                     XSendEvent(_sapp.x11.display, event->xclient.data.l[0], 0, 0, &reply);
      |                                                   ~~~~~~~~~~~~~~~~~~~~~^~~
```

I'll try to test and merge the PR sometime next week.

tomas:
Great, thanks!

tomas:
By the way, I should mention that I also added support to dropping URLs onto Sokol on my own local copy. I'm not sure the test on your end will work without that.

Would you be interested in a PR containing that change? I haven't added support for other OS's though.

floooh:
I'm going to look into the PR now.

I missed your last reply, apologies for not responding. If the URL support is a somewhat small change I'm happy to merge another PR.

floooh:
Hmm, I'm running into a problem with the change and file dropping. When dropping files, only the first drop works, and after that the ```_sapp_x11_get_window_property()``` call here returns 0, which causes the drop operation to be ignored:

https://github.com/floooh/sokol/blob/c635518ce0ed957a118ce4fc3afccfbfcfda0466/sokol_app.h#L11551-L11557

Sometimes it seems to recover but I haven't identified a pattern yet.

With your change removed, it always works. Any ideas?

I'm testing on Ubuntu 22.04.1 LTS

floooh:
Eh, I think I found the problem. When sending the XdndFinished events, the ```reply.xclient.window``` is set to the wrong window. I think this needs to be set to the source window, not the application window.

E.g. these two lines needs to be changed to ```reply.xclient.window = _sapp.x11.xdnd.source```:

https://github.com/floooh/sokol/blob/c635518ce0ed957a118ce4fc3afccfbfcfda0466/sokol_app.h#L11515

https://github.com/floooh/sokol/blob/c635518ce0ed957a118ce4fc3afccfbfcfda0466/sokol_app.h#L11570

When I hack the accepted MIME type to ```text/x-moz-url```, Firefox seems to behave properly (e.g. dragging the URL from the navbar still works after the first attempt).

Can you check if this fixes your problem without the separate XdndFinished message?

I'm going to commit this as a to master in a sec.

floooh:
I'm closing the PR for now to better keep track of stuff I took care of, but feel free to comment on the closed PR or open a new PR.

tomas:
Great find! I'll test and see whether that fix resolves the issue when dragging URLs. 

Are you planning to include the "hack" to support the `text/x-moz-url` mimetype?

floooh:
> Are you planning to include the "hack" to support the text/x-moz-url mimetype?

No, because the public drag/drop API would need to be extended (the functions ```sapp_get_num_dropped_files()``` and ```sapp_get_dropped_file_path()``` should probably be renamed to ```sapp_get_num_dropped_items()``` and ```sapp_get_dropped_item()```, and there should be a new type ```sapp_dropped_item``` which not only contains the 'path', but also a mime type so that the caller can check *what* has actually been dropped.

tomas:
I see. For what it's worth, I'm constantly dragging URLs from one window to another so I believe it's a pretty common use case. :)

PR #766, opened 2022-12-23
override compile-time limits

allcreater:
Complex graphics pipelines require a greater number of resource bindings than Sokol provides by default. This request introduces the possibility to override certain compile-time constraints.

It's also possible to query sg_limits for the actual values.

It's time to unleash modern GPU capabilities :)


floooh:
Superseded by https://github.com/floooh/sokol/pull/1330

PR #769, opened 2023-01-02
gen_zig.py: update function pointer syntax

kcbanner:
Now that zig stage1 no longer exists, `meta.FnPtr` was removed since there is only one kind of function pointer syntax now.

### DIFF
--- a/bindgen/gen_zig.py
+++ b/bindgen/gen_zig.py
@@ -328,7 +328,7 @@ def gen_struct(decl, prefix):
         elif is_const_prim_ptr(field_type):
             l(f"    {field_name}: ?[*]const {as_zig_prim_type(util.extract_ptr_type(field_type))} = null,")
         elif util.is_func_ptr(field_type):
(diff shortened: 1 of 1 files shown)


floooh:
Ah right, thanks for the update! I'll merge the PR, and the resulting merge-commit should then trigger the bindings generation.

floooh:
PS: I guess the line ```const meta = @import("std").meta;``` can be removed too now. I'll do that in a separate commit, one sec.

kcbanner:
Oops, yep that can be removed now too 👍

floooh:
PPS: I guess this switch can be removed too (but the compileError needs to remain), but I'm fine with keeping it in until it starts to break (unless you want to provide another PR, I'm currently still focused on emulator stuff and don't have much time for testing)

https://github.com/floooh/sokol-zig/blob/4ffa880ae4285df38d5abc62c437094416b313e5/src/sokol/gfx.zig#L21-L25

kcbanner:
I'll send a PR for that later tonight 👍

PR #776, opened 2023-01-19
Align validation with greatest common requirement

nmr8acme:
While the Metal backend will accept a larger buffer neither the GL nor the DirectX backend will. By making validation stricter we prevent code developed with the Metal backend from crashing when run on GL or DX.

See e.g. validation for GL: https://github.com/floooh/sokol/blob/b3aecb21825f3191a9a150bc0918fc8c7df7855c/sokol_gfx.h#L7985

Also, it may only crash with validation enabled.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -11702,7 +11702,7 @@ _SOKOL_PRIVATE void _sg_mtl_apply_uniforms(sg_shader_stage stage_index, int ub_i
     SOKOL_ASSERT(_sg.mtl.state_cache.cur_pipeline->slot.id == _sg.mtl.state_cache.cur_pipeline_id.id);
     SOKOL_ASSERT(_sg.mtl.state_cache.cur_pipeline->shader->slot.id == _sg.mtl.state_cache.cur_pipeline->cmn.shader_id.id);
     SOKOL_ASSERT(ub_index < _sg.mtl.state_cache.cur_pipeline->shader->cmn.stage[stage_index].num_uniform_blocks);
-    SOKOL_ASSERT(data->size <= _sg.mtl.state_cache.cur_pipeline->shader->cmn.stage[stage_index].uniform_blocks[ub_index].size);
+    SOKOL_ASSERT(data->size == _sg.mtl.state_cache.cur_pipeline->shader->cmn.stage[stage_index].uniform_blocks[ub_index].size);
 
     /* copy to global uniform buffer, record offset into cmd encoder, and advance offset */
     uint8_t* dst = &_sg.mtl.cur_ub_base_ptr[_sg.mtl.cur_ub_offset];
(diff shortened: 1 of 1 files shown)


floooh:
Hmmm..... I think this should be solved differently, already in the validation function for sg_make_shader(), the uniform block sizes provided there should be checked against a common max size (not sure at the moment what this size is, either 16 or 64 KBytes I'd guess).

The other option would be to make the max uniform block size a runtime variables exposed via sg_query_features(), this wouldn't help much for writing portable code, but would allow code to take advantage of bigger limits in some 3D backends.

nmr8acme:
Maybe I am misunderstanding, or maybe my fix is wrong, but it sounds like you're talking about a separate issue.

The issue I identified and tried to fix doesn't relate to max UB size, but rather validating that the sg_range argument passed to sg_apply_uniforms is properly validated size wise. The GL backend validates that the range is == the expected size, but the Metal backend, as well as _sg_validate_apply_uniforms, accept ranges >= the expected size. This leads too-large buffers passed to sg_apply_uniforms to succeed on Metal, but fail validation on GL. See what I mean?

floooh:
Doh... github just threw away my carefully researched reply :D 

TL;DR: yes, it was a misunderstanding and your PR makes sense :)

GL and D3D11 both have an assert in their apply-uniforms backend function which checks for the exact size:

- https://github.com/floooh/sokol/blob/b3aecb21825f3191a9a150bc0918fc8c7df7855c/sokol_gfx.h#L7985
- https://github.com/floooh/sokol/blob/b3aecb21825f3191a9a150bc0918fc8c7df7855c/sokol_gfx.h#L9938

...but for reasons lost in history, in the Metal backend this is a ```<=```: https://github.com/floooh/sokol/blob/b3aecb21825f3191a9a150bc0918fc8c7df7855c/sokol_gfx.h#L11705

...I checked via git blame, and it's always been like that, so it's most likely just an oversight (although I'm not 100% sure if I had some problems with padding bytes at the end of uniform block structs in Metal which then resulted in some sort of size mismatch, but in any case that should no longer be an issue.

One request, while you're at it, can you maybe also fix the assert in ```_sg_mtl_apply_uniforms()``` from ```<=``` to ```==``` (here: https://github.com/floooh/sokol/blob/b3aecb21825f3191a9a150bc0918fc8c7df7855c/sokol_gfx.h#L11705).

Apart from that, the ticket I wrote still makes sense, even if it's an entirely different topic, heh.

nmr8acme:
RE lost reply: Ugh, I hate it when that happens.

RE validation consistency in _sg_mtl_apply_uniforms: Yep, will do

floooh:
I'll try to test and merge in the next few days. Just wrapping up the sokol_audio.h update today.

floooh:
Ok, merged. Many thanks! I did some randomly picked tests in the sokol-samples, my emulators, and v6502r, and they all work fine.

nmr8acme:
Many thanks to you, I would have been stuck on OpenGL ES 2 for a lot longer without sokol_gfx.

PR #787, opened 2023-02-10
OpenGL: two stage texture binding bug

allcreater:
There was a specific problem with the OpenGL backend: despite the API providing support for at least 12(`SG_MAX_SHADERSTAGE_IMAGES`) slots for textures for the vertex shader and 12 for the fragment shader, the OpenGL implementation could bind only SG_MAX_SHADERSTAGE_IMAGES combined for both stages because of the length of the binding cache.

I propose fixing this issue by increasing the cache size to the value of `SG_MAX_SHADERSTAGE_IMAGES * SG_NUM_SHADER_STAGES` so that all images can be bound. The actual number of supported images is known at runtime, `_sg.gl.max_combined_texture_image_units`, and is already used to prevent unsupported bindings.

PS: the previous PR #766 could be used to work around these restrictions as well, but that is an independent improvement with its own goal to provide the possibility to increase limits itself (there are shaders where a total of 24 textures is not enough).

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -3920,6 +3920,10 @@ typedef struct {
     GLuint texture;
 } _sg_gl_texture_bind_slot;
 
+enum {
+    _SG_GL_IMAGE_CACHE_SIZE = SG_MAX_SHADERSTAGE_IMAGES * SG_NUM_SHADER_STAGES,
+};
+
 typedef struct {
     sg_depth_state depth;
     sg_stencil_state stencil;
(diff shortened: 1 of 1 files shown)


floooh:
Yep, the PR makes complete sense. I'll just change the enum to a define (since it's not a public constant and just in the implementation that's fine - such constants only need to be an enum if they are part of the language bindings).

floooh:
Ok merged. Thanks for the PR! I'll followup with a commit where I'll expose GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS in the sg_limits struct.

PR #788, opened 2023-02-11
Add new sokol_poll.h layer

Toctave:
Basic polling layer on top of sokol_app's event API. Only supports keyboard input for now.

### DIFF
--- a/sokol_poll.h
+++ b/sokol_poll.h
@@ -132,4 +132,8 @@ bool spoll_keydown(sapp_keycode key)
     return g_key_down[BIT_INDEX(key)] & BIT_MASK(key);
 }
 
+/* Macro cleanup */
+#undef BIT_INDEX
+#undef BIT_MASK
+
 #endif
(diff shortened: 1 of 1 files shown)


floooh:
It's a good PR for discussion, but I'll probably won't merge it, sorry :)

I have a couple more things in mind apart from just managing a key-pressed bitfield:

- have an "any" and "all" function which take bitmasks to check of any of a subset or all of a subset of keys is currently pressed
- then there's the whole problem area of missing presses when the down- and up-event happen in the same frame, so "is key X currently pressed" becomes a "was key X pressed at any time during the last frame" => which in turn means that the library needs to know when a new frame starts... for a solution there should probably a separate bitmask for key-down events and key-up events in a frame, which are then 'consolidated' into a key-pressed bitmask.
- and the 'key mask' should probably be a 'button mask' which at least includes mouse buttons
- and it should also be possible to poll the current mouse position
- ...and while at it, it would be nice to tackle gamepad support, but not sure if this should go into a separate header - but the input polling header should at least somehow integrate the gamepad input.

...etc... etc...

Toctave:
Yeah I thought of this as a starting point / discussion PR, ofc it won't be merged until it fits what you had in mind.

PR #792, opened 2023-02-16
sokol_audio.h: Include alloca.h on Linux

darkerbit:
Required by a `snd_pcm_hw_params_alloca` call in `_saudio_alsa_backend_init`.

This change makes it so the implementation links correctly on Linux in `-std=c99` mode, instead of relying on implicit `alloca` only provided by GNU extensions.


floooh:
...hmm ALSA is really full of surprises :/ Thanks for the PR!

PR #797, opened 2023-02-23
fixed mouse position on event SAPP_EVENTTYPE_FILES_DROPPED (MacOS)

allcreater:
Hi, it looks like there is a small bug at sokol_app under MacOS: when the files drop occurred, the `SAPP_EVENTTYPE_FILES_DROPPED` listener got the incorrect mouse position. 

I think it can be fixed somehow like this :)

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3968,6 +3968,12 @@ _SOKOL_PRIVATE void _sapp_macos_frame(void) {
         }
         if (!drop_failed) {
             if (_sapp_events_enabled()) {
+                NSPoint mouse_pos = [sender draggingLocation];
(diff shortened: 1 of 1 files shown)


floooh:
Thanks for the PR! I actually need to check if any of the other backend even try to set mouse position to the correct drop location or if they are just somewhat right by accident (because the mouse location is tracked during the drag-n-drop).

IIRC it wasn't even intended that the mouse position in the drop event is even meaningful, just that anything has been dropped anywhere in the window.

I'll try to look into this "soon-ish".

floooh:
Hmm, not sure why I forgot to close this PR, but I had fixed the issue separately in May 2023:

https://github.com/floooh/sokol/blob/master/CHANGELOG.md#20-may-2023

Issue #800, opened 2023-03-06
Optionally capturing keyboard input in web app


SanderMertens:
Hey! I just tried to embed a canvas inside a web app that has a few other widgets that need keyboard input, but noticed that all keyboard inputs are intercepted by the canvas, which I think is due to these lines:

```js
    emscripten_set_keydown_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW, 0, true, _sapp_emsc_key_cb);
    emscripten_set_keyup_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW, 0, true, _sapp_emsc_key_cb);
    emscripten_set_keypress_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW, 0, true, _sapp_emsc_key_cb);
```

Is it possible to add an option that conditionally adds/removes these listeners on canvas focus/blur?

floooh:
Can you test whether replacing those with ```_sapp.html5_canvas_selector``` does the right thing for you?

I don't remember why I treated key events special here, maybe because otherwise the arrow keys would scroll the page.

If you find a solution that works for you we can add a new config option to ```sapp_desc``` which controls the behaviour.

SanderMertens:
I tried changing it to `_sapp.html5_canvas_selector`, but then the canvas doesn't receive keyboard events, not even when focused.

floooh:
The PR I just merged should allow full control over bubbling individual events now:

https://github.com/floooh/sokol/pull/975

Also see the changelog entry: 

https://github.com/floooh/sokol/blob/master/CHANGELOG.md#27-jan-2024

PR #801, opened 2023-03-06
app: add JS function to enable/disable capturing canvas keyboard events

SanderMertens:
Possible solution for https://github.com/floooh/sokol/issues/800

After loading the module you can call this function to enable/disable keyboard event capture:
```js
Module.sokol_capture_keyboard_events(false);
```

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -4859,6 +4859,12 @@ EM_JS(void, sapp_js_remove_dragndrop_listeners, (const char* canvas_name_cstr), 
     canvas.removeEventListener('drop',      Module.sokol_drop);
 });
 
+EM_JS(void, sapp_js_add_capture_keyboard_events, (void), {
+    Module.sokol_capture_keyboard_events = function(capture) {
+        __sapp_emsc_capture_keyboard_events(capture);
+    };
+});
+
 /* called from the emscripten event handler to update the keyboard visibility
     state, this must happen from an JS input event handler, otherwise
     the request will be ignored by the browser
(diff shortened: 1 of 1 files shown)


SanderMertens:
Here is an example of the solution working in practice:
https://www.flecs.dev/explorer/?wasm=https://www.flecs.dev/explorer/playground.js

When you click on the canvas I toggle call `Module.sokol_capture_keyboard_events` with `true` or `false`.

floooh:
Hmm... not a fan of this solution TBH :D 

I need to tinker around a bit with having a smaller sokol-canvas along side with other HTML elements on a scrollable page... there must be a solution that doesn't require hooking and unhooking event handlers (I hope).

floooh:
...that reminds me: sokol_app.h has actually the ability to consume keyboard events via ```sapp_consume_event()```, which muyst be called from inside the event handler callback. I had added this specifically for the browser platform when working on https://floooh.github.io/visual6502remix/ to capture special key combinations which are intepreted by the browser (like Ctrl+S or Ctrl+O)

I wonder if this would also work for your use case?

SanderMertens:
Is the goal of `sapp_consume_event` to stop an event from propagating? I think that could be useful yeah, but doesn't solve the issue of the PR which is kind of the reverse (prevent the canvas from consuming keyboard events that were meant for the UI).

The problem I'm running into is more like that of a game editor: in play mode WASD should be used to move the player around, but obviously not when I'm in editor mode.

floooh:
Hmm ok, yeah, consume event does the opposite of what you want. But if every event would bubble up by default you would simply ignore specific events in your WASM event handling code, and they would always arrive up on the HTML page.

But I guess I know what the problem is (just stumbled over this again when starting to work on mobile character input).

There's this hardwired table of keyboard events that never bubble up:

https://github.com/floooh/sokol/blob/7373dda48c5e62953cc7ec942da90fc6a25989fa/sokol_app.h#L5397-L5459

...meaning that currently specific keys, like the arrow keys, are always 'consumed', no matter if ```sapp_consume_event()``` is called or not.

Obviously, this hardwired table is bad, I'll try to tinker around with this a bit. Since 'vanilla' Sokol Emscripten apps are usually not embedded into another page (but instead cover the whole client area), this hardwired table may not be necessary anymore, and even if it is still needed, there should be a more flexible way to set up such 'auto-consumed' events.

floooh:
...just removing the switch-case has a couple of unexpected issues, even with the WebGL canvas covering the entire window client rectangle, for instance pressing the Tab key will move the focus to the browser navigation bar (which I guess is sometimes desired, other times not).

Maybe a ```sapp_desc.html5_disable_auto_consume_events``` is enough for your use case? (you would then set this to true, which means all events are bubbled up to the HTML page)?

SanderMertens:
Hm, I don't think that solving this in native code would work for me. For example, in this app you can click on the canvas to give it focus and move the camera around with WASD, _as if_ you were running it standalone: https://www.flecs.dev/explorer/?wasm=https://flecs-hub.github.io/playground/etc/playground.js

That same wasm image can also be ran standalone as a "vanilla" sokol emscripten app, where it obviously should capture keyboard events: https://flecs-hub.github.io/playground/etc/

If this setting would be hardcoded in the wasm image, you couldn't run the same image standalone & in an editor context. It also wouldn't allow you (I think) to switch between modes where the canvas is accepting/not accepting events.

floooh:
The changes in https://github.com/floooh/sokol/pull/975 should allow full control over HTML event bubbling now (also see: https://github.com/floooh/sokol/blob/master/CHANGELOG.md#27-jan-2024)

PR #803, opened 2023-03-12
Fix warning relating to glTexParameterfv

pr0g:
Fix warning C4113: 'PFN_glTexParameterfv' differs in parameter lists from 'PFNGLTEXPARAMETERFVPROC'

See https://registry.khronos.org/OpenGL-Refpages/gl4/html/glTexParameter.xhtml for reference (third parameter should be `const` qualified).


floooh:
Oh, interesting, I wonder why that warning slipped through on my side (because the Windows CI builds are compiled with ```/W4 /WX``` but C4113 seems to be /W1: https://learn.microsoft.com/he-il/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4113?view=msvc-160

floooh:
Merged. Many thanks!

pr0g:
> Merged. Many thanks!

No problem at all! 🙂 Thanks for merging 🥳

Yes I believe it showed up as I'm using Glad as my OpenGL loader and this showed up by chance as I'd forgotten to define `SOKOL_EXTERNAL_GL_LOADER` 🤦‍♂️

PR #814, opened 2023-04-02
gh actions: replace unmaintained actions-rs/toolchain action

striezel:
Basically all of the `actions-rs/*` actions are unmaintained. See <https://github.com/actions-rs/toolchain/issues/216> for more information. Due to their age they generate several warnings in CI runs, for example in https://github.com/floooh/sokol/actions/runs/4583912323:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions-rs/toolchain@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

To get rid of those warnings the occurrences of `actions-rs/toolchain` are replaced by [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain).

### DIFF
--- a/.github/workflows/gen_bindings.yml
+++ b/.github/workflows/gen_bindings.yml
@@ -261,11 +261,9 @@ jobs:
         with:
           name: ignore-me-rust
           path: src/sokol
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           toolchain: stable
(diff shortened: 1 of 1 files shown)


floooh:
Thanks! I'll also apply the same fix to the sokol-rust repo.

striezel:
Yes, that is a good idea. :)

PR #815, opened 2023-04-03
_sg_mtl_begin_pass: explicit store action for depth and stencil attachments (MTLStoreActionStore)

allcreater:
Sometimes two passes with different color attachments must share a depth or/and stencil buffers. 

The default Metal action is "don't care" so that the buffers could be invalidated (and it happens on some iOS versions)

I suggest setting the default action to "store" to provide support of such passes.


floooh:
Yeah, you're right, change of plans :)

I would prefer however, if only offscreen render passes would need to store the depth/stencil buffer result. (e.g. only the first half of your

allcreater:
@floooh I guess the first half is perfectly fine, especially if two calls of default pass is undefined behavior :)

Thank you for your attention :)

floooh:
I have committed your fix for offscreen render targets only here: https://github.com/floooh/sokol/commit/e56c596810d8be80772314d1a99378ae0b998e6a

As for the 'proper cleanup', I'll try to tackle this before coming back to finish the WebGPU backend (that's the next big thing on my list because it looks like WebGPu is now actually close to release).

floooh:
(ah sorry, I wasn't aware that you also fixed the PR)

allcreater:
@floooh It's OK, you already mentioned me a couple of times, and also I'll be glad to do anything else later :)

PR #817, opened 2023-04-07
sokol_app.h macos: quit menu

medvednikov:
Right now there's no way to quit a Sokol application on macOS.

This adds a quit menu and a cmd+q shortcut.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3765,6 +3765,26 @@ _SOKOL_PRIVATE void _sapp_macos_frame(void) {
     _sapp.macos.window.acceptsMouseMovedEvents = YES;
     _sapp.macos.window.restorable = YES;
 
+     // Quit menu
+     NSMenu* menu_bar = [[NSMenu alloc] init];
+     NSMenuItem* app_menu_item = [[NSMenuItem alloc] init];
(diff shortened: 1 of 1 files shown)


floooh:
Hi, there's a couple of similar PRs and issues already, the main reason why I haven't merged any of them yet is that they all suffer from one or another problem (for instance the 'Quit' being hardcoded, so it won't be properly localized.

I would prefer a solution which allows the menu to be configured from the outside.

Until then it's probably better to do the menu setup outside of sokol_app.h

medvednikov:
Right now there's no way to quit a Sokol app on macos, so I think it's a good temporary fix, that takes care of this pretty big limitation.

Later, when the menu situation is figured out, it can be replaced/updated.

floooh:
> Right now there's no way to quit a Sokol app on macos

You can quit the application programmatically, either "hard" or "soft" (where the application has a chance to show a "Really Quit?" dialog). See here:

https://github.com/floooh/sokol/blob/7373dda48c5e62953cc7ec942da90fc6a25989fa/sokol_app.h#L733-L807

medvednikov:
Yeah, I have a full screen app, so not an option there. There's no window frame.

PR #820, opened 2023-04-23
Fix character input on HTML5 mobile.

floooh:
Android/iOS differences:

- imgui-sapp on Android works completely as expected when tapping the multi-line text input field
- imgui-sapp on iOS requires a double click to show and hide the keyboard (e.g. the request to open or hide the keyboard only happens in the **next** handled event)

Remaining problems:

- When calling ```sapp_show_keyboard(true)``` in chips emulators inside TOUCHES_BEGAN input handler, iOS Safari zooms the canvas in.
- Same situation, on iOS Safari, the keyboard is closed on each tap on the keyboard.
- ...those things do not happen in imgui-sapp though when double-tapping the multiline text input field.
- ...on Android, when calling ```sapp_show_keyboard(true)``` inside TOUCHES_BEGAN, keyboard shows up then immediately disappears in the next frame.
- ...on Android, calling ```sapp_show_keyboard(true)``` in TOUCHES_ENDED, keyboard shows up but then canvas zooms in (similar to Safari), and keyboard disappears after each key press (also similar to Android)

...why do the emulators and the imgui-sample behave so differently? Only difference seems to be that ```sapp_show_keyboard``` is called outside the event handler when using sokol_imgui.h??

...implementing that same behaviour in the emulators doesn't make a difference though...

...but maybe the same tap which causes the helper input text field to become focused also unfocuses it immediately? But when that's the case why doesn't it happen in the imgui-sapp sample?

...all this mess is probably a good reason to go with the original plan to remove this mobile text input hack completely. It's brittle as f*ck.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -4779,7 +4779,7 @@ EM_JS(void, sapp_js_create_textfield, (void), {
     });
     _sapp_inp.addEventListener("input", (ev) => {
         // put some upper limit on the max text field input length
-        if (ev.target.value.length < SAPP_EMSC_MAX_INPUT_CODE_POINTS) {
+        if (ev.target.value.length < 32) {
             withStackSave(() => {
                 const cstr = allocateUTF8OnStack(ev.target.value);
                 __sapp_emsc_ontextinput(cstr);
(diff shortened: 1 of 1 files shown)


floooh:
Closing as won't fix.

PR #829, opened 2023-05-11
Add support for mouse magnification events on MacOS.

castano:
Also flag trackpad events with SAPP_MODIFIER_TRACKPAD.

The purpose of this PR is to enable applications to support more natural for trackpad interactions, allowing them to pinch to zoom and drag to translate.

I'm hoping I can add support for this on windows through the WM_GESTURE messages. Not sure if that works with trackpads or touch screens only.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -103,6 +103,7 @@
     MOUSE_DOWN          | YES     | YES   | YES   | ---   | ---     |  YES
     MOUSE_UP            | YES     | YES   | YES   | ---   | ---     |  YES
     MOUSE_SCROLL        | YES     | YES   | YES   | ---   | ---     |  YES
+    MOUSE_MAGNIFY       | NO      | YES   | NO    | ---   | ---     |  NO
     MOUSE_MOVE          | YES     | YES   | YES   | ---   | ---     |  YES
     MOUSE_ENTER         | YES     | YES   | YES   | ---   | ---     |  YES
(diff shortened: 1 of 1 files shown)


floooh:
Hmm not sure about this one (to have any gesture recognition events inside sokol_app.h), since gesture recognition is vastly different between the various operating systems.

My (long term) plan was to have a separate header (like sokol_gestures.h) which implements its own gesture recognizers (taken from Android NDK example code) on top of the low level touch events.

I had something like this in Oryol for pan, pinch and 'multi-tap': https://github.com/floooh/oryol/blob/master/code/Modules/Input/private/gestureDetectors.cc

...this would first require to add touch (pad) events to the 'desktop platforms' though.

floooh:
...on the other hand it's a very simple change... but I'll need to think a bit longer about this PR than the other two :)

castano:
I've been playing around with the WM_GESTURE messages, but they only seem to work with touch screens and not with touchpads. Any approach that relies on touch messages will have the same issue, since as far as I know touchpads do not report touch messages. However, some windows apps in my laptop seem to be able to react to scroll and zoom touchpad events! Not sure how they do it. I've tried Spy++ but it does not seem to work on windows 11. I'll report if I can find a way to get this to work on windows like it does on macos.

cschar:
interested in this feature! especially for MacOS trackpad pinch-zoom gestures

PR #830, opened 2023-05-11
Update event modifiers on drag and drop events.

castano:
In order to detect if any modifier keys are being pressed at the end of a drop event.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -7041,6 +7041,7 @@ _SOKOL_PRIVATE void _sapp_win32_files_dropped(HDROP hdrop) {
     if (!drop_failed) {
         if (_sapp_events_enabled()) {
             _sapp_init_event(SAPP_EVENTTYPE_FILES_DROPPED);
+            _sapp.event.modifiers = _sapp_win32_mods();
             _sapp_call_event(&_sapp.event);
         }
     }


floooh:
This is a good change, but I'll need to check if this fix is also needed on macOS and Linux.

castano:
I don't know about Linux, but this certainly needs more work on MacOS. There's no NSEvent object available in the drop event handler, so it wasn't immediately obvious how to update the modifiers. It should not be hard to set these based on the most recent value, though.

floooh:
Starting to look into this now.

macOS seems to have a class-method (aka static method) on NSEvent which provides the current event mask without requiring an actual event object: https://developer.apple.com/documentation/appkit/nsevent/1535211-modifierflags?language=objc

floooh:
I'm doing a few slightly related fixes in my merge-branch, notes to self:

- set dx/dy to zero in `SAPP_EVENTTYPE_MOUSE_ENTER/LEAVE`:
    - [ ] macos
    - [ ] emsc
    - [ ] linux
    - [ ] win32
- set mouse-pos in `SAPP_EVENTTYPES_FILES_DROPPED`:
    - [ ] macos (previously this was the drag-start pos outside the window)
    - [ ] emsc (was already working)
    - [ ] linux
    - [ ] win32
- dx/dy should be zero `SAPP_EVENTTYPES_FILES_DROPPED`:
    - [ ] macos
    - [ ] emsc (was already working)
    - [ ] linux
    - [ ] win32
- set modifier keys in `SAPP_EVENTTYPE_FILES_DROPPED`:
    - [ ] macos
    - [ ] emsc
    - [ ] linux
    - [ ] win32

floooh:
Ok merged. Many thanks for the PR!

castano:
Awesome, thanks!

PR #831, opened 2023-05-11
Mouse wheel message is setting invalid mouse position.

castano:
The MOUSEWHEEL messages provide the mouse position in absolute coordinates, so updating it the same way you respond to other mouse messages is not correct. One solution would be to translate the mouse coordinates to client space, but it's not clear to me why this is being updated here. 

This problem was introduced recently and my application was working fine before it. I can implement a different workaround if the mouse_update call is necessary for reasons that I'm not aware of. For example, we could pass a flag to transform the coordinates when necessary.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -7242,11 +7242,9 @@ _SOKOL_PRIVATE LRESULT CALLBACK _sapp_win32_wndproc(HWND hWnd, UINT uMsg, WPARAM
                 }
                 break;
             case WM_MOUSEWHEEL:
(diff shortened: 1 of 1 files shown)


castano:
The other messages do not seem to have this problem.

floooh:
Ok merged! Thanks for spotting this issue and providing a PR. I'll try to look at least into https://github.com/floooh/sokol/pull/830 over the next few days (as I said there, need to check the other platforms, if Linux and macOS needs the same fix I'll do it as part of the merge.

PR #835, opened 2023-05-20
Added checks to avoid needless calls to glVertexAttribPointer and Div…

LorenRoosendaal:
I noticed that in some specific cases we were paying a good percentage of extra cost per sg_apply_bindings (dropping raw draw calls processed in a little benchmark by as much as 35%) because it doesn't check if the newly bound buffer actually has different Attributes (in optimized production code they can often be the same). Added a little fix.

Hope I didn't slip up anywhere, this is just a quick commit from the browser ;-) looks clean to me though!

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -7966,18 +7966,27 @@ _SOKOL_PRIVATE void _sg_gl_apply_bindings(
             SOKOL_ASSERT(vb);
             gl_vb = vb->gl.buf[vb->cmn.active_slot];
             vb_offset = vb_offsets[attr->vb_index] + attr->offset;
-            if ((gl_vb != cache_attr->gl_vbuf) ||
-                (attr->size != cache_attr->gl_attr.size) ||
+            if (gl_vb != cache_attr->gl_vbuf) 
+            {
+                _sg_gl_cache_bind_buffer(GL_ARRAY_BUFFER, gl_vb);
+                cache_attr_dirty = true;
+            }
+            //It is entirely possible and even common (in production applications)
+            //that the newly bound buffer has the same attributes as the previous one
+            //we can save some state change and gl call performance here.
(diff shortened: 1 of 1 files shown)


floooh:
Thanks, I think that should work. I'll give it a whirl later today!

floooh:
I'm starting to look into the PR now.

floooh:
Ah ok, I found a flaw with the PR :)

It's not very obvious from the GL spec, but the currently bound vertex buffer isn't independent from the vertex attribute state, but instead the vertex buffer binding is part of the vertex attribute state, otherwise pulling different vertex component from different vertex buffers wouldn't work. The glVertexAttribPointer() call basically 'captures' the current buffer bindings and stores it in the vertex attrib state.

From https://registry.khronos.org/OpenGL-Refpages/gl4/html/glVertexAttribPointer.xhtml:

Screenshot 2023-05-21 at 12 31 22

This means that just changing the buffer bindings doesn't change any existing vertex attribute state (e.g. existing vertex attribute definitions will still refer to the buffer that was bound when at the time glVertexAttribPointer was called last time).

I'm also somewhat sure that in the olden days of sokol_gfx.h, this code actually treated the buffer binding as a separate state (so it looked similar to your PR) but this fell apart when I added support for pulling vertex components from multiple vertex buffers (unfortunately this change must have been before I combined sokol_gfx.h into a single header, so git blame stops at that point).

There's also this 'subtle fix' for the vertex attribute divisor from 5 years ago which moved the glVertexAttribDivisor call from being independent from the glVertexAttribPointer call into the same code block:

https://github.com/floooh/sokol/commit/fa960b02393082ec47a9e2715d7c820053b0a75e

In this case I'm not sure though if the glVertexAttribDivisor call can be put into an additional if-block, need to carefully read the GL spec and wiki again.

In any case, I'll close the PR without merging, sorry about that.

PR #839, opened 2023-06-01
Create GL context with DEBUG_BIT set when SOKOL_DEBUG is defined.

castano:
This mirrors the behavior of D3D11 and allows applications to use OpenGL debug output functionality.

Note that in non-debug contexts, the OpenGL implementation may not generate messages even if debug output is enabled, so this is required for debug output to work in all cases.

Questions:
- Should we check for the presence of KHR_debug before enabling this?
- Should this be an app_desc parameter instead of a compile time definition?


castano:
Any updates about this? It'd be great to merge this to make it easier to maintain my sokol_app branch.

floooh:
Thanks for the reminder. I'll try to give the PR a whirl this evening or over the weekend.

floooh:
Ok, I'm not seeing any problems on my old and cheap Intel GPU laptop, guess that's a good sign ;)

Merging...

castano:
Thanks!

PR #846, opened 2023-06-23
use addr instead unsafeAddr

Angluca:
warning:: `unsafeAddr` is a deprecated alias for `addr`
use `addr` instead.

### DIFF
--- a/.github/workflows/gen_bindings.yml
+++ b/.github/workflows/gen_bindings.yml
@@ -125,7 +125,7 @@ jobs:
     steps:
       - uses: jiro4989/setup-nim-action@v1
         with:
-         nim-version: 1.6.12
+         nim-version: devel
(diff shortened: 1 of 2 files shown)


floooh:
Nice! Will try to take care of this and the PR in sokol-nim over the weekend.

floooh:
Ok, same NIm version compatibility problem  in the CI pipeline like in https://github.com/floooh/sokol-nim/pull/21#issuecomment-1604483680.

Can you try upgrade the nim-version here?

https://github.com/floooh/sokol/blob/177e2c79bee74acebaeeb8d5b29f7f168858ff78/.github/workflows/gen_bindings.yml#L128

floooh:
...hmm, no idea what's going wrong there in the Nim setup...

floooh:
Ok merged, thanks for the PR!

Issue #40, opened 2018-04-01
Image Export


thennequin:
Hi Andre,
I use sokol since few weeks and (for now) I think the only missing feature is the export of image to CPU side.
I don't known if this request is compatible with Metal and GLES API but I think it's Ok for DX11 and OpenGL.
And maybe you don't want to add new feature to your lib?
Anyway, thanks for this awesome little and simple lib.

floooh:
Yes this will be tricky to do across 3D APIs, I think a better option is to allow access to the internal resource structs, so that the 3D objects are accessible.

There's also a few other options as workaround:
- you can inject your own 3D API textures and buffers, so that you have direct access to them in your own code (see the inject samples for instance: https://github.com/floooh/sokol-samples/blob/master/glfw/inject-glfw.c),
- another option is to add your own sokol 'extension functions' outside the header, but inside the implementation source, since in the implementation source file you already have access to all private functions and structures.

thennequin:
Ok thanks for your answer, I think I will do the second option.

thennequin:
First version with OpenGL support only.
https://gist.github.com/thennequin/719ef9c3db55f93ca16e161cc95d71ab

Issue #39, opened 2018-03-26
Visualize depth buffer


dbacchet:
Hi, thanks for creating sokol, I find very refreshing working with it!
Question:
do you have somewhere a snippet of code or an example on how I can access/visualize the depth buffer?
I naively tried to attach the depth_image to a color attachment but it doesn't seem to work for me.

For example, in the `offscreen-glow.c` sample, if I modify the default draw state like this (around line 242):
```c
    sg_draw_state default_ds = {
        .pipeline = default_pip,
        .vertex_buffers[0] = vbuf,
        .index_buffer = ibuf,
        .fs_images[0] = depth_img // was: "color_img"
    };
```
I see no difference in the generated output image.

Similarly, using the `mrt-glfw.c` sample, if I change the debug view in the main loop to also draw the depth, like this (around line345):
```c
        sg_draw(0, 4, 1);
        for (int i = 0; i < 3; i++) {
            sg_apply_viewport(i*100, 0, 100, 100, false);
            dbg_ds.fs_images[0] = offscreen_pass_desc.color_attachments[i].image;
            sg_apply_draw_state(&dbg_ds);
            sg_draw(0, 4, 1);
        }
        for (int i = 3; i < 4; i++) {
            sg_apply_viewport(i*100, 0, 100, 100, false);
            dbg_ds.fs_images[0] = offscreen_pass_desc.depth_stencil_attachment.image;
            sg_apply_draw_state(&dbg_ds);
            sg_draw(0, 4, 1);
        }
        sg_end_pass();
```
what I get is a duplication of the last color attachment that has been bound (see the following screenshot).
<img width="478" alt="screen shot 2018-03-26 at 7 00 37 am" src="https://user-images.githubusercontent.com/10245297/37910927-8fd44320-30c3-11e8-8665-b1a29f7f2eb1.png">

Probably I'm trying to "map" my OpenGL mindset to Sokol, and doing something wrong.
Any suggestion on the right way to use the depth buffer in a shader?

thanks!

floooh:
Using a depth image object as texture won't work unfortunately (and it should warn about this in the validation layer, looks like I forgot a check).

The reason (at for GL) is that a sokol image with depth-pixelformat is created as a glRenderBuffer object (see here: https://github.com/floooh/sokol/blob/ce6414d9eb24d7462fca0365aeb30916b8f33b53/sokol_gfx.h#L2738)

[...]

PR #849, opened 2023-06-30
Update zig bindgen produce source that matches "zig fmt" style

nurpax:
for #39, #40

### DIFF
--- a/bindgen/gen_zig.py
+++ b/bindgen/gen_zig.py
@@ -212,7 +212,7 @@ def as_c_arg_type(arg_type, prefix):
     elif is_const_struct_ptr(arg_type):
         return f"[*c]const {as_zig_struct_type(util.extract_ptr_type(arg_type), prefix)}"
     elif is_prim_ptr(arg_type):
-        return f"[*c] {as_zig_prim_type(util.extract_ptr_type(arg_type))}"
+        return f"[*c]{as_zig_prim_type(util.extract_ptr_type(arg_type))}"
     elif is_const_prim_ptr(arg_type):
         return f"[*c]const {as_zig_prim_type(util.extract_ptr_type(arg_type))}"
     else:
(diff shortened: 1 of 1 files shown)


nurpax:
Tested by running this in the `bindgen` sub-directory:

```
import glob
import subprocess

def main():
    for file in glob.glob('sokol-zig/src/sokol/*.zig'):
        # Open the file and read its contents
        with open(file, 'rt', encoding='utf-8') as f:
            file_contents = f.read()
        p = subprocess.run('zig fmt --stdin', input=file_contents, shell=True, text=True, capture_output=True)
        with open(f'{file}.fmt', 'wt', encoding='utf-8') as f:
            f.write(p.stdout)
        diff_fname = f'{file}.patch'
        p = subprocess.run(["diff", '-u', file, f'{file}.fmt'], text=True, capture_output=True)
        with open(diff_fname, 'wt', encoding='utf-8') as f:
            f.write(p.stdout)
        print('wrote', diff_fname)

if __name__ == '__main__':
    main()
```

This writes `<zig src>.fmt` (zig fmt'd version) and `<zig src>.patch` files in the zig source directory.  Patch is expected to be empty if bindgen output matches what zig fmt outputs.

I didn't test compiling the output.  I'd wait for the pending zig devel changes to go in first.

floooh:
Very nice, thanks! I'll merge right after the CI pipeline ran through.

floooh:
Ok, here's the plan: I'll merge this PR into a new branch `[bindgen-sokol-zig-0.11.0](https://github.com/floooh/sokol/tree/bindgen-sokol-zig-0.11.0)` and then will use this together with the pending PR in sokol-zig to get the 0.11.0 branch in sokol-zig uptodate.

floooh:
Ok, I think the branch zig-0.11.0 is now back in good shape. This also has the latest headers from the main branch, including example fixes (had some trouble merghing because of the fmt fixes, hopefully zig-0.11.0 isn't too far out so that we can put everything back on the respective master branches)

PR #851, opened 2023-07-02
fix public function implementation prefix for egl functions

AnujAsher:
it was a typo SOKOL_APP_IMPL -> SOKOL_API_IMPL


floooh:
Whooops, thanks for catching that!

PR #858, opened 2023-07-11
Update gen_nim.py

Angluca:
c: int uint = nim: int uint = x64:8byte, x32:4byte
c: type * = nim: ptr type or pointer = c:null, nim=nil
nim use byte or uint8 instead char not use char type name now, 
But have cchar because is c type.


floooh:
Hmm are these changes required because the most recent nim dev version breaks now?

I'd rather wait until nim stabilizes again to reduce the frequency of necessary updates (I'm starting to think that the unsafeAddr => addr change was probably a bit too hasty, because it forces people to move from the stable version to the dev version).

PR #859, opened 2023-07-15
sokol_debugtext.h: fix doc code example typo

stianhj:
Found a missing `(` in debugtext code example.


floooh:
Thanks for catching that!

PR #877, opened 2023-09-06
Fix -Wpointer-sign warnings in sokol_gfx.h

danielchasehooper:
The unused out parameter was the wrong sign, removing it fixes the -Wpointer-sign warnings


floooh:
Hmm, I was confused too when looking at the code, but `GLsizei` is actually typedefed from a regular 32-bit signed integer, so that code should be good.

(also I checked and looks like -Wpointer-sign is part of the default warning set: https://clang.llvm.org/docs/DiagnosticsReference.html#wpointer-sign - otherwise I would have added it to the sokol-test compile options, currently that's `-Wall -Wextra -Werror -Wsign-conversion`)

danielchasehooper:
My bad, one of the headers I was using had `GLsizei` incorrectly defined as `unsigned int`.

PR #878, opened 2023-09-08
fix errant framebuffer bind

danielchasehooper:
the framebuffer was getting bound even when msaa hadn't changed it


floooh:
Oops yeah looks like a bug. This was a "fix" for a rendering corruption problem on Chrome on Windows.

I'll need to check whether the fix was actually only working because of the new bug. I'll try to look into this today (only downside is I have no Windows machine for testing around today).

Thanks for the PR!

floooh:
Ok tested (only on my Mac though) and merged. Many thanks!

PR #881, opened 2023-09-10
fecth response callback called with 'cancelled' flag but not the 'finished' one

guillaumeblanc:
Hi,

I modified sokol fetch unit tests to reproduce a functional behavior that doesn't match with documentation.
When sfetch_cancel is called between sfetch_send and sfetch_dowork, then the response callback is called with cancelled flag but not finished.
This behavior differs from documentation and makes fetch callback implementation less straightforward.

**I couldn't find a fix yet.**
Maybe there's a better way to implement this test with the existing ones.

Side note: Thanks for the incredibly well thought, easy to use and yet very efficient sokol APIs.

Regards,
Guillaume

### DIFF
--- a/tests/functional/sokol_fetch_test.c
+++ b/tests/functional/sokol_fetch_test.c
@@ -811,3 +811,36 @@ UTEST(sokol_fetch, load_file_cancel) {
     T(load_file_cancel_passed);
     sfetch_shutdown();
 }
+
+bool load_file_cancel_finished_failed = false;
+void load_file_cancel_finished_callback(const sfetch_response_t *response)
+{
+    if (response->cancelled && !response->finished)
+    {
+        load_file_cancel_finished_failed = true;
+    }
+}
+
+UTEST(sokol_fetch, load_file_cancel_finished)
+{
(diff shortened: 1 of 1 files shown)


floooh:
Good catch! I actually need to debug a bit what happens in `sfetch_dowork` with such 'early-cancelled' request.

Normally this code should take care of the problem:

https://github.com/floooh/sokol/blob/d4ac122f36d7659a18b312fd4fa2317fb9e06a63/sokol_fetch.h#L2500-L2503

...but the response callback may already be called before that here:

https://github.com/floooh/sokol/blob/d4ac122f36d7659a18b312fd4fa2317fb9e06a63/sokol_fetch.h#L2475-L2477

I think such "early cancelled" requests should be dealt with already in the first loop (https://github.com/floooh/sokol/blob/d4ac122f36d7659a18b312fd4fa2317fb9e06a63/sokol_fetch.h#L2467-L2479), so that such requests are not even dispatched into the `user_incoming` ring buffer (instead call the response callback with finished+cancelled and free the item right away).

I'll try to look into it in the next few days.

PR #893, opened 2023-09-17
take texture target into account when binding to framebuffer

danielchasehooper:
Fixes a crash triggered when creating a sg_pass for a sg_image that had an injected GL_TEXTURE_RECTANGLE texture


floooh:
Good fix, thanks!

PR #897, opened 2023-09-19
make error message match validation logic

danielchasehooper:


floooh:
Good catch, thanks! This validation check changed a while ago to require an exact size match (the comment is also still misleading but I'll fix this after merging your PR).

PR #898, opened 2023-09-19
sokol_fetch.h: fix response state when cancelling a request before dispatch (fixes #882)

### LINKED ISSUE #882: sokol_fetch.h: fetch response callback called with 'cancelled' flag but not the 'finished' one
Hi,

When sfetch_cancel is called between sfetch_send and sfetch_dowork, then the response callback is called with cancelled flag but not finished.
This behavior differs from documentation and makes fetch callback implementation less straightforward.

I modified sokol fetch unit tests to reproduce a functional behavior that doesn't match with documentation, submitted as PR #881

Side note: Thanks for the incredibly well thought, easy to use and yet very efficient sokol APIs.

Cheers,
Guillaume

floooh:
This also sets the `finished`, `failed` and `error_code` fields in the response when a request has been cancelled before it was dispatched, also some minor code cleanup around cancelling requests, and adding two more related tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## Updates
 
+#### 19-Sep-2023
+
+- sokol_fetch.h: fixed a minor issue where a request that was cancelled before it was dispatched
+  had an incomplete response state set in the response callback (the `finished`, `failed` and
+  `error_code` fields were not set). This fixes issue https://github.com/floooh/sokol/issues/882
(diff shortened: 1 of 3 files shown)


floooh:
Closing this PR in favour of https://github.com/floooh/sokol/pull/898 (I replicated your test, and also added another test which checks behaviour when cancelling after the dispatch from outside the response callback).

PR #899, opened 2023-09-19
add callbacks for default gl framebuffer. 

### LINKED ISSUE #892: Rendering to different default gl framebuffers - sokol_gfx.h
In the OpenGL build of our app, each document window has its own `CAOpenGLLayer`. I've looked into using `sg_context`s in order make `sg_begin_default_pass` render to the correct CAOpenGLLayer framebuffer[^1]. Unfortunately sg_contexts affect resource pooling, which I do not want. All documents share the same glyph buffer and other resources for common visual elements, and I don't want those resources to be destroyed when one document is closed[^2]. I do not want changing the default pass's framebuffer to affect anything to do with resource lifetimes. 

I propose adding:
```
typedef struct sg_gl_context_desc {
    const GLuint (*default_framebuffer_cb)(void);
    const GLuint (*default_framebuffer_userdata_cb)(void*);
    void* user_data;
} sg_gl_context_desc;

typedef struct sg_context_desc {
    ...
    sg_gl_context_desc gl;
} sg_context_desc;
```

That way you can return the desired framebuffer via that callback, and would match the apis of the other graphics backends[^3].

Related to #885 

[^1]: Also I haven't seen any documentation guaranteeing that the framebuffer that is bound in `drawInCGLContext:pixelFormat:forLayerTime:displayTime:`/`CGLSetCurrentContext` is always the same. So even if resources weren't a concern with sg_context, it seems possible that the framebuffer bound during sg_setup_context might later not be the same framebuffer when sg_begin_default_pass is called

[^2]:  I have all CAOpenGLLayers use the same CGLContextObj for this reason. 

[^3]:I don't love callbacks since they're harder to read than straight-line code. I think there could be a nicer API along the lines of
sg_begin_pass(sg_default_pass(&(sg_default_pass_desc){
   .gl.framebuffer = fbo, 
   .width = width,
   .height = height
}), &pass_action)
This would allow writing functions that accept a pass as a parameter and that pass could target either an offscreen buffer or the framebuffer; all without affecting the receiving function's implementation. This would also improve #885.

danielchasehooper:
fixes #892

There is a simpler fix of adding this function, but I figured you might not want to add backend-specific public apis.
```
void sg_set_default_gl_framebuffer(GLuint framebuffer) {
    _sg.gl.cur_context->default_framebuffer = framebuffer;
}
```
This is nicer for the end user than the callback approach, because you don't have to create the callback and a global. and set the global before calling sg_begin_default_pass.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -3246,8 +3246,8 @@ typedef struct sg_wgpu_context_desc {
 } sg_wgpu_context_desc;
 
 typedef struct sg_gl_context_desc {
-    unsigned int (*default_framebuffer_cb)(void);
-    unsigned int (*default_framebuffer_userdata_cb)(void*);
+    uint32_t (*default_framebuffer_cb)(void);
+    uint32_t (*default_framebuffer_userdata_cb)(void*);
     void* user_data;
 } sg_gl_context_desc;
 
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -7043,6 +7043,7 @@ _SOKOL_PRIVATE void _sg_gl_reset_state_cache(void) {
 }
 
 _SOKOL_PRIVATE void _sg_gl_setup_backend(const sg_desc* desc) {
+    _SOKOL_UNUSED(desc);
     SOKOL_ASSERT(desc->context.gl.default_framebuffer_cb == NULL || desc->context.gl.default_framebuffer_userdata_cb == NULL);
     
     // assumes that _sg.gl is already zero-initialized
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -3245,6 +3245,12 @@ typedef struct sg_wgpu_context_desc {
     void* user_data;
 } sg_wgpu_context_desc;
 
+typedef struct sg_gl_context_desc {
+    GLuint (*default_framebuffer_cb)(void);
+    GLuint (*default_framebuffer_userdata_cb)(void*);
+    void* user_data;
+} sg_gl_context_desc;
+
 typedef struct sg_context_desc {
     sg_pixel_format color_format;
     sg_pixel_format depth_format;
@@ -3252,6 +3258,7 @@ typedef struct sg_context_desc {
     sg_metal_context_desc metal;
     sg_d3d11_context_desc d3d11;
     sg_wgpu_context_desc wgpu;
+    sg_gl_context_desc gl;
 } sg_context_desc;
 
 /*
@@ -7036,7 +7043,8 @@ _SOKOL_PRIVATE void _sg_gl_reset_state_cache(void) {
 }
 
 _SOKOL_PRIVATE void _sg_gl_setup_backend(const sg_desc* desc) {
-    _SOKOL_UNUSED(desc);
+    SOKOL_ASSERT(desc->context.gl.default_framebuffer_cb == NULL || desc->context.gl.default_framebuffer_userdata_cb == NULL);
+    
     // assumes that _sg.gl is already zero-initialized
     _sg.gl.valid = true;
 
@@ -7725,6 +7733,12 @@ _SOKOL_PRIVATE void _sg_gl_begin_pass(_sg_pass_t* pass, const sg_pass_action* ac
         #if defined(SOKOL_GLCORE33)
         glDisable(GL_FRAMEBUFFER_SRGB);
         #endif
+        if (_sg.desc.context.gl.default_framebuffer_userdata_cb) {
+            _sg.gl.cur_context->default_framebuffer = _sg.desc.context.gl.default_framebuffer_userdata_cb(_sg.desc.context.gl.user_data);
+        } else if (_sg.desc.context.gl.default_framebuffer_cb) {
+            _sg.gl.cur_context->default_framebuffer = _sg.desc.context.gl.default_framebuffer_cb();
+        }
+
         glBindFramebuffer(GL_FRAMEBUFFER, _sg.gl.cur_context->default_framebuffer);
     }
     glViewport(0, 0, w, h);
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -7044,7 +7044,7 @@ _SOKOL_PRIVATE void _sg_gl_reset_state_cache(void) {
 
 _SOKOL_PRIVATE void _sg_gl_setup_backend(const sg_desc* desc) {
     _SOKOL_UNUSED(desc);
-    SOKOL_ASSERT(desc->context.gl.default_framebuffer_cb == NULL || desc->context.gl.default_framebuffer_userdata_cb == NULL);
+    SOKOL_ASSERT(desc->context.gl.default_framebuffer_cb == 0 || desc->context.gl.default_framebuffer_userdata_cb == 0);
     
     // assumes that _sg.gl is already zero-initialized
     _sg.gl.valid = true;
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -3246,8 +3246,8 @@ typedef struct sg_wgpu_context_desc {
 } sg_wgpu_context_desc;
 
 typedef struct sg_gl_context_desc {
-    GLuint (*default_framebuffer_cb)(void);
-    GLuint (*default_framebuffer_userdata_cb)(void*);
+    unsigned int (*default_framebuffer_cb)(void);
+    unsigned int (*default_framebuffer_userdata_cb)(void*);
     void* user_data;
 } sg_gl_context_desc;
 
--- a/bindgen/gen_rust.py
+++ b/bindgen/gen_rust.py
@@ -355,6 +355,8 @@ def funcptr_result_c(field_type):
     res_type = field_type[: field_type.index("(*)")].strip()
     if res_type == "void":
         return ""
+    elif is_prim_type(res_type):
+        return as_rust_prim_type(res_type)
     elif util.is_const_void_ptr(res_type):
         return " -> *const core::ffi::c_void"
     elif util.is_void_ptr(res_type):
--- a/bindgen/gen_zig.py
+++ b/bindgen/gen_zig.py
@@ -271,6 +271,8 @@ def funcptr_result_c(field_type):
     res_type = field_type[:field_type.index('(*)')].strip()
     if res_type == 'void':
         return 'void'
+    elif is_prim_type(res_type):
+        return as_zig_prim_type(res_type)
     elif util.is_const_void_ptr(res_type):
         return '?*const anyopaque'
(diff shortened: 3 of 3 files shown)


danielchasehooper:
Ok looks like I'll have to edit gen_zig.py to fix the error...before I invest more time in this, does this seem like you'll merge this PR once the tests pass?

floooh:
I don't like this solution because it adds yet another fairly random backend-specific public API function. I think it's better to put this stuff as callbacks into the `sg_context_desc` struct, by reintroducing an `sg_gl_context_desc gl;` nested struct. That way it's consistent with the other platforms.

danielchasehooper:
Ok, I think you misunderstood my PR message. This PR actually **does** reintroduce `sg_gl_context_desc`, check out the files changed. The PR message mentioned a simpler approach while acknowledging that you probably wouldn't like it.

sokol_gfx.h
```diff
@@ -7037,6 +7044,8 @@ _SOKOL_PRIVATE void _sg_gl_reset_state_cache(void) {
 
 _SOKOL_PRIVATE void _sg_gl_setup_backend(const sg_desc* desc) {
     _SOKOL_UNUSED(desc);
+    SOKOL_ASSERT(desc->context.gl.default_framebuffer_cb == NULL || desc->context.gl.default_framebuffer_userdata_cb == NULL);
```
floooh, line 7047: You can omit the `== NULL`, or at least replace the `NULL` with `0` (yeah I know theoretically a NULL pointer doesn't have to be zero, but that's a very esoteric assumption).

(I would prefer removing the `== NULL` alltogether.

floooh:
> Ok, I think you misunderstood my PR message. This PR actually **does** reintroduce `sg_gl_context_desc`, check out the files changed. The PR message mentioned a simpler approach while acknowledging that you probably wouldn't like it.

Ah right, I didn't look at the actual changes, only saw that function. The simplified PR looks alright (but please check out my request above).

Cheers!

sokol_gfx.h
danielchasehooper: removing `== NULL` changes the meaning. I think you're thinking of `!= NULL` being removable

danielchasehooper:
I'll let you revisit this PR after a full night's rest before I dig into making changes to gen_zig.py :)

sokol_gfx.h
floooh: ah right... I was confused because in other places I have asserts like this:

`SOKOL_ASSERT(desc->context.d3d11.render_target_view_cb || desc->context.d3d11.render_target_view_userdata_cb);`

...but in this case, one of the two callbacks *must* be set. I wonder if the assert is necessary at all, because it's ok that both callbacks can 0, and if for some reason both callbacks are set, then one of the two will be called anyway (a proper check would be "either both callbacks are zero, or exactly one callback is provided").

floooh:
> I'll let you revisit this PR after a full night's rest before I dig into making changes to gen_zig.py :)

Hmm, from what I'm seeing, this shouldn't be necessary any longer with that setter function removed (this had a GLuint type, which the bindings generator scripts would reject - but in any case, the correct solution in that case would be to remove the GLuint with an uint32_t, since no backend specific types must show up in the public API).

floooh:
(note to self: `sg_begin_default_pass_with_context(pass_action, sg_pass_context ctx)`)?

Pass-context would contain: width, height, pixel formats, sample count + the backend-specific sg_xxx_context structs with callbacks)...

danielchasehooper:
> the correct solution in that case would be to remove the GLuint with an uint32_t, since no backend specific types must show up in the public API).
> 
> PS: CI is still failing though, weird. Anyway, I'm off for today :)

The PR as of right now does not have GLuint in the public API. The CI is failing because the gen_zig.py doesn't support function pointers that return uint32_t, as is required by `uint32_t (*default_framebuffer_cb)(void);` I'll look into fixing that now.

sokol_gfx.h
danielchasehooper: Replaced NULL with 0. I noticed there are 19 other places that NULL is used in sokol_gfx.h, how do you decide when to use NULL and when to use 0?

danielchasehooper:
> (note to self: `sg_begin_default_pass_with_context(pass_action, width, height, sg_context_desc* ctx)`)?

I like this idea. Much nicer than callbacks.

floooh:
I need to think about this idea a bit more, but unrelated to that change, if we would make the association of resource objects with sg_contexts optional, would that help (plus any required changes so that the context would only need to be switched before a default pass? In that situation, a context would basically just describe a separate default framebuffer by pixel formats, sample count and a separate set of callback functions.

You would then create an sg_context for every 'output view' and switch to it right before the default pass:

```c
    sg_activate_context(ctx);
    sg_begin_default_pass(...);
    ...
    sg_end_pass();
```

...unlike the GLFW multiwindow-sample there should only be a single `sg_commit()` call though (not sure yet if this would require more changes).

The sg_begin_default_pass_with_context() would then just be 'API sugar' which would take an sg_context handle instead, and just get rid of the `sg_activate_context()` pass.

But TBH, this might require more internal changes than I'm currently ready to integrate ;)

So I'd actually first like to go with a 'minimally invasive' procedure that works for your use case, and after that maybe continue working on a cleaner overall structure.

(e.g. another thing that popped into my mind was having a single `sg_begin_pass()` function which takes a parameter struct, which would be configured differently for different types of passes, e.g. an offscreen pass would be:

```c
sg_begin_pass(&(sg_pass_options){
    .pass = pass_object,
    .action = { ... },
});
```

A vanilla default pass would look like this:
```c
sg_begin_pass(&(sg_pass_options){
    .width = sapp_width(),
    .height = sapp_height(),
    .action = { ... },
});
```

...and a default pass into a specific context would look like this:

```c
sg_begin_pass(&(sg_pass_options){
    .width = sapp_width(),
    .height = sapp_height(),
    .action = { ... },
    .context = ctx,
});
```

Also, a better name for `sg_context` and `sg_context_desc` would actually be `sg_swapchain` and `sg_swapchain_desc`.

These are big API changes that I don't want to do in a random 'drive-by' PR though, but I feel like this is would provide a much cleaner overall structure.

...btw, on GL, are you actually using a single GL context? If yes how does that work in the window system glue, what I've seen so far is that each 'view' has its own GL context (like in GLFW for instance).

danielchasehooper:
If you want a minimally invasive approach that handles my use case, the changes in this PR as it stands now *is* that, and brings the GL API in line with the others (in terms of all having callbacks). 

If we want to go with a more meaningful change, I think a single "begin pass" function that can be used for both offscreen and default passes is the nicest solution. That would allow user functions to accept a `sg_pass_options` parameter and draw to either offscreen or default framebuffer without special cases. In my codebase I have a function kind of like that called `sg_begin_surface_pass(Surface *surface, sg_pass_action *action)`, where the surface can represent either the default framebuffer or an offscreen pass. Actually now that I'm thinking about it, we could add a `sg_begin_pass_options(sg_pass_options *, sg_pass_action *)` function (fairly non-invasively?) that sets a few internal state variables and then calls _sg_begin_pass. I feel like the pass action should be a separate parameter from the options struct since it's required and "what we're drawing to" is separate from "what we draw to it (load/clear)". 

I think altering the behavior of sokol_gfx's contexts is less appealing because drawing to a different destination is conceptually more like a pass and less like GL's concept of a context. And like you said, contexts are sort of an old idea that doesn't fit well with newer apis, so you don't really want to invest more in that concept. Rendering to multiple destinations is something that I want to do with Metal too.

So in summery, (Depending on how much you want to alter the API) we could do one of these 3:
- stick with callbacks; accept this PR as is. 
-  *add* `sg_begin_pass_options(sg_pass_options *, sg_pass_action *)` (non-api-breaking)
-  *update* `sg_begin_pass` to accept options struct, and *remove* `sg_begin_default_pass` (api-breaking)

---

>...btw, on GL, are you actually using a single GL context? If yes how does that work in the window system glue, what I've seen so far is that each 'view' has its own GL context (like in GLFW for instance).

Yes, I use a single GL context. I overrode this function in CAOpenGLLayer to return the same global `CGLContextObj` for all layers

```
// from CAOpenGLLayer.h

/* Called by the CAOpenGLLayer implementation when a rendering context
 * is needed by the layer. Should return an OpenGL context with
 * renderers from pixel format 'pf'. The default implementation
 * allocates a new context with a null share context. */

- (CGLContextObj)copyCGLContextForPixelFormat:(CGLPixelFormatObj)pf;
```

Later I get called in `- (void)drawInCGLContext:(CGLContextObj)glContext
             pixelFormat:(CGLPixelFormatObj)pixelFormat
            forLayerTime:(CFTimeInterval)timeInterval
             displayTime:(const CVTimeStamp *)timeStamp` to do the actual drawing. the `glContext` is the same for all layers, but when I call `CGLSetCurrentContext(glContext);`, a unique GL_DRAW_FRAMEBUFFER_BINDING is set.

---

>...unlike the GLFW multiwindow-sample there should only be a single sg_commit() call though (not sure yet if this would require more changes).

Given that each CAOpenGLLayer calls *me* when it needs redrawing, I'd have to find a way to run sg_commit() at the end of the current event loop in order to guarantee it is called after any/all windows did their drawing. Sort of messy, but I don't see a way to share mutable sg resources between windows without that requirement.

On another note, thanks for taking my feedback seriously and wanting to get this right.

floooh:
> If you want a minimally invasive approach that handles my use case, the changes in this PR as it stands now is that, and brings the GL API in line with the others (in terms of all having callbacks).

Oki doki, then let's do this as the first step! I guess I'll take another week for catching up on PRs and issues, before going back to the WebGPU backend.

Thanks for your input btw, it helps me to look differently at old ideas that had been baked into the API from the start. Thinking back I never was quite happy with the `sg_begin_default_pass()` function, I think a unified `sg_begin_pass()` with flexible options, and some sort of swapchain abstraction is the way forward.

floooh:
Looking into this now...

floooh:
...and merged! I also upated a small blurb to the changelog.

floooh:
...oops, something broke in the Rust bindings, need to figure out why this doesn't trigger in the regular tests. I'm on it...

https://github.com/floooh/sokol/actions/runs/6285075508/job/17067124611#step:5:113

floooh:
...ok, this was just a missing `->` before the return type. Still, I need to figure out why this didn't show up in the CI pipeline.

floooh:
...heh ok the `test-rust` step in the Rust bindings GH Actions workflow file used the wrong artifact download path, also explains some weird behaviour I saw sporadically in the past.

danielchasehooper:
Thanks for accepting the PR! Looking forward to helping with #904 next!

PR #900, opened 2023-09-19
call _SG_ERROR/_SG_WARN *after* _SG_LOGMSG 

danielchasehooper:
call _SG_ERROR/_SG_WARN *after* _SG_LOGMSG so that if the logging callback panics on errors/warnings, the log message will have made it to the console before the panic. Really helps with debugging.


floooh:
Hmm, the original behaviour is actually intended. Both errors and warnings are not supposed to terminate execution, only the panic log level.

For instance, in the validation layer, there's a single panic at the end if any validation had failed (and before that the validation layer may output multiple error log messages):

https://github.com/floooh/sokol/blob/b803c9a0214c6ab6dcb9cc6dd9d30d7ace4eda1e/sokol_gfx.h#L14656

danielchasehooper:
My apologies, I thought it was expected that errors would cause an abort. I misread the sokol_log implementation.

Issue #843, opened 2023-06-19
emscripten main loop hack


voidware:
Hi,

Just got the latest code. A few minor changes needed here, no problems. Seems working fine. thanks.

So;

I'd really like to eliminate my Sokol hacks in `sokol_app.h` so to use the official codebase. Firstly, thanks for fixing the `emsc_event->code[0]` keyboard problem. One less hack of mine needed!

This is an emscripten build and I thought I'd try to eliminate another hack. I've had this hack in my code for over a year and it works fine, but the official code just produces a blank screen with no errors.

```
_SOKOL_PRIVATE void _sapp_emsc_run(const sapp_desc* desc) {
...

    /* start the frame loop */

    // this is the sokol version 
    ///emscripten_request_animation_frame_loop(_sapp_emsc_frame, 0);

    // this is my version
    emscripten_set_main_loop(_sapp_emsc_frame_void, 0, true);
}
```

and;

```
_SOKOL_PRIVATE void _sapp_emsc_frame_void()
{
    // XX HACK!
    _sapp_emsc_frame(0, 0);
}
```

I'm at a loss to explain why the official main loop doesn't work and mine does because i can't see anything wrong with it.

Any ideas? Thanks.

floooh:
Don't know tbh, I just tried the reverse by putting your code into sokol_app.h and that worked (even though it passes zero to the time parameter of the _sapp_emsc_frame() function (which I would expect to cause problems, e.g. divisions by zero).

Is your `_sapp_emsc_frame()` function otherwise identical? I would check next to put some printf() into the frame callback to check whether this is actually called for each frame.

voidware:
Hi. Thanks for your reply.

Standard Sokol emsc flow, i get one frame called and that's it.

Here's what happens;


```
int main(int argc, char* argv[]) {
    Sp_desc desc = sokol_main(argc, argv);
    _sapp_emsc_run(&desc);
    return 0;
}
```

then;

```
_SOKOL_PRIVATE void _sapp_emsc_run(const sapp_desc* desc) {
...               
    emscripten_request_animation_frame_loop(_sapp_emsc_frame, 0);
}
```

This renders a single frame, and then returns from `_sapp_emsc_run` which returns to `main`, which return.

Then everything stops!

*However*

Doing this;

```
_SOKOL_PRIVATE void _sapp_emsc_run(const sapp_desc* desc) {
...               
    emscripten_set_main_loop(_sapp_emsc_frame_void, 0, true);
}
```

[...]

PR #915, opened 2023-10-07
Use `emscripten_set_main_loop` in sokol app

ambrusc:
Problem: It appears that Emscripten's async APIs don't get a chance to call user callbacks in Sokol App. For instance, when using [`emscripten_async_wget_data`](https://emscripten.org/docs/api_reference/emscripten.h.html#c.emscripten_async_wget_data), Firefox shows the request completed with 200 OK, and the data available, but the user callbacks are never called.

Root Cause: From [docs about `emscripten_request_animation_frame_loop`](https://emscripten.org/docs/api_reference/html5.h.html#c.emscripten_request_animation_frame_loop), it seems to be intended as a low level API, and is likely not handling callbacks. Switching to `emscripten_set_main_loop` fires the callbacks, and still uses [`requestAnimationFrame` under the hood](https://emscripten.org/docs/api_reference/emscripten.h.html#c.emscripten_set_main_loop).

Proposal: Use `emscripten_set_main_loop` instead of `emscripten_request_animation_frame_loop` in Sokol App.

NOTE: I haven't dug too deeply into what else `emscripten_set_main_loop` does, or if it duplicates functionality elsewhere in Sokol App. Maybe we just need to hook up the async API callbacks explicitly somewhere else?

I don't mean to open a glut of pull requests, but I thought that maybe others could benefit from some of these findings. If I've misunderstood something or these PRs are not helpful, please just let me know and I'll close them! :-)

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -5782,8 +5782,8 @@ _SOKOL_PRIVATE void _sapp_emsc_unregister_eventhandlers() {
     #endif
 }
 
-_SOKOL_PRIVATE EM_BOOL _sapp_emsc_frame(double time, void* userData) {
-    _SOKOL_UNUSED(userData);
+_SOKOL_PRIVATE void _sapp_emsc_frame() {
+    const double time = emscripten_performance_now();
     _sapp_timing_external(&_sapp.timing, time / 1000.0);
 
     #if defined(SOKOL_WGPU)
@@ -5824,9 +5824,8 @@ _SOKOL_PRIVATE EM_BOOL _sapp_emsc_frame(double time, void* userData) {
         _sapp_emsc_unregister_eventhandlers();
         _sapp_call_cleanup();
         _sapp_discard_state();
-        return EM_FALSE;
+        emscripten_cancel_main_loop();
     }
-    return EM_TRUE;
 }
 
 _SOKOL_PRIVATE void _sapp_emsc_run(const sapp_desc* desc) {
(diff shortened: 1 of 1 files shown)


floooh:
Hmm, there was a discussion about `emscripten_set_main_loop()` vs `emscripten_request_animation_frame_loop()` here:

https://github.com/floooh/sokol/issues/843

...the callback / event loop problem is new to me though. If the completion callbacks in emscripten_async_wget_data() are not called that sounds more like a bug in the Emscripten runtime to me.

My 'counter proposal' would be to implement the set-main-loop method as an alternative, activated by a preprocessor define similar to `SOKOL_NO_ENTRY` or `SOKOL_WIN32_FORCE_MAIN`.

(my main gripe is that the frame timestamp needs to be aquired separately instead of being provided by the request-animation-loop callback, this might theoretically introduce more jitter).

ambrusc:
Off topic, but I just wanted to thank you for creating this awesome set of libraries. The more I use them the more I really really like them. I'm also really enjoying the initializer/desc-style.

I didn't want to make too much more traffic on these PRs as you have a much better understanding of how things work under the hood. You make a good point about timestamp jitter for instance. Blargh. (Why can't platform-APIs be consistent?)

floooh:
Closing this because I just merged https://github.com/floooh/sokol/pull/997 in favour of this PR (also see changelog entry: https://github.com/floooh/sokol/blob/master/CHANGELOG.md#02-mar-2024).

Still, many thanks for the PR of course!

PR #919, opened 2023-10-19
Allow overriding the gl error check macro

danielchasehooper:
Our project had horrible development performance due to all of sokol's `glGetError` calls (that function is super slow in chromium), this PR allows us to disable sokol's gl error checks, while leaving all of sokols asserts enabled for development. At the end of the frame we `assert(glGetError()==0)` as a sanity check. if that assert ever gets triggered, we can re-enable sokol's default `_SG_GL_CHECK_ERROR` definition for debugging


floooh:
Makes sense, yeah. Thanks!

PR #933, opened 2023-10-31
ifdef out GL_FRAMEBUFFER_UNDEFINED from webgl builds

danielchasehooper:
GL_FRAMEBUFFER_UNDEFINED doesn't exist in webgl2, ifdef it out.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -8299,9 +8299,11 @@ _SOKOL_PRIVATE sg_resource_state _sg_gl_create_pass(_sg_pass_t* pass, _sg_image_
         const GLenum fb_status = glCheckFramebufferStatus(GL_FRAMEBUFFER);
         if (fb_status != GL_FRAMEBUFFER_COMPLETE) {
             switch (fb_status) {
+                #if defined(SOKOL_GLCORE33)
                 case GL_FRAMEBUFFER_UNDEFINED:
                     _SG_ERROR(GL_FRAMEBUFFER_STATUS_UNDEFINED);
                     break;
+                #endif
                 case GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT:
                     _SG_ERROR(GL_FRAMEBUFFER_STATUS_INCOMPLETE_ATTACHMENT);
                     break;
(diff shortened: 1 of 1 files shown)


danielchasehooper:
Not sure if `#if !defined(SOKOL_GLES3)` would be better? Not that it matters right now, but maybe that would be more robust to future gl versions?

floooh:
Strange, sokol_gfx.h includes `GLES3/gl3.h` on Emscripten:

https://github.com/floooh/sokol/blob/bd7fa9300dc2e2edca087cf436d1c5c07bd754fc/sokol_gfx.h#L4217-L4221

...and this includes the definition:

https://github.com/emscripten-core/emscripten/blob/4de99429ecb854e346b32edc04cfa261c37b05d9/system/include/GLES3/gl3.h#L830

The Emscripten builds are also working fine in CI:

https://github.com/floooh/sokol-samples/actions/runs/6697850493

https://github.com/floooh/sokol/actions/runs/6697703312

No idea how it could fail tbh...

floooh:
Ah, now I see, the WebGL2 spec removed it (see: https://registry.khronos.org/webgl/specs/latest/2.0/)

```
5.8 Default Framebuffer
WebGL always has a default framebuffer. The FRAMEBUFFER_UNDEFINED enumerant is removed from the WebGL 2.0 API.
```

Since Emscripten uses the GLES3/gl3.h header it compiles though even if WebGL2 doesn't have that error value. Are you using different headers than what Emscripten provides?

danielchasehooper:
Ah, yeah we're not using emscripten. we use clang's wasm target directly and wrote our own webgl headers and javascript glue. I guess we'll go the route of matching GLES3 headers. Thanks for the comments!

floooh:
FYI I have solved this a bit differently (will be part of https://github.com/floooh/sokol/pull/985):

```c
#if defined(SOKOL_GLES3)
    // on WebGL2, GL_FRAMEBUFFER_UNDEFINED technically doesn't exist (it is defined
    // in the Emscripten headers, but may not exist in other WebGL2 shims)
    // see: https://github.com/floooh/sokol/pull/933
    #ifndef GL_FRAMEBUFFER_UNDEFINED
    #define GL_FRAMEBUFFER_UNDEFINED 0x8219
    #endif
#endi
```

Issue #535, opened 2021-07-06
sokol_app vulkan


pplux:
sokol_app is an extraordinary library, I'm using it with my own render implementation all the time. I have plans to switch to vulkan and I was wondering if you plan to add support for vulkan (not the gfx part, just the app/context) or if you will be willing to accept PR's working towards that goal.

As said, not asking for #285 , just the sokol_app part.

floooh:
Hmm, not a fan of the idea for at least two reasons:

- the Vulkan setup wouldn't be some isolated chunk of code, but instead needs to be implemented inside the existing platform backends which support Vulkan (Win32, Linux, Android, possibly macOS), that's a lot of "disruption" for the existing code
- without a Vulkan backend in sokol_gfx.h, the Vulkan code in sokol_app.h would "just sit there" without tests and samples using that code

Once it makes sense to remove all OpenGL backends (for instance because of bitrot in OpenGL implementations) I'll be most likely more in favour of the idea, because this would mean ripping out some big and ugly code. Vulkan would probably be more code, but definitely not as ugly as the GL code. But that's not anytime soon I think.

There's also a big change in (very slow) progress implementing multiwindow support for sokol_app.h, so currently isn't a good time to do big PRs anyway I'm afraid.

pplux:
Thanks! I will stick to glfw meanwhile :)

cshenton:
So for anyone else going down this path, I've had good results so far with just not defining a graphics API, then commenting out the bits of `sokol_app.h` that complained (the preprocessor errors and the corresponding function calls to undefined methods). 

That so far has enabled me to rely on sokol to handle the window creation, etc., and I just used `sapp_win32_get_hwnd` to get the information I needed to create a `VkSurface` on my platform. 

I might be in a minefield though. Time will tell 😆

pplux:
I've done something similar for manual webgpu, and it also worked... so it might be worth giving a try? Maybe it could be a sokol_app with no backend and let the user deal with the specifics for vulkan/webgpu/anything else....

mr-sihc:
how far did this go in the time?
i mean pretty much anything uses vulkan nowdays
so, this would probably be a very good thing!

how far is it/ if at all?

floooh:
[...]

PR #938, opened 2023-11-10
Add support for SOKOL_NOAPI backend

castano:
I often want to use sokol_app on Vulkan applications. Vulkan support has been requested in the past: https://github.com/floooh/sokol/issues/535, but it does not appear that this will be offered in the near future. 

However, it's fairly easy to use Sokol for the window creation and event handling functionality and keep all the Vulkan-specific code in the application code. In my initial implementation I was simply providing a VkInstance in the app description and sokol_app's only responsability was to create a VkSurface. I think that was unnecessarily complicated and caused problems with the shutdown order. The surface needs to be destroyed before the instance, but after the swapchain, so it's not possible to rely on the cleanup callback to destroy all the vulkan resources at once. Instead it was much more simple to move all the responsibilities to the app and make the sokol_app code API agnostic.

So far it does not seem to be necessary to add additional callbacks. In my applications API initialization is done on init, shutdown on cleanup, and resizing of the swapchain in response to the SAPP_EVENTTYPE_RESIZED event.

This certainly needs more testing and validated on other platforms, but I'm putting it out as a draft PR in order to gather feedback.

I think one could also use this feature with other APIs such as OpenGL, Metal and D3D11, in order to gain more control over some aspects of the swapchain, such as adding support for custom formats or controlling the presentation mode, without having to modify sokol_app's internals.

Note, I added the `const void* sapp_android_get_native_window(void)` API in order to obtain a pointer to the `ANativeWindow`, which is necessary in order to create a Vulkan surface. I also moved the EGL attributes out of the _sapp_android_t struct and used the _sapp_egl_t struct instead. This is what's causing most of the changes in this PR.


floooh:
Apologies for not responding yet. I'll try to catch up with PRs over the next couple of weeks.

pplux:
Ah! I was going to do the same thing and propose it as a PR, but @castano did it first. 

My use case is playing with webgpu on native, so I will probably need the NO_API also for mac and linux (I can have a look). In any case, it will be great to add this to sokol_app as an option to let sokol handle the window, but let the user manually create the graphical context.

pplux:
Hi! I was going to suggest also a "NOAPI" approach, in my case to manually handle webgpu on native (linux, mac, windows). It will be very useful, and practical to only use sokol_app for the window creation, events, etc... but manually let the user handle the graphical context. 

I would like to do this also on linux and mac, but for now I will only test on mac, will you be open @floooh to this?

@castano I can help with mac, and maybe linux later.

pplux:
At least, speaking for webgpu, existing init and cleanup methods are more than enough to get it working with no issues. It will be useful to have a post-window closed cleanup too, but I can manage to work around it by using the NOENTRY approach and controlling exactly when sokol ends. With a extra post-cleanup callback I wouldn't need it at all, but it is fine.

pplux:
I built a minimal native wgpu app with sokol, using @castano NOAPI define here:

https://github.com/pplux/minimal-wgpu-native

It was needed on Windows, otherwise it will crash at destruction when the GL context is destroyed (after wgpu overrides it).

On Mac, the Metal backend seems to work fine out of the box.

dortamiguel:
I also have a vulkan application and I will like to use sokol only to handle the window and the events. Is it hard to implement a custom vulkan context that creates the surface? I'm not sure where to start to make something like that

pplux:
Hi @floooh, is it anything we could do to help with the NOAPI approach? I'm still thinking this is a veeery nice feature for sokol_app for others to build on different render backends (like wgpu in my case). Can we help in any way?

PR #942, opened 2023-11-21
In nim bindings, set NDEBUG in release mode

geotre:
related to [this discussion](https://github.com/floooh/sokol-nim/pull/25), this PR adds a simple check that sets NDEBUG when the Nim compiler is called with the release flag


floooh:
Thanks! I'll let the CI pipeline check the PR, but will then merge it tomorrow evening. It's already late here :)

floooh:
And merged. Thanks!

PR #943, opened 2023-11-24
Update generated Nim converter asserts

geotre:
See [this issue](https://github.com/floooh/sokol-nim/issues/26). This PR fixes the asserts to use `<=`


floooh:
And merged. Thanks!

PR #948, opened 2023-12-01
Using const for local variables that are never mutated

abhirag:
This is in support of https://github.com/floooh/sokol-zig/pull/47


floooh:
Thanks!

PR #959, opened 2023-12-21
sokol_nuklear.h: change snk_handle_event to return a bool if handled

adamrt:
This method was recommended by @vurtun: https://github.com/vurtun/nuklear/issues/189#issuecomment-236587263

This is an older comment but the functionality works as expected.

### DIFF
--- a/util/sokol_nuklear.h
+++ b/util/sokol_nuklear.h
@@ -145,7 +145,11 @@
     --- if you're using sokol_app.h, from inside the sokol_app.h event callback,
         call:
 
-        void snk_handle_event(const sapp_event* ev);
+        bool snk_handle_event(const sapp_event* ev);
+
+        This will feed the event into Nuklear's event handling code, and return
+        true if the event was handled by Nuklear, or false if the event should
+        be handled by the application.
 
     --- finally, on application shutdown, call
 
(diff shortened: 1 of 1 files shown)


adamrt:
Related: Why does the current code store the state in new frame, but then apply it in event handling. Why not just do it all in the event handling instead of carrying the state through.

adamrt:
If there is anything you would like adjusted or changed, please let me know @floooh

floooh:
Starting to look into the PR later today, apologies for the delay.

adamrt:
No apology necessary. Thanks for taking the time.

floooh:
...please ignore the failed Emscripten CI run, unrelated to the PR and already fixed in master.

floooh:
> Related: Why does the current code store the state in new frame, but then apply it in event handling. Why not just do it all in the event handling instead of carrying the state through.

IFIR the sokol_nuklear.h header was mostly copied over from sokol_imgui.h by the original author (sokol_imgui.h has diverged in the meantime because the Dear ImGui input model has changed to input events -> to fix exactly the problems which made this per-frame code necessary in the first place).

Specifically for mouse buttons, it was necessary to delay a button-up into the next frame, otherwise on touchpads it would frequently happen that at button-down happened in the same frame as button up, which causes the event to be ignored (Dear ImGui fixed this by adding an internal event queue).

floooh:
...and merged. Thanks!

I also added a small blurb to the changelog.

PR #961, opened 2023-12-26
Fix Managed buffers updates on Metal with discrete GPU

valentinivanov:
It seems that a typo in enum names prevented Metal buffer updates in Managed mode on Metal backend with a discrete GPU (for example, MBP 2019 16" with AMD graphics). As a result many samples relying on buffer updates wouldn't display anything (i.e. ImGUI, Nuklear, animations etc).


floooh:
Wow yeah, definitely a bug. Thanks for catching that. Unfortunately I can't directly test this since I only have an M1 Mac and mid-2014 Intel Mac for testing.

floooh:
And merged, thanks!

Issue #727, opened 2022-10-15
iOS/macOS: finally get rid of MTKView


floooh:
MTKView is just too brittle, with all sorts of workaround for regressions, which then at some point are no longer needed, but cause different issues. Bite the bullet and replace MTKView with CAMetalLayer (not quite trivial because this means sokol_app.h needs to manage its own depth-stencil-buffer and MSAA render target, and probably needs to implement all sorts of shit to make resizing and device rotation work.

DagAgren:
When using CAMetalLayer, I never bother listening to size changes. Instead, when starting each render, I just check `currentDrawable.texture.width` and `currentDrawable.texture.height`, which will contain the correct current pixel size.

floooh:
Fixed via https://github.com/floooh/sokol/pull/1444 and https://github.com/floooh/sokol/pull/1447

Issue #700, opened 2022-08-06
Research: macOS MTKView wobbly resizing...


floooh:
Resizing with MTKView results in a 'wobbly' window content because the backing store doesn't exactly the current window size, and MTKView scaling the content.

See: https://thume.ca/2019/06/19/glitchless-metal-window-resizing/

What appears to work great for high-dpi apps is to set the MTKView's layerContentsPlacement policy to topleft, and enable autoResizeDrawable:

```objc
_sapp.macos.view.layerContentsPlacement = NSViewLayerContentsPlacementTopLeft;
_sapp.macos.view.autoResizeDrawable = true;
```
...and then not manually setting the view's drawable size (however whether that's done or not doesn't seem to make a difference).

This gives a wonderfully stable window content when resizing. Unfortunately it breaks under some circumstances:

- non-highdpi apps only render into the top-left quarter
- moving a highdpi window from a Retina screen to a non-Retina screen has the same effect (only rendering into the top-left quarter)

...this needs more investigation, also consider dropping MTKView completely.

floooh:
Fix in #963, many thanks to @Seb-degraff for picking up this long standing issue :)

PR #963, opened 2024-01-07
macOS: workaround wobbly MTKView resizing

Seb-degraff:
Workaround for the MTKView resizing glitch by "anchoring" the layer to the window corner opposite to the currently manipulated corner (or edge). This prevents the content stretching back and forth during resizing. This is basically this: https://github.com/floooh/sokol/issues/700

As pointed out in the [glitchess resizing](https://thume.ca/2019/06/19/glitchless-metal-window-resizing/) blogpost, `layerContentsPlacement = .topLeft` helps because it prevents stretching the content of the layer while waiting for the correctly sized frame to be rendered. However it works only if the rendered elements are top left aligned. If there are elements that are rendered in the bottom right of the frame and the user resizes from the top left of the window, these elements will appear to shake as they get moved when resizing, then "aligned back" when the frame is rendered.

My solution is to detect which corner is being resized and set layerContentsPlacement accordingly in windowWillStartLiveResize.

Before:

https://github.com/floooh/sokol/assets/6556843/2172ed47-7ef7-476e-ac92-a5c74077e04d

Notice the blue squares stretching while resizing. Pausing the video near the end makes it even more obvious. 

After:

https://github.com/floooh/sokol/assets/6556843/7047c9ba-244d-4421-ac43-7ada9a393b53

Notice the blue squares stability independent of which corner is being dragged. Pausing the video makes the latency apparent in the form of grey padding on the resized sides. Barely noticeable at 120 fps.

This workaround can be removed if/when migrating to CAMetalLayer: https://github.com/floooh/sokol/issues/727

Untested: dragging a window from a high dpi to a non high dpi monitor (I can test this in a few days if needed).

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3560,6 +3560,7 @@ _SOKOL_PRIVATE void _sapp_macos_update_dimensions(void) {
     else {
         _sapp.dpi_scale = 1.0f;
     }
+    _sapp.macos.view.layer.contentsScale = _sapp.dpi_scale; // NOTE: needed because we set layerContentsPlacement to a non-scaling value in windowWillStartLiveResize.
     const NSRect bounds = [_sapp.macos.view bounds];
     _sapp.window_width = (int)roundf(bounds.size.width);
     _sapp.window_height = (int)roundf(bounds.size.height);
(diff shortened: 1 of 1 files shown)


floooh:
Wow, interesting and unexpectedly simple fix! I'll give it a whirl tomorrow and see how it looks on my side!

floooh:
Gave it a whirl and it looks great!

I tested on the configs I have here with the samples imgui-sapp and imgui-highdpi-sapp:

- 2021 M1 14"MBP, Sonoma 14.1.2: looking perfect
- 2014 Intel 13"MBP, latest Big Sur: looking perfect
- an external 1080p@60Hz monitor attached to the M1 Mac: moving windows to the external display, resizing has different types of artefacts but all in all a slight improvement than before (for some reason, *everything* has always been slower on this external monitor connected to my M1 Mac, so we shouldn't pay too much attention to this config, on my old Mac with 60Hz integrated display it looks as good as one can expect)

...going to merge now...

floooh:
...and merged. Many thanks for picking up the issue and creating this perfect little PR :)

...will also add a changelog update now.

Seb-degraff:
Thank you for testing and merging so fast! And for the kind words!

By the way I really appreciate your work on the sokol libs, it's not an understatement to say they changed my programmer life! So I'm glad to be able to contribute back :)

For good measure I also tested this setup:
 - 2023 M2 Pro 16'' with an external 4K monitor manually set to 1080p@60Hz (so not retina). Resizing works as expected. One minor thing I noticed is that when dragging the window from the high-dpi screen to the 1080p screen, there are two frames after the transition that are wrong, 2x too big. I guess that's because the scaleFactor is adjusted right away, but there is already frames in flight with the old frame buffer resolution? I think this such a rare situation that we don't need to bother fixing it though... Here is a screen capture of the artefact:

https://github.com/floooh/sokol/assets/6556843/1b1f574a-6f2a-459a-95d9-01f28409c846

PR #965, opened 2024-01-13
sokol_app.h: GLES support on Linux with GLX

automainint:


floooh:
What would be the use case for GLES support on desktop operating system?

FWIW, the GLX support in sokol_app.h is more or less on the chopping block (only EGL in the future).

I'm not entirely opposed to the PR, but I need a good real-world use case, since the combinatorial explosion of build options in the Linux backend which need to be tested is already now worrying ;)

automainint:
> What would be the use case for GLES support on desktop operating system?

I want the same graphics API so I don't have to write different implementations for a cross-platform application when it's not performance-critical.

I think, the most inconvenient thing when working with different APIs is that GLES has it's own version of the shading language. Perhaps there's a different way to achieve similar results, e.g. using another shading language that will compile into desktop GLSL or GLSL ES, but I'm not sure if that's easier that just initializing the same context.

floooh:
> Perhaps there's a different way to achieve similar results, e.g. using another shading language that will compile into desktop GLSL or GLSL ES, but I'm not sure if that's easier than just initializing the same context.

As I wrote in the issue thread, the best option is to use one of the shader cross-compilation tools. If there would be a single 3D API that works across all platforms all the 3D wrapper APIs and shader cross-compilers wouldn't need to exist in the first place ;)

The two popular libraries I'm aware of which integrate a shader cross-compiler into the runtime are ANGLE (for GLES) and the WebGPU implementation (Dawn and wgpu.rs), integrating the shader compiler into the runtime adds a lot of binary bloat though, that's for instance why sokol_gfx.h delegates the shader cross-compilation step to offline tools.

floooh:
Btw, when using the Linux backend with EGL, there is already GLES3 support. Would hat work for you? I've been planning to kick out GLX support anyway at some point.

automainint:
> Btw, when using the Linux backend with EGL, there is already GLES3 support. Would hat work for you? I've been planning to kick out GLX support anyway at some point.

It works. I decided to stay on GLCORE for desktop and GLES for web or mobile. I use nanovg for drawing, it supports both GLCORE and GLES, and I almost don't have to interact with OpenGL directly.

PR #967, opened 2024-01-16
Additional texture pixel formats [BC3, BC7, ETC2 with sRGB, ASTC 4x4]

allcreater:
This request is adding support for some broadly used compressed texture formats, which could be very useful with the basis_universal texture transcoding library.

Extended set of "known" formats with SRGB support:
  * BC3_SRGBA
  * BC7_SRGBA
  * ETC2_SRGB8
  * ETC2_SRGB8A8
 
 "Brand-new" ASTC format:
  * ASTC_4x4_RGBA,
  * ASTC_4x4_SRGBA
    
4x4 subset is selected as the most broadly used [and the only one of ASTC formats that is supported by the transcoding library](https://github.com/BinomialLLC/basis_universal/blob/ad9386a4a1cf2a248f7bbd45f543a7448db15267/transcoder/basisu_transcoder.h#L69C26-L69C26).

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -1648,6 +1648,13 @@ typedef enum sg_pixel_format {
     SG_PIXELFORMAT_ETC2_RG11,
     SG_PIXELFORMAT_ETC2_RG11SN,
 
+    SG_PIXELFORMAT_BC3_SRGBA,
+    SG_PIXELFORMAT_BC7_SRGBA,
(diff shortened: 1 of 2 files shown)


floooh:
Giving it a quick test now...

floooh:
And merged. Thanks! 

Only thing I changed was moving the SRGB pixel formats closer to their non-SRGB equivalents in the `sg_pixel_format` enum, and a small if-else coding style change.

PR #970, opened 2024-01-20
Fix strict-prototype warnings.

waywardmonkeys:
Need to specify `(void)` as the empty arg list in C rather than just `()`.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -537,7 +537,7 @@
     Like clipboard support, drag'n'drop support must be explicitly enabled
     at startup in the sapp_desc struct.
 
-        sapp_desc sokol_main() {
+        sapp_desc sokol_main(void) {
             return (sapp_desc) {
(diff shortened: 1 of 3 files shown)


floooh:
Thanks!

The Emscripten CI problem is unrelated (looks like the latest Emscripten SDK has some breaking changes in the WebGPU APIs).

waywardmonkeys:
> Thanks!
> 
> The Emscripten CI problem is unrelated (looks like the latest Emscripten SDK has some breaking changes in the WebGPU APIs).

Maybe I'll take a look at fixing that ... tomorrow perhaps since it is almost 11pm here now on Saturday. :)

floooh:
Don't worry, I'll look into it now. Have a good night :)

floooh:
Ok, I have `-Wstrict-prototypes` enabled now for the Clang and GCC C builds in the CI tests. Thanks for making me aware of the issue :)

Issue #979, opened 2024-01-25
Redundant buffer bindings on metal when only changing vertex buffer offset


staminajim:
In metal, if you send commands which only change the vertex buffer offset, (ie. not binding a different buffer) you get lots of warnings about unnecessary bindings in Xcode. This is pretty common when using shared resource pools.

According to the Apple documentation you should use `setVertexBufferOffset` instead of `setVertexBuffer` on the render encoder: https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/MTLBestPracticesGuide/BufferBindings.html

Incoming PR which I'm using in my engine which resolves this

PR #980, opened 2024-01-25
Don't rebind vertex buffers in Metal if only the offset index has changed

staminajim:
In Metal, if you're reusing a vertex buffer with a different offset, you can use `setVertexBufferOffset` instead of `setVertexBuffer`: https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/MTLBestPracticesGuide/BufferBindings.html

If you don't you get loads of redundant buffer binding warnings when running in Xcode (and presumably this is also less efficient for sending data to the GPU)

Fixes https://github.com/floooh/sokol/issues/979

I'm using this change in my game engine and seems to be working perfectly fine


floooh:
The fix looks suspiciously simple ;)

TBH I'm not sure why I haven't done this in the first place when all the caching machinery is already there, I will have to dig a bit into the code and git history to make sure this doesn't revert any previous fixes.

But yes, I agree that it is at least a good thing to get rid of those Metal debugger warnings. I'll try to look into the PR during the weekend.

floooh:
...it's crazy how brittle GH Actions is... (iOS build failed because it couldn't clone sokol)

floooh:
I'll give this PR a whirl now.

floooh:
Ah right, there *is* already a code path which skips the setVertexBuffer alltogether if both the buffer and offset match, just not the path for only updating the offset if the buffer matches but not the offset...

...your change has a duplicate SOKOL_ASSERT, but I'll fix that during the merge.

floooh:
Ok merged. Many thanks!

I also removed a couple of unused pointers from the Metal state cache while at it (see: https://github.com/floooh/sokol/commit/41e7b04405b0423548d5f1a58cddb413ef8bae5a), and updated the changelog.

staminajim:
Wonderful thanks, great job :)

Issue #757, opened 2022-12-07
sometimes, on lanch an executable file could not get window focus


dazkzi:
on mac，open an executable file, in the meantime move the mouse，the new window will not get focus

this will not happen when using sdl2 or glfw

floooh:
Do you build the executable as a command line application, or as a an application bundle? I seem to remember that I have seen this behaviour with command line builds, but not with regular applications.

Also, if you already build as app bundle, what does your .plist config file look like?

dazkzi:
This only occurs on the command line executable, the bundle is fine.

GLFW and SDL2 are working fine when start the executable from command line.

floooh:
Ok, I'll try to find out what GLFW does differently on startup than sokol_app.h, it might be some attribute that's normally provided by the .plist file but needs to be set programmatically otherwise.

floooh:
Hmm, I can't seem to reproduce the issue here. I'm compiling the events-sapp.c sample as cmdline app (not as app bundle), and the window has focus right from the start (e.g. just moving the mouse reports mouse events to the app).

Can you help me reproduce the issue? For instance, what macOS and Xcode version are you on? (I'm on macOS Ventura 13.1 and Xcode 14.0.1).

dazkzi:
I found that if it takes too much time in init, this thing will happen. Such as add this empty loop in init.
```
for (int i = 0; i < 50000; i++) {
    printf("zzzzz\n");
}
```

floooh:
Hmm, good information. Still weird though because there's nothing special about the init callback, it's just called in the first frame before the frame callback. All application- and window-system setup has already happened at that time.

I also checked GLFW, and sokol_app.h is already doing the same things (but I remember that I had this sort of problem before adding the "ActivationPolicy" setting. It basically comes down to these 3 things:

https://github.com/floooh/sokol/blob/1b9b047099a249281112262259135fe699e36342/sokol_app.h#L3831-L3833

floooh:
Apparently fixed via #982, although I was unable to reproduce the original issue on my M1 Mac with macOS 14.3 (e.g. Apple might have fixed it too?).

PR #982, opened 2024-01-28
Sokol app focus steal

zoo-3d:
Hi, this pr solves the issue [sometimes, on luanch an executable file could not get window focus](https://github.com/floooh/sokol/issues/757).

it has a time interval between app launch and the app fully activated.
the time interval becomes longer when init_cb is slower.

if there are mouse events(e.g. mouse move) during the interval, then app can not be activated and focus will lost even when app launched and order front.
I read some Apple documents and found nothing about this situation.I think maybe it's a bug in Cocoa Appkit.
By the way, GTK also has this problem. And GTK solved it even though they are not 100% sure about the root cause.
see: https://gitlab.gnome.org/GNOME/gtk/-/issues/2342    

As a walkaround, we send a focus event in front of event queue.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3916,12 +3916,12 @@ _SOKOL_PRIVATE void _sapp_macos_frame(void) {
     [NSEvent setMouseCoalescingEnabled:NO];
 
     /* 
-    it has a time interval between app did launch and the window fully actived.
+    it has a time interval between app did launch and the window fully activated.
     the time interval becomes longer when init_cb is slower.
 
-    if there are mouse events(eg. mouse move) during the interval, then app can not be actived and focus will lost evan when app launched and order front.
+    if there are mouse events(eg. mouse move) during the interval, then app can not be activated and focus will lost even when app launched and order front.
     I read some Apple documents and found nothing about this situation.I think maybe it's a bug in Cocoa Appkit.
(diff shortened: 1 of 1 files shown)


floooh:
Interesting fix and thanks for the investigation! I'll try to have a closer look and do some testing within the next couple of days.

floooh:
Btw, I think I can reproduce the issue on my M1 Mac with Sonoma, but not sure if it's such a big issue. I simulated a long init callback by putting a `sleep(5)` into the init function of the clear-sapp sample.

During those 5 seconds, the window is unfocused:

Screenshot 2024-01-31 at 14 55 14

...but as soon as the first frame is rendered, the window becomes focused automatically:

Screenshot 2024-01-31 at 14 55 22

With your fix the window seems to be focused immediately (but can't be interacted with anyway).

In any case, I guess the fix doesn't do any harm either...

floooh:
...ah wait the original issue was about cmdline apps, not app bundles... I'll test that too...

floooh:
PPS: whether the window starts focused or unfocused when there's a long delay between window creation and the first random frame seems to be fairly random, but with and without the focusEvent fix. E.g. not sure if the fix is actually redundant in Sonoma, but if it helps with older macOS versions then it's fine.

Issue #983, opened 2024-01-28
sokol_app.h: Resize callback is never unregistered giving javascript errors


edubart:
## Problem
In HTML5 platform with `html5_canvas_resize` set to `false`, after `sapp_run()` finishes and you resize the Window, you will get lots of javascript errors, such as:

```
DOMException: Document.querySelector: '' is not a valid selector
    findEventTarget http://localhost:6931/test.js:1
    _emscripten_get_element_css_size http://localhost:6931/test.js:1
    uiEventHandlerFunc http://localhost:6931/test.js:1
    jsEventHandler http://localhost:6931/test.js:1
    registerOrRemoveHandler http://localhost:6931/test.js:1
    registerUiEventCallback http://localhost:6931/test.js:1
    _emscripten_set_resize_callback_on_thread http://localhost:6931/test.js:1
    _rivemu_start_record http://localhost:6931/test.js:1
    ccall http://localhost:6931/test.js:1
    rivemuRecord http://localhost:6931/test.html:155
    rivemuRecord http://localhost:6931/test.html:144
    rivemuInsertCartridge http://localhost:6931/test.html:136
    onclick http://localhost:6931/test.html:1
lockdown-install.js:1:94626
```

## Solution
Adding this line to `_sapp_emsc_unregister_eventhandlers` fixes the issue:
```c
emscripten_set_resize_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW, 0, true, 0);
```

## Use case

I am trying to make a HTML page where you can test multiple sokol apps, each one app calls `sapp_run()` and take ownership of the canvas and at some point the app ends. To have this work properly `sapp_run()` should always cleanup nicely so the next `sokol_run()` can run.

floooh:
Yep makes sense. Do you want to provide a PR?

edubart:
Just made the PR https://github.com/floooh/sokol/pull/984 to fix that.

floooh:
Thanks! PR has been merged. I'll write a small blurb in the changelog too.

PR #984, opened 2024-01-28
sokol_app.h: unregister canvas resize callback when quitting

edubart:
Fix for https://github.com/floooh/sokol/issues/983


floooh:
And merged. Thanks!

PR #987, opened 2024-02-01
Check isnan Width and Height

kassane:
based on D bindings: https://github.com/kassane/sokol-d/issues/5#issuecomment-1921244802

cc: @floooh


floooh:
Makes sense, yeah. I'll wait for the CI tests to go green (because I only see `isnan()` as C99 feature and wonder if C++ knows it too...

Do you know how those values became NaNs in the first place. If it's already mentioned in the thread just point me to the right comment :)

floooh:
> to the fact that the float/double in D is nan instead of 0.0f by default.

...wait does that mean that an unitialized float/double in D is initialized as NaN? That would be quite crazy because NaNs have a tendency to spread (they turn every expression result into NaNs too).

floooh:
Ok all green and merged. Thanks!

PR #994, opened 2024-02-22
[sokol_imgui] add  the optional backend interface for font management

Dvad:
As a happy sokol_imgui user, I am humbling proposing this PR. It splits the font texture creation in a API function and add a function for the font texture destruction. Those functions are  considered optional function of the imgui backends.

For instance they are implemented for opengl3 on the main imgui repo:

https://github.com/ocornut/imgui/blob/659fb41d0a23efbb9ea6cf74f51ecae0a51575b5/backends/imgui_impl_opengl3.h#L39C1-L42C3

I find those functions are useful in two cases:
- when testing different fonts raterization config,  it is convenient to recreate the texture dynamically. The imgui's author provides an example of such application: https://gist.github.com/ocornut/b3a9ecf13502fd818799a452969649ad
- when creating non default font, this allows to remove the boilerplate of uploading the font on GPU for the user of sokol_imgui.

Does this proposal make sense?

### DIFF
--- a/util/sokol_imgui.h
+++ b/util/sokol_imgui.h
@@ -543,6 +543,8 @@ SOKOL_IMGUI_API_DECL bool simgui_handle_event(const sapp_event* ev);
 SOKOL_IMGUI_API_DECL int simgui_map_keycode(sapp_keycode keycode);  // returns ImGuiKey_*
 #endif
 SOKOL_IMGUI_API_DECL void simgui_shutdown(void);
+SOKOL_IMGUI_API_DECL void simgui_create_fonts_texture(void);
+SOKOL_IMGUI_API_DECL void simgui_destroy_fonts_texture(void);
 
 #ifdef __cplusplus
 } // extern "C"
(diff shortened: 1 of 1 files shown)


floooh:
Good idea! But I wonder if we should make create-fonts function a bit more flexible by passing some data in.

For instance, it would be nice if I could replace this code block in the imgui-highdpi-sapp.c sample which sets up a TTF font texture with the new function:

https://github.com/floooh/sokol-samples/blob/6a03725117ce0175a45fa2d0c51ad3272c955cb5/sapp/imgui-highdpi-sapp.cc#L45-L67

...but as you can see here I'm also using a different sampler with LINEAR filtering:

https://github.com/floooh/sokol-samples/blob/6a03725117ce0175a45fa2d0c51ad3272c955cb5/sapp/imgui-highdpi-sapp.cc#L58-L63

Maybe `simgui_create_fonts_texture()` should allow to pass in optional parameters (like either a sampler object, or maybe better just a minimal set of sampler params, like min+mag filter?

Or do you solve this same problem currently in some other way? (since you are writing: "...when testing different fonts raterization config...").

Dvad:
This is good suggestion!

I have not played either the sampler in my testing so far, the tool I linked allow to switch from free type to stb_truetype with different parameter but the texture sampler was not one.

Dvad:
I gave it a shot and submitted the sister PR: 
https://github.com/floooh/sokol-samples/pull/135

floooh:
Ok, looking into this PR now.

floooh:
And merged, many thanks! (also updated the changelog)

floooh:
FYI: I'm currently porting sokol_imgui.h to Dear ImGui, and I *think* those functions are now obsolete, since Dear ImGui now has a general 'callback protocol' which backends need to implement to create/update/destroy textures managed by Dear ImGui.

PR #997, opened 2024-02-22
[sokol_app] add an option html5_use_emsc_set_main_loop

Dvad:
Implementation of the idea suggested in https://github.com/floooh/sokol/issues/843#issuecomment-1959114136

`emscripten_set_main_loop` is a higher level function compared to `emscripten_request_animation_frame_loop` and seems to play nicer with various `emcc` options (EXIT_RUNTIME or PROXY_TO_PTHREAD). The downside is that we have to compute the timestamp on our side and hence it might be less accurate than the one the browser returns with `[requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)`.

I tested both implementation on a test app. Not sure if we need more testing and how to test.

In particular for my use case, only the version `SOKOL_EMSC_USE_SET_MAIN_LOOP` works when `PROXY_TO_PTHREAD` is enabled. But A few more changes are needed for this case in sokol_app which will be for another PR.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1711,6 +1711,7 @@ typedef struct sapp_desc {
     bool html5_bubble_wheel_events;     // same for wheel events
     bool html5_bubble_key_events;       // if true, bubble up *all* key events to browser, not just key events that represent characters
     bool html5_bubble_char_events;      // if true, bubble up character events to browser
+    bool html5_use_emsc_set_main_loop;  // if true, use set_main_loop instead of request_animation_frame_loop
     bool ios_keyboard_resizes_canvas;   // if true, showing the iOS keyboard shrinks the canvas
 } sapp_desc;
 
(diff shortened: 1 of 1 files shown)


floooh:
See my comment https://github.com/floooh/sokol/issues/843#issuecomment-1960941234 about preprocessor define vs runtime flag.

floooh:
...also check the build error (I just added that flag recently):

```
/home/runner/work/sokol/sokol/tests/compile/../../sokol_app.h:5837:47: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
 5837 | _SOKOL_PRIVATE void _sapp_emsc_frame_main_loop() {
      |                                               ^
```

(the function must have an explicit void param: `void _sapp_emsc_frame_main_loop(void)`)

Dvad:
`SOKOL_EMSC_USE_SET_MAIN_LOOP` -> `_sapp.desc.html5_use_emsc_set_main_loop`

floooh:
Giving the PR a whirl now...

floooh:
Ok merged. I also added another config option `.html5_emsc_set_main_loop_simulate_infinite_loop` to be used as the `simulate_infinite_loop` param to `emscripten_set_main_loop` since this was also mentioned in https://github.com/floooh/sokol/issues/843.

...and I changed the texcube-sapp sample to use the set-main-loop method.

Also updated the changelog.

Many thanks for the PR!

Issue #995, opened 2024-02-22
[sokol_app] Status of clipboard support on X11?


Dvad:
The X11 clipboard getter is left non implemented currently :

```
SOKOL_API_IMPL const char* sapp_get_clipboard_string(void) {
    if (!_sapp.clipboard.enabled) {
        return "";
    }
    #if defined(_SAPP_MACOS)
        return _sapp_macos_get_clipboard_string();
    #elif defined(_SAPP_EMSCRIPTEN)
        return _sapp.clipboard.buffer;
    #elif defined(_SAPP_WIN32)
        return _sapp_win32_get_clipboard_string();
    #else
        /* not implemented */
        return _sapp.clipboard.buffer;
    #endif
}
```

I wonder if there is a reason to that and if you would be open to have an implementation added. If so, I am happy to submit a PR.

Thank you for this really great library!

floooh:
Hmm, as far as I can see there's currently no clipboard support at all in the sokol_app.h X11 backend. I don't know how "terrible" an implementation via Xlib would be, but if you want to give it a shot and provide a PR I won't stop you ;) (a good starting point and inspiration would be GLFW).

floooh:
I ended up merging the 'extended' PR here which is based on your PR: https://github.com/floooh/sokol/pull/1108

PR #998, opened 2024-02-23
[sokol_app] basic X11 clipboard

Dvad:
Implement clibpard feature using Xlib. This is a basic version which only supports utf8 target and the clibpoard selection. (not the selection + middle click workflow).

Simplified adaptation of glfw in https://github.com/glfw/glfw/blob/master/src/x11_window.c

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -10031,28 +10031,19 @@ _SOKOL_PRIVATE void _sapp_x11_lock_mouse(bool lock) {
     XFlush(_sapp.x11.display);
 }
 
-_SOKOL_PRIVATE Bool _sa_sapp_x11_is_sel_prop_new_value_notify(Display* display, XEvent* event, XPointer pointer)
-{
-    XEvent* notification = (XEvent*) pointer;
-    return event->type == PropertyNotify &&
-           event->xproperty.state == PropertyNewValue &&
-           event->xproperty.window == notification->xselection.requestor &&
-           event->xproperty.atom == notification->xselection.property;
-}
-
 _SOKOL_PRIVATE void _sapp_x11_set_clipboard_string(const char* str) {
     SOKOL_ASSERT(_sapp.clipboard.enabled && _sapp.clipboard.buffer);
     size_t len = strlen(str) + 1;
-    if (len > _sapp.clipboard.buf_size) {
+    if (len > (size_t)_sapp.clipboard.buf_size) {
         _SAPP_ERROR(CLIPBOARD_STRING_TOO_BIG);
     }
     XSetSelectionOwner(_sapp.x11.display, _sapp.x11.CLIPBOARD, _sapp.x11.window, CurrentTime);
-    if (XGetSelectionOwner(_sapp.x11.display, _sapp.x11.CLIPBOARD) != _sapp.x11.window){
+    if (XGetSelectionOwner(_sapp.x11.display, _sapp.x11.CLIPBOARD) != _sapp.x11.window) {
         _SAPP_ERROR(LINUX_X11_FAILED_TO_BECOME_OWNER_OF_CLIPBOARD);
     }
 }
 
-_SOKOL_PRIVATE const char* _sa_sapp_x11_get_clipboard_string() {
+_SOKOL_PRIVATE const char* _sa_sapp_x11_get_clipboard_string(void) {
     // Note: This function is a simplified version of implementation in glfw
     SOKOL_ASSERT(_sapp.clipboard.enabled && _sapp.clipboard.buffer);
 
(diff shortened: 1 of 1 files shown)


Dvad:
@floooh here is my version regarding https://github.com/floooh/sokol/issues/995.

By the way, is there a code style or code formatter I should use on the code when submitting PR?

floooh:
That looks reasonably simple :) Somehow and suddenly there's a wave of PRs coming in, it might be a couple of days before I have time to properly look at the PR.

In the meantime, have a look at the failed CI build though please:

https://github.com/floooh/sokol/actions/runs/8015098805/job/21903872711?pr=998

Unused variable warnings should be supressed with `_SOKOL_UNUSED(var)`, and C functions without args must have an explicit void arg (e.g. `const char* _sa_sapp_x11_get_clipboard_string(void)`).

(the CI builds have extremely picky warning settings)

sokol_app.h
```diff
@@ -11362,7 +11483,7 @@ SOKOL_API_IMPL void sapp_set_clipboard_string(const char* str) {
     #elif defined(_SAPP_WIN32)
         _sapp_win32_set_clipboard_string(str);
     #else
-        /* not implemented */
+        _sapp_x11_set_clipboard_string(str);
```
floooh, line 11486: Here's why the Android and iOS builds also fail: the `_sapp_x11_set_clipboard_string` needs to go into its own `#elif defined(_SAPP_LINUX)` block, and the `#else` block needs to remain empty.

sokol_app.h
```diff
@@ -11378,8 +11499,7 @@ SOKOL_API_IMPL const char* sapp_get_clipboard_string(void) {
     #elif defined(_SAPP_WIN32)
         return _sapp_win32_get_clipboard_string();
     #else
-        /* not implemented */
-        return _sapp.clipboard.buffer;
+        return _sa_sapp_x11_get_clipboard_string();
```
floooh, line 11502: ...ditto here (`#elif defined(_SAPP_LINUX)` and the generic `#else` needs to remain as before.

floooh:
PPS: 

> By the way, is there a code style or code formatter I should use on the code when submitting PR?

I don't have clang-format setup yet, you should mainly check that the opening `{` doesn't go into its own line, and the else branches look like this `} else {` (the latter I only changed a little while ago, so there may still be some places with the old else-style.

But please don't run a formatter over the code, that tends to bury the actual changes.

Dvad:
Updated. Hopefully all failure are gone.

> Somehow and suddenly there's a wave of PRs coming in, it might be a couple of days before I have time to properly look at the PR.

All good, thanks in the first place for considering them.

floooh:
FYI I'll start testing this PR next, but not sure yet if I'll get around to merge today or tomorrow.

floooh:
Hmm, when testing, pasting from the system clipboard *into* the sokol application works, but not pasting out of it into the system clipboard (e.g. `sapp_set_clipboard_string()` seems to be a no-op.

You can test this by running `imgui-sapp`, go to `Widgets => Text Input => Multi-line text input` mark some text, press Ctrl-X or Ctrl-C. Then go into a regular desktop application and try to paste there (on KDE, there's also a Klipper app which lets you inspect the clipboard content). The cut/copied text from the sokol-app doesn't make it into the system clipboard (the function _sapp_x11_set_clipboard_string is being called though).

Looking at the GLFW code just leaves me confused TBH :D We should try to get this fixed though before 

image

(I tested on an Ubuntu 23.10 laptop with KDE desktop session)

Dvad:
Thanks for testing!

Hum, I need to investigate, both direction are definitely working on my machine. (Fedora 38, Gnome running on X11).
Yeah GLFW is a bit confusing, I simplified a bit by cutting the select + middle-click flow. I wonder if I have simplified too much and dropped support for some cases without noticing.

floooh:
Interesting that it works on your side, can you describe step by step how you cut/copy out of an sokol-app application into the system? Maybe I messed something up or encountered an edge case.

Dvad:
Ctrl+X/Ctrl+C from the `Multi-line text input` then Ctrl+V my terminal worked (I use tilix).
Since Xorg clipboard is a peer to peer protocol, I was thinking maybe my code does not play nice with what the clipboard manager do? Or it could be a Wayland/Mir vs X11 thing?

Dvad:
BTW, I did not abandon this PR, but I am lacking time to test on a ubuntu machine to debug currently. My attempt to use a virtual machine was not successfull so far.

qwx9:
Not sure what the best form is on github, I opened a new PR to submit an amended patch. Thanks for your work!

floooh:
@Dvad I'm closing this PR in favour of https://github.com/floooh/sokol/pull/1108. Unfortunately you didn't make it into the git history, I wonder if we could have handled that better. I *did* mention you in the changelog though.

PR #1001, opened 2024-02-26
skip framebuffer invalidation when possible

danielchasehooper:
sokol was invalidating the depth/stencil attachment when there was none attached. Because of this, we were seeing webGL's gl.InvalidateFramebuffer in performance profiles when the call wasn't needed at all.

This PR only clears the depth or stencil attachments if the pass has a depth/stencil attachment.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -8679,10 +8679,10 @@ _SOKOL_PRIVATE void _sg_gl_end_pass(void) {
                 invalidate_atts[att_index++] = (GLenum)(GL_COLOR_ATTACHMENT0 + i);
             }
         }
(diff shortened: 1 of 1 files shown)


floooh:
Good fix, thanks! Interesting that a function like glInvalidateFramebuffer() shows up in performance profiling, I would expect that this is more or less a no-op.

I'll replace the `!= 0` with `SG_INVALID_ID` though and then merge.

floooh:
Ok, I did this minor code style fixed and merged: https://github.com/floooh/sokol/commit/d7110f6b9e2b4f8b460943b9f1a3d9e2738758c0

Thanks!

PR #1007, opened 2024-03-09
Initial (readonly) storage buffer support.

floooh:
First step will introduce storage buffers without compute shaders. This allows more flexibility for providing random-access data to the GPU (e.g. replacing float textures as 'poor man's storage buffers').

- Not supported on WebGL2 and on the macOS GL backend (since macOS GL only goes up to 4.1)
- Will require to bump the minimum GL version to 4.3 on Linux and Windows.
- Unrelated: while at it, bump minimum GL version to 4.1 on macOS (will enable to fill some feature gaps later)
- ??? GLES3.1 on Anrdoid and iOS ??? (not on iOS since it only supports GLES 3.0, not initially on Android)
- on D3D11 uses HLSL ByteAddressBuffers as opposed to StructuredBuffers, this is more flexible but more awkward when used with manually written HLSL (convenient again with sokol-shdc though)
- only 

To investigate:

- ~~Allow multiple bind points? (e.g. allow to bind as storage buffer, but also as vertex buffer) => means getting rid of 'buffer type' in favour of 'bind flags'~~ => not initially
- ~~D3D11: when is using an UnorderedAccessView instead of a regular BufferView required?~~
- ReadOnly ~~vs ReadWrite vs WriteOnly storage buffers? Possible to write to StorageBuffers without compute shaders?~~ => possible but non-trivial, initially only support readonly buffers, writing from vertex- oder fragment-shaders requires resource barriers even in OpenGL. Delay until compute shader support.
- ~~Add a special flag `sg_buffer_desc.d3d11_structured_buffer` as fallback to allow using more convenient StructureBuffer in shader? Does this require providing the element size when creating the shader resource view?~~
- ???

TODO:
- [x] fix broken shadows-depthtex sample
- [x] update samples webpage
- [x] query GL version for glGetIntegerv() and roll back the changes in `sg_environment`.
- [x] pixel noise artifacts in the `sbuftex-sapp` on my NVIDIA RTX 2070 Win10 PC across WebGPU, D3D11 and OpenGL (fixed, turned out to be a precision problem when converting between float and integer which only seems to manifest on NVIDIA)
- [ ] final smoke test:
    - [x] sapp-samples on:
        - [x] macOS + Metal
        - [x] macOS + GL
        - [x] iOS + Metal
        - [x] iOS + GL
        - [x] Windows + D3D11
        - [x] Windows + GL
        - [x] WebGL2
        - [x] WebGPU
        - [x] Linux
        - [x] Android
    - [x] backend-specific samples:
        - [x] d3d11 samples
        - [x] metal samples
        - [x] emsc samples
        - [x] wgpu samples
        - [x] glfw samples
            - [x] macOS
            - [x] Windows
            - [x] Linux
    - [x] bindings:
        - [x] zig
        - [x] odin
        - [x] nim
        - [x] rust
- [ ] deploy samples webpage
    - [x] WebGPU
    - [ ] WebGL2
- [x] blog post
- [x] update changelog
- [ ] merge:
    - [ ] https://github.com/floooh/sokol-samples/pull/140
    - [x] https://github.com/floooh/sokol-tools/pull/120
    - [x] https://github.com/floooh/sokol-zig/pull/61
    - [x] https://github.com/floooh/sokol-odin/pull/16
    - [x] https://github.com/floooh/sokol-nim/pull/29
    - [x] https://github.com/floooh/sokol-rust/pull/29
- [ ] update:
    - [ ] blogpost links
    - [ ] https://github.com/floooh/chips-test
    - [ ] https://github.com/floooh/v6502r
    - [ ] https://github.com/floooh/pacman.c
    - [ ] https://github.com/floooh/doom-sokol
    - [ ] https://github.com/floooh/qoiview
    - [ ] https://github.com/floooh/cimgui-sokol-starterkit
    - [ ] https://github.com/floooh/kc85.zig
    - [ ] https://github.com/floooh/pacman.zig
- [x] publish sokol-tools-bin
- [x] add 'unsupported' debug text to storage buffer samples
- [x] documentation update:
    - [x] sokol_gfx.h
    - [x] sokol_app.h
- [x] fix sokol-shdc + webgpu problems
- [x] only allow readonly storage buffers in shader desc (validation layer check)
- [x] ~~cleanup `SG_BUFFERTYPE_*` naming? (e.g. `SG_BUFFERTYPE_VERTEXBUFFER` => `SG_BUFFERTYPE_VERTEX`)?~~ => not worth the hassle
- [x] rename config define `SOKOL_GLCORE33` => `SOKOL_GLCORE`
- [x] pass major and minor GL version in `sg_environment.gl`
- [x] basic storage buffer support in sokol_gfx.h:
    - [x] glcore
    - [x] gles3? (unsupported)
    - [x] d3d11
    - [x] metal
    - [x] wgpu
- [x] 'raw' 3D API samples:
    - [x] vertex pulling via storage buffers
        - [x] d3d11
        - [x] metal
        - [x] wgpu
        - [x] glfw
- [x] sokol-app samples:
    - [x] simple vertex pulling sample
    - [x] new instancing sample which uses vertex pulling
    - [x] new ozz-skinning sample which passes bone info via storage-buffer
    - [x] a sample which uses matrices in storage buffers
    - [x] a sample which read storage buffer from fragment shader - use the uniformtypes-sapp as base, and also bind a texture to the same stage (to check if bind slots are assigned correctly):
        - [x] macOS + Metal
        - [x] WebGPU (currently broken: Binding size (768) is larger than the size (96) of [Buffer "color-palette"].)
        - [x] Linux + GL
        - [x] Windows + D3D11
        - [x] Windows + GL
    - [x] ~~shader-write to storage buffer~~
- [x] sokol_app.h + sokol_glue.h updates:
    - [x] sg_environment => gl major/minor version
- [x] fix sokol_gfx_imgui.h
    - [x] buffer panel => storage buffer
    - [x] update shader-desc panel
    - [x] update apply-bindings-panel
- [x] recompile embedded WebGPU and GLSL shaders (because WGPU bind slots and GLSL versions have changed):
    - [x] sokol_imgui.h
    - [x] sokol_gl.h
    - [x] sokol_nuklear.h
    - [x] sokol_debugtext.h
    - [x] sokol_spine.h
    - [x] sokol_fontstash.h
- [x] update bindings (plus a new vertex pulling sample):
    - [x] zig
    - [x] odin
    - [x] nim
    - [x] rust
- [x] sokol-shdc update:
    - [x] glsl330 => glsl410 and glsl430
    - [x] remove glsl100
    - [x] cleanup reflection parsing (also need to parse regular structs)
    - [x] reflection parsing: SSBOs
    - [x] 'skip' the internal struct for code-generated ssbo struct
    - [x] SPIRVCross writes gl_InstanceIndex as `gl_InstanceID + SPIRV_Cross_BaseInstance` (fixed: SPIRVCross has options for that)
 - [x] ~~figure out how to deal with write/read barriers for storage buffers, e.g. https://registry.khronos.org/OpenGL-Refpages/gl4/html/glMemoryBarrier.xhtml, conservative solution: issue a barrier after each draw that involves writable storage buffers.~~ => defer storage buffer write support into a followup update since it comes with a couple of problems that need solving:
     - barriers must be injected between a draw call writing to a storage buffer and another draw call reading from that buffer, so need to come with a basic concept of how to handle such resource barriers (make a backend API survey first)
     - SPIRVCross seems to generate no return value for (vertex) shaders which write to storage buffers
     - might want to handle the following situations properly:
         - shaders that only have a vertex- but no fragment-shaders
         - render passes without any attachments (and essentially only serve as barriers => do we want to have something like a "barrier pass"?


jakubtomsu:
Oh, are you planning to eventually support compute shaders, on selected backends? That would be absolutely amazing, it's probably the number one feature I'm missing in sokol_gfx right now. Especially now I've moved on to 3D graphics. The storage buffers seem like a good first step!

floooh:
Yes, compute shaders is planned afterward for the next 'big update' after this initial storage buffer support update. I need to come up with a 'resource barrier strategy' though, I'll most likely take WebGPU as a blueprint.

fiserj:
This is great milestone, thank you! (I can finally stop abusing uniforms for that 🙈)

roig:
Amazing work, thank you for this update.

PR #1013, opened 2024-03-20
Fix a couple of typos.

waywardmonkeys:


floooh:
Thanks again! (man, I had a hunch that 'environment' and 'attachment' aren't great names, because I made that typo mistake a couple of times during development)

PR #1016, opened 2024-03-26
Add logo assets

jakubtomsu:
I recently noticed the default sokol "S" icon isn't available anywhere and it's only generated in [code](https://github.com/floooh/sokol/blob/3853a86b50a9b5bf890171d1bd830c73f9e699e1/sokol_app.h#L3252). That's fine, but I wanted to use it to credit the project. So I made an 1:1 remake of the logo, both the "S" and full name, with the original colors and the Amstrad CPC font.
I'm sure this could be useful to other people as well, in case they want to credit the library.

I also added the logo to the top of the README, so it's a bit more flashy that way. In case you don't like the look let me know.

By the way, this is just something I thought would make a good first contribution, but I plan to contribute actual code too :)


floooh:
Heh, good idea. TBH I didn't put much thought into the default icon, but it kinda stuck and I like it now ;)

Thanks for the PR!

PR #1020, opened 2024-04-07
Delete the code that always force debug text alpha to zero.

kochol:
Now, the debug text can be properly blended with other renderings during offscreen rendering.


floooh:
Oops, the actual bug is that the src and dst alpha factors are flipped, I'll fix that in a commit (just removing the two lines to use the defaults is a bit confusing IMHO).

floooh:
Ok, I committed this change instead:

https://github.com/floooh/sokol/commit/6747384927e4520bd3391bba81bf416567655239

Many thanks for the PR, and apologies for closing instead of merging.

PR #1021, opened 2024-04-10
Update Odin to latest build with LLVM17 and enable MacOS again

jakubtomsu:
This is a port of https://github.com/floooh/sokol-odin/pull/10


floooh:
Awesome! Thanks again!

PR #1023, opened 2024-04-11
Improve odin bindings, DLL support

jakubtomsu:
This PR does a number of modifications to the Odin binding generator, and the project structure:
- Add support for DLLs, using unity builds to compile all sokol libs into a single DLL (Windows only)
  - the primary reason for this is hot reloading
  - added `sokol.c` for unity builds
- Define the `#config` values as constants, so they can be easily used elsewhere
- Use a custom debug constant instead of `ODIN_DEBUG` (which is still used by default), which can override either all sokol libs or individual modules
- Add some empty lines to the generated bindings to improve readability :) (this is my personal preference, doesn't have anything to do with the rest of this PR)
- move the `c` folder which contains all C files and the build scripts into the `sokol` folder with odin packages. This means the `sokol` folder is now completely standalone and can be easily copied into the `source` directory of any project.

Most of these changes are what I used in my game Solar Storm. Sometimes it's just what I personally prefer, but I think it ultimately does make the bindings easier to work with.

> Note: these changes are still a bit experimental as I couldn't test 100% everything before submitting the PR (like the automatic uploading of bindings on each commit).


jakubtomsu:
Linked PR from the sokol-odin repo: https://github.com/floooh/sokol-odin/pull/11

jakubtomsu:
It's a pretty big PR, so in case you don't like something let me know

floooh:
Looks good! I'll try to give it a whirl on the weekend before merging.

I was wondering why you moved the top-level scripts to build the C libraries into a subdirectory (because I preferred to have them visible at the top level), but this is a good reason:

> This means the sokol folder is now completely standalone and can be easily copied into the source directory of any project.

jakubtomsu:
Thank you! I agree the build scripts look nicer in the top-level folder, but this way just fits the odin workflow a bit better...

floooh:
And merged. Many thanks!

PR #1025, opened 2024-04-11
[sokol_gfx] Assign labels as debug names to D3D11 resources

jakubtomsu:
This change assigns the "label" field form resource desc structs to D3D11 resource objects, using 
`SetPrivateData` with `WKPDID_D3DDebugObjectName`. This code is enabled only when `SOKOL_DEBUG` is defined.

Note: this requires sokol_gfx to also link with `dxguid.lib`.

Note: some resources have multiple "sub-resources", e.g. a shader has a vertex and a fragment shader resource. In those cases the label is the same for all sub-resources.

This is useful especially in RenderDoc ([docs](https://renderdoc.org/docs/how/how_annotate_capture.html#:~:text=tex2d%2D%3ESetPrivateData(WKPDID_D3DDebugObjectName%2C%20sizeof(%22Example%20Texture%22)%2C%20%22Example%20Texture%22)%3B)) and in D3D11 debugging messages.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -4295,6 +4295,9 @@ inline int sg_append_buffer(sg_buffer buf_id, const sg_range& data) { return sg_
     #pragma comment (lib, "user32")
     #pragma comment (lib, "dxgi")
     #pragma comment (lib, "d3d11")
+    #if defined(SOKOL_DEBUG)
+    #pragma comment (lib, "dxguid") // for debug-only D3DDebugObjectNames
+    #endif
     #endif
 #elif defined(SOKOL_METAL)
     // see https://clang.llvm.org/docs/LanguageExtensions.html#automatic-reference-counting
@@ -9222,6 +9225,12 @@ _SOKOL_PRIVATE void _sg_gl_update_image(_sg_image_t* img, const sg_image_data* d
 #define _sg_d3d11_Release(self) (self)->lpVtbl->Release(self)
 #endif
 
+#if defined(__cplusplus)
+#define _sg_d3d11_SetPrivateData(self, guid, DataSize, pData) (self)->SetPrivateData(guid, DataSize, pData)
+#else
+#define _sg_d3d11_SetPrivateData(self, guid, DataSize, pData) (self)->lpVtbl->SetPrivateData(self, &guid, DataSize, pData)
+#endif
+
 //-- D3D11 C/C++ wrappers ------------------------------------------------------
 static inline HRESULT _sg_d3d11_CheckFormatSupport(ID3D11Device* self, DXGI_FORMAT Format, UINT* pFormatSupport) {
     #if defined(__cplusplus)
@@ -9991,6 +10000,9 @@ _SOKOL_PRIVATE sg_resource_state _sg_d3d11_create_buffer(_sg_buffer_t* buf, cons
             _SG_ERROR(D3D11_CREATE_BUFFER_FAILED);
             return SG_RESOURCESTATE_FAILED;
         }
+        #if defined(SOKOL_DEBUG)
+        _sg_d3d11_SetPrivateData(buf->d3d11.buf, WKPDID_D3DDebugObjectName, (UINT)strlen(desc->label), desc->label);
+        #endif
     }
     return SG_RESOURCESTATE_VALID;
 }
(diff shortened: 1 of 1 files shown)


floooh:
Good change, but I want to give it a whirl before merging.

We should try a change though which I think/hope allows us to get rid of the dxguid dependency:

In sokol_app.h I'm defining any required COM IIDs inline like this:

https://github.com/floooh/sokol/blob/91ec4b5fa95b1e3e3b2e21cfeac85d03cdc9a8ef/sokol_app.h#L6317-L6319

Maybe we can do the same inline definition for the `WKPDID_D3DDebugObjectName` ID.

I seem to remember that linking with dxguid caused problems in some scenarios (like compiling via Zig which uses the mingw2 headers instead of the Windows SDK).

Also another change request regarding this block:

```
#if defined(__cplusplus)
#define _sg_d3d11_SetPrivateData(self, guid, DataSize, pData) (self)->SetPrivateData(guid, DataSize, pData)
#else
#define _sg_d3d11_SetPrivateData(self, guid, DataSize, pData) (self)->lpVtbl->SetPrivateData(self, &guid, DataSize, pData)
#endif
```

Please create a wrapper function like this instead (all D3D API functions called by sokol-gfx have such a wrapper:

https://github.com/floooh/sokol/blob/91ec4b5fa95b1e3e3b2e21cfeac85d03cdc9a8ef/sokol_gfx.h#L9354-L9360

That's all for now :)

floooh:
PS: did you also check that calling `SetPrivateData` with a nullptr works?

jakubtomsu:
Oh yeah, defining the GUIIDs inline is a very good idea, I didn't notice the ones that were already there, that's why I linked dxguid.

> Please create a wrapper function like this instead

The reason why I didn't create a function like this is because `SetPrivateData` is called on many different types, all child of `ID3D11Device`. And while the polymorphism works in C++, the C version needs to explicitly use the VTable. So I made a macro similar to `_sg_d3d11_Release` or `_sg_d3d11_AddRef`. I'm not aware of any way to make this work in C, apart from writing the function for each resource type. Should I do that instead?

> did you also check that calling SetPrivateData with a nullptr works in case no label is defined?

Oops, good catch! Somehow I completely forgot about that, since I always set the labels in my own code. But looking at the [docs](https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicechild-setprivatedata#:~:text=Pointer%20to%20the%20data%20to%20be%20stored%20with%20this%20device%20child.%20If%20pData%20is%20NULL%2C%20DataSize%20must%20also%20be%200%2C%20and%20any%20data%20previously%20associated%20with%20the%20specified%20guid%20will%20be%20destroyed.) it should be fine.

floooh:
> The reason why I didn't create a function like this is because SetPrivateData is called on many different types

Ah right, in that case the macro totally makes sense. But can you add a little explanation comment above the macro? Because such things tend to come up after a while again ;)

>But looking at the docs it should be fine.

Agreed, looks like nullptr is valid.

jakubtomsu:
Alright, I added the GUID constant and comments in the last commit

floooh:
Last changes look good.

I just noticed a problem though I overlooked at first, those strlen calls on nullptrs:

```c
(UINT)strlen(desc->label)
```

..AFAIK strlen on a nullptr is UB.

My suggestion would be to change the ` _sg_d3d11_SetPrivateData` into something simpler and remove any redundancies (for the C++ version):

```cpp
_sg_d3d11_setlabel(self, label) (self)->SetPrivateData(_sg_win32_refguid(_sg_d3d11_WKPDID_D3DDebugObjectName), label ? (UINT)strlen(label) : 0, label)
```

...calling this would then look a bit simpler:

```c
_sg_d3d11_setlabel(buf->d3d11.buf, desc->label);
```

jakubtomsu:
Yeah the setlabel macro is a very good idea, it makes it quite a lot simpler. I didn't know about the strlen UB, thanks for letting me know.

floooh:
Giving the PR a whirl now...

floooh:
Nice, it also works in the Visual Studio graphics debugger:

image

floooh:
And merged. I also added a blurb to the changelog.

PR #1026, opened 2024-04-11
[sokol_gfx] Increase _SG_STRING_SIZE from 16 to 32

jakubtomsu:
I'm not sure why the static string size limit was so low, since it's used for vertex attribute names. I ran into issues where a regular variable name of a shader vertex vertex/instance attribute would trigger this limit.

So I changed it to 32 which seems like a more reasonable limit for variable names...


floooh:
Yes agreed, 32 is a more reasonable size.

PR #1029, opened 2024-04-12
[odin bindings] Exclude 'NUM' enum values

jakubtomsu:
There is no reason to have a `NUM` field at the end of an enum in Odin. Because unlike in C, Odin supports `len(My_Enum)`. And the additional enum field also interferes with things like enumerated arrays and bit_sets.


floooh:
Thanks! I'll try looking at your other PRs later today.

jakubtomsu:
Thank you! The other PRs I've submitted were mostly things I ran into while working on my game. But I never got around to submitting them upstream, so I did that now all at once :)

PR #1045, opened 2024-05-11
Mark WebGPU 32-bit float texture formats as "filterable" if device supports it

jdah:
Marks R32F, RG32F, and RGBA32F as filterable on wgpu backend by checking feature flag (according to the table here https://gpuweb.github.io/gpuweb/#texture-format-caps). Seems to be OK-widely supported as well (https://web3dsurvey.com/webgpu/features/float32-filterable), though not sure if that's a concern.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -6483,6 +6483,12 @@ _SOKOL_PRIVATE void _sg_pixelformat_sr(_sg_pixelformat_info_t* pfi) {
     pfi->render = true;
 }
 
+_SOKOL_PRIVATE void _sg_pixelformat_sfr(_sg_pixelformat_info_t* pfi) {
+    pfi->sample = true;
+    pfi->filter = true;
+    pfi->render = true;
+}
+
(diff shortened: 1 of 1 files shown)


floooh:
Thanks for the PR, I'll give it a whirl now. Do you have already code working with the change? I wonder if I have any accidential hardwired code paths in the validation layer which assume that float textures are not filterable (but I'll try to find out while testing).

floooh:
Ok, I need to do a couple of changes in sokol_app.h (I'll do this in my intermediate merge-branch).

Basically extending this code block with optional features:

https://github.com/floooh/sokol/blob/67339198b75a7d04da3676db2b56514bf363d4b1/sokol_app.h#L5771-L5777

...while at it I also fixed the BC vs ETC2 detection (both are available on Apple Silicon), and the missing ASTC check.

I'll do a couple of filterable-float checks next and then merge.

floooh:
Ok merged, many thanks! As I wrote above, I also fixed the feature detection in sokol_app.h, and update the changelog.

PR #1048, opened 2024-05-14
Changes to support generating the sokol-fetch bindings in zig

trace-andreason:
Related to this issue: https://github.com/floooh/sokol-zig/issues/63. Not ready to be merged, just thought it might be helpful for that discussion

### DIFF
--- a/bindgen/gen_all.py
+++ b/bindgen/gen_all.py
@@ -7,7 +7,6 @@ tasks = [
     [ '../sokol_glue.h',           'sglue_',    ['sg_'] ],
     [ '../sokol_time.h',           'stm_',      [] ],
     [ '../sokol_audio.h',          'saudio_',   [] ],
-    [ '../sokol_fetch.h',          'sfetch_',     [] ],
     [ '../util/sokol_gl.h',        'sgl_',      ['sg_'] ],
     [ '../util/sokol_debugtext.h', 'sdtx_',     ['sg_'] ],
     [ '../util/sokol_shape.h',     'sshape_',   ['sg_'] ],
@@ -26,8 +25,12 @@ for task in tasks:
     gen_nim.gen(c_header_path, main_prefix, dep_prefixes)
 
 # Zig
+zig_tasks = [
+    *tasks,
+    [ '../sokol_fetch.h', 'sfetch_', [] ],
+]
 gen_zig.prepare()
-for task in tasks:
+for task in zig_tasks:
     [c_header_path, main_prefix, dep_prefixes] = task
     gen_zig.gen(c_header_path, main_prefix, dep_prefixes)
 
--- a/bindgen/gen_all.py
+++ b/bindgen/gen_all.py
@@ -7,6 +7,7 @@ tasks = [
     [ '../sokol_glue.h',           'sglue_',    ['sg_'] ],
     [ '../sokol_time.h',           'stm_',      [] ],
     [ '../sokol_audio.h',          'saudio_',   [] ],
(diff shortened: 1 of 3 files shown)


floooh:
...that fetch callback should actually be handled here:

https://github.com/floooh/sokol/blob/f70cc07f0fcd763ab6ff0171a6b8fe1dca3ed367/bindgen/gen_zig.py#L339-L340

...I'm currently investigating why it isn't...

floooh:
Aha, because it's a typedef in the sokol_fetch.h header, and that's not detected in the bindings generation script.

If I remove the `sfetch_callback_t` typedef from sokol_fetch.h and replace the callback declaration in `sfetch_request_t` to this:

```c
typedef struct sfetch_request_t {
    ...
    void (*callback) (const sfetch_response_t*);
    ...
} sfetch_request_t;
```

...it works, and generates the following Zig struct:

```zig
pub const Request = extern struct {
    channel: u32 = 0,
    path: [*c]const u8 = null,
    callback: ?*const fn ([*c]const Response) callconv(.C) void = null,
    chunk_size: u32 = 0,
    buffer: Range = .{},
    user_data: Range = .{},
};
```

...can you integrate this sokol_fetch.h change in your PR and remove your "manual exception hack"?

trace-andreason:
made those changes

floooh:
...for the `sfetch_desc` override I would actually prefer `sfetch_get_desc` which would then end up as `fetch.getDesc()` on the Zig side.

For `continue` maybe `resume`? Since it's "resuming from paused state".

...tbh, `sfetch_resume()` would also be a good change in the C header.

floooh:
...one other thing I currently wonder about is whether we should restrict sokol_fetch.h to the Zig bindings for now, e.g. remove the sokol_fetch item from the tasks array in gen_all.py, and instead do this down in the gen_zig block:

```python
zig_tasks = [
    *tasks,
    [ '../util/sokol_fetch.h', 'sfetch_', [] ],
]
for task in zig_tasks:
   ...
```

trace-andreason:
turns out resume is also reserved

trace-andreason:
I went ahead and named it `continue_fetching`, but open to any other suggestion. All other comments are incorporated.

floooh:
Yeah `continue_fetching` is fine :)

trace-andreason:
bindings build step is going to fail until `sokol_fetch.c` file is added to the repo

floooh:
Yep, I think I can take over from here. I'll also need to do a quick check if the dropped typedef breaks anything (I don't think so, but technically it's a breaking change). I'll try to look into it later today.

floooh:
Zig master might have had a breaking change in LazyPath. I'm looking into it.

floooh:
Ok, took a bit longer then I had hoped, but the sokol-zig build.zig is now fixed for the current zig head version.

floooh:
I'll go ahead and merge this PR.

The next step would then be in sokol-zig again to add the C source file to build.zig here:

https://github.com/floooh/sokol-zig/blob/7d569775df73d1e869a1df12dfc647e56c7cb77a/build.zig#L265

...and the fetch.zig module here:

https://github.com/floooh/sokol-zig/blob/7d569775df73d1e869a1df12dfc647e56c7cb77a/src/sokol/sokol.zig#L9

Ideally we would also have a new sample, but I can do that at a later time.

floooh:
...also added a little blurb to the CHANGELOG about the theoretically breaking change in sokol_fetch.h

PR #1055, opened 2024-05-24
Fix include to work on FreeBSD

kimshrier:
You do not need to include alloca.h on FreeBSD in order to have the prototype for the alloca function.  It is defined in stdlib.h.


floooh:
Oki doki thanks! Note though that I cannot guarantee BDS compatibility at all times because I only test changes on a vanilla Linux distro (Ubuntu).

Issue #1071, opened 2024-07-02
sokol_gfx.h: wrong define in log related to SOKOL_TRACE_HOOKS 


AlfayezDev:
https://github.com/floooh/sokol/blob/b1221d1f2bfc3cbb22f566cfa8f31b4073c287ae/sokol_gfx.h#L3595C1-L3596C1

Log states
```
_SG_LOGITEM_XMACRO(TRACE_HOOKS_NOT_ENABLED, "sg_install_trace_hooks() called, but SG_TRACE_HOOKS is not defined")
```
when it should be
```
_SG_LOGITEM_XMACRO(TRACE_HOOKS_NOT_ENABLED, "sg_install_trace_hooks() called, but SOKOL_TRACE_HOOKS is not defined")
```

PR #1072, opened 2024-07-02
logs/ oudated SOKOL_TRACE_HOOKS define

AlfayezDev:
fixes
https://github.com/floooh/sokol/issues/1071


floooh:
Oops yeah good catch. Thanks!

PR #1077, opened 2024-07-09
Add extra right shift scancode.

skejeton:
This unusual scancode (`0x136`) appears on my keyboard as a scancode for right shift, instead of the normal `0x36`. 

From MSDN: 

> Indicates whether the key is an extended key, such as the right-hand ALT and CTRL keys that appear on an enhanced 101- or 102-key keyboard. The value is 1 if it is an extended key; otherwise, it is 0.

Even though scancode for right shift already exists.


floooh:
...and merged. Thanks!

PR #1086, opened 2024-07-28
Added emscripten support for the Nim's gen script

Nazariglez:
Related to https://github.com/floooh/sokol-nim/pull/31


floooh:
Thanks for the PRs! I'm a bit busy atm with this Chrome WebGL2 regressions thingy (there goes the weekend lol), but I'll try to focus on those PRs some time next week.

floooh:
NOTE: that CI tests are expected to break because they don't have the changes in sokol-nim, don't worry about that.

Nazariglez:
There is nor rush at all, thanks @floooh!

floooh:
...checking now whether the CI runs through with the changes in sokol-nim merged, this will take a little bit...

floooh:
...ok actually got a real error in the CI pipeline now:

```
sokol/gfx.nim(2032, 5) Error: expected: ':', but got: 'keyword when'
```

floooh:
There's this difference between the generated version and sokol-nim:

```
diff --git a/src/sokol/gfx.nim b/src/sokol/gfx.nim
index bc8f184..b4758a0 100644
--- a/src/sokol/gfx.nim
+++ b/src/sokol/gfx.nim
@@ -2006,7 +2006,7 @@ when defined emscripten:
   const d3d11* = false
   const metal* = false
   const emscripten* = true
-elif defined gl:
+elif when defined gl:
   const gl*    = true
   const d3d11* = false
   const metal* = false
```

I guess that `when` should be there, I'll try to fix that locally.

floooh:
Ok, I did this minor fix and then merged: https://github.com/floooh/sokol/commit/b073c61c6e1d27d3e8e40ad83ebca3bad62a9b01

I didn't actually try to build for Emscripten, only for native macOS.

I'll add a little blurb to the changelog next.

Thanks for the PR!

PR #1095, opened 2024-08-16
Fix Jai bindings for Metal on MacOS

colinbellino:
Added missing `,link_always` when using the Metal backend on MacOS, which caused linker issues on some versions of MacOS.

This should fix the following linker error when using Metal:
```console
lld-macos: error: undefined symbol: ___isPlatformVersionAtLeast
```


floooh:
Thanks! Bindings should be automatically updated: https://github.com/floooh/sokol/actions/runs/10422193201

PR #1107, opened 2024-09-10
Fixes a missing wgl destruction (SOKOL_NOAPI)

pplux:
When we use SOKOL_NOAPI the cleanup of the window still was trying to destroy an opengl context.


floooh:
And merged. Thanks!

PR #1108, opened 2024-09-12
sokol_app: basic x11 clipboard support

qwx9:
Code based on PR #998 and https://github.com/X11-good-tools/x11clipboard.
Handling TARGETS request fixes _sapp_x11_set_clipboard_string() not working on some setups.
Additional minor formatting changes.

Not thrilled about polling or about how much code this all adds, but I don't really see a simpler way of doing this.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1603,6 +1603,7 @@ typedef struct sapp_allocator {
     _SAPP_LOGITEM_XMACRO(LINUX_X11_OPEN_DISPLAY_FAILED, "XOpenDisplay() failed") \
     _SAPP_LOGITEM_XMACRO(LINUX_X11_QUERY_SYSTEM_DPI_FAILED, "failed to query system dpi value, assuming default 96.0") \
     _SAPP_LOGITEM_XMACRO(LINUX_X11_DROPPED_FILE_URI_WRONG_SCHEME, "dropped file URL doesn't start with 'file://'") \
+    _SAPP_LOGITEM_XMACRO(LINUX_X11_FAILED_TO_BECOME_OWNER_OF_CLIPBOARD, "X11: Failed to become owner of clipboard selection") \
     _SAPP_LOGITEM_XMACRO(ANDROID_UNSUPPORTED_INPUT_EVENT_INPUT_CB, "unsupported input event encountered in _sapp_android_input_cb()") \
     _SAPP_LOGITEM_XMACRO(ANDROID_UNSUPPORTED_INPUT_EVENT_MAIN_CB, "unsupported input event encountered in _sapp_android_main_cb()") \
     _SAPP_LOGITEM_XMACRO(ANDROID_READ_MSG_FAILED, "failed to read message in _sapp_android_main_cb()") \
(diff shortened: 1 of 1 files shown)


sokol_app.h
```diff
@@ -9609,7 +9613,11 @@ _SOKOL_PRIVATE void _sapp_x11_release_error_handler(void) {
 }
 
 _SOKOL_PRIVATE void _sapp_x11_init_extensions(void) {
-    _sapp.x11.UTF8_STRING             = XInternAtom(_sapp.x11.display, "UTF8_STRING", False);
+    if (_sapp.clipboard.enabled) {
+        _sapp.x11.UTF8_STRING         = XInternAtom(_sapp.x11.display, "UTF8_STRING", False);
+        _sapp.x11.CLIPBOARD           = XInternAtom(_sapp.x11.display, "CLIPBOARD", False);
+        _sapp.x11.TARGETS                = XInternAtom(_sapp.x11.display, "TARGETS", False);
```
floooh, line 9619: A minor thing but can you fix the indentation here?

floooh:
Thanks @qwx9, I'll try to look into PR in more detail and do some testing soon-ish, but most likely not beforre Monday (16-Sep).

sokol_app.h
qwx9: Done, thanks

floooh:
Starting to look into the PR now, but won't get done today.

floooh:
I did some coding style fixes:

https://github.com/floooh/sokol/compare/master...qwx9-master

...in general it seems to work, I can copy/paste in and out of the imgui-sapp sample. But one thing I need to check closer is whether everything works as expected when `_sapp.clipboard.enabled` is false.

For instance the event `SelectionRequest` is always handled, but the handler code checks the Atoms `_sapp.x11.CLIPBOARD` and `_sapp.x11.TARGETS`, but those are only initialized when `_sapp.clip.enabled` is true.

I couldn't seem to cause the event to be fired in an app that has `_app.clipboard.enabled` set to false though.

In one case I also got some sort of invalid target error during debugging from X11 with the application aborting, but I couldn't reproduce this problem and unfortunately I also lost the error message.

In any case, I will continue testing tomorrow, focusing on the 'clipboard disabled' code paths.

floooh:
...and merged with some code cleanup by me.

Many thanks!

fafok29:
I noticed small issue when running on gnome(wayland), copying from sokol app when window is minimized takes 3+ seconds and app I’m copying to freezes for that time, but gnome(xorg) works fine, so not sure if anything can be done here. I used imgui example for test and my own app with the same result, it seems like we are waiting in XPending, so maybe event is just not delivered fast enough for unknown reasons. Also, I tried with multiple applications I’m copying to with no difference.

floooh:
Hmm ok, I tested on my Linux laptop which runs a KDE Plasma session on Wayland, but I didn't test when the window is minified. Maybe it's somehow related to Wayland not calling the frame callback when the window is fully obscured (and I guess also when minified). But I haven't actually checked if that's also the case on XWayland.

PR #1110, opened 2024-09-15
Fix _saudio_backend_shutdown for _SAUDIO_APPLE

badlogic:
The code path tries to return void. Does not compile for obvious reasons.

### DIFF
--- a/sokol_audio.h
+++ b/sokol_audio.h
@@ -2473,7 +2473,7 @@ void _saudio_backend_shutdown(void) {
     #elif defined(SAUDIO_ANDROID_SLES)
         _saudio_sles_backend_shutdown();
     #elif defined(_SAUDIO_APPLE)
-        return _saudio_coreaudio_backend_shutdown();
+        _saudio_coreaudio_backend_shutdown();
     #else
     #error "unknown platform"
     #endif


floooh:
Whoops interesting, do you get a compile error with that? Even though it's an oversight, AFAIK it's valid C/C++, 

For instance the sokol_gfx.h C++ wrappers also return void:

https://github.com/floooh/sokol/blob/1eb96dd0f96b9ea73065f9078244c2255c2b75d9/sokol_gfx.h#L4333-L4335

floooh:
...merged, thanks!

Would still be good to know why that caused trouble on your end.

badlogic:
Returning a value from a void function makes the compiler sad. See also the
"missing" return statements for the other code paths in that function.

badlogic:
Apologies, only saw your last message due to email usage...

I have `-Werror,-Wpedantic` on, which will complain, even if it's valid as per spec.

```
[build] sokol_audio.h:2476:9: error: void function '_saudio_backend_shutdown' should not return void expression [-Werror,-Wpedantic]
[build]         return _saudio_coreaudio_backend_shutdown();
[build]         ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

floooh:
Ah ok. Building with `-Wpedantic` is brave, I'm surprised that even works tbh ;)

I build the sokol test code with `-Wall -Wextra -Werror -Wsign-conversion -Wstrict-prototypes`, but specifically left out `-Wpedantic`.

floooh:
I wonder now if returning void might be valid C++, but not valid C - would be another subtle difference ;)

badlogic:
Could be. My code base is C11, haven't tested with C++. Probably doesn't
matter, because Sokol also supports MSVC :)

PR #1121, opened 2024-10-13
Allow SG_PIXELFORMAT_RG11B10F to be a render target on SOKOL_GLCORE

kcbanner:
Page 209 of https://registry.khronos.org/OpenGL/specs/gl/glspec46.core.pdf (and page 181 of https://registry.khronos.org/OpenGL/specs/gl/glspec43.core.pdf for 4.3) show `R11F_G11F_B10F` as color renderable.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -7494,7 +7494,6 @@ _SOKOL_PRIVATE void _sg_gl_init_pixelformats(bool has_bgra) {
         _sg_pixelformat_all(&_sg.formats[SG_PIXELFORMAT_BGRA8]);
     }
     _sg_pixelformat_all(&_sg.formats[SG_PIXELFORMAT_RGB10A2]);
-    _sg_pixelformat_sf(&_sg.formats[SG_PIXELFORMAT_RG11B10F]);
     _sg_pixelformat_sf(&_sg.formats[SG_PIXELFORMAT_RGB9E5]);
     _sg_pixelformat_srm(&_sg.formats[SG_PIXELFORMAT_RG32UI]);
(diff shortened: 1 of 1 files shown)


floooh:
Good fix, but while looking at the existing pixelformat init code and the GLES3/WebGL2 spec, it would be better to do a small code cleanup first, because the initialization of `SG_PIXELFORMAT_RG11B10F` is in the wrong place...

Basically: SG_PIXELFORMAT_RG11B10F is also renderable on GLES3 and WebGL3 if the EXT_color_buffer_float extension is present (and sokol-gfx already checks for that)...

But that needs some code cleanup, maybe you could do this as part of the PR?

The initialization for `SG_PIXELFORMAT_RG11B10F` needs to be moved down into the function `_sg_gl_init_pixelformats_float`and apart from renderable we also need find out under what conditions the format is blendable (but I guess for a start we can just use the same conventions as for the other float pixel formats in that function.

The rest should then work automatically, e.g. that function `_sg_gl_init_pixelformats_float` is called with all values true on GLCORE, and with the actually available extensions on GLES3/WebGL2.

Thanks!

kcbanner:
That makes sense! I've updated the branch to reflect those changes.

`EXT_float_blend` only specifies that buffers with 32-bit floating point components are blendable, so I didn't mark it as blendable.

floooh:
...and merged, thanks! I'll also write a short blurb in the changelog.

floooh:
...here it is in the pixelformats-sapp sample btw (the red cross is for the missing 'blending' feature, as expected). It also looks the same on WebGL2 (on desktop at least):

Screenshot 2024-10-14 at 11 33 30

PR #1134, opened 2024-10-21
sokol_imgui.h: changed ImTextureID type from void* to uint64_t

DctrNoob:
This PR brings `sokol_imgui.h` up to date with Dear ImGui v1.91.4 which changes the type of `ImTextureID` from `void*` to `uint64_t`.


floooh:
Ok, merged. I also added a little blurb to the changelog. Thanks!

This change didn't actually require any changes in the sokol-samples, but there was an unrelated minor breakage in the cimgui API (the ImDrawList_AddCallback signature has changed).

PR #1137, opened 2024-10-22
Add explicit cast to convert void* to ImTextureID to fix a Clang C++ compile time error.

someposer:
Fixes the following compiler error when using Sokol headers in a C++ project.

```shell
In file included from src/application.cpp:12:
libs/sokol/util/sokol_imgui.h:2434:24: error: incompatible pointer to integer conversion assigning to 'ImTextureID' (aka 'unsigned long long') from 'void *'
 2434 |     io->Fonts->TexID = simgui_imtextureid(_simgui.default_font);
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

This fix is similar to #1134 however, it passes the tests.


floooh:
Closing in favor of #1134 (which I just merged), nonetheless, many thanks for submitting the PR :)

PR #1138, opened 2024-10-23
[nuklear] Added command to enable cursors with default font.

dlannan:
This adds the cursor to be available for use when using the default font. 

To enable cursors in a sokol application:
- First hide the native mouse: ```sapp_show_mouse(false);```
- Then show the nuklear mouse: ```nk_style_show_cursor(ctx);```
The nuklear show must be called with a valid context.

If not using the default font, then after a font is initialized and added manually call:
```nk_style_load_all_cursors(ctx, &your_font_atlas);```
And then use the above operations to show the nuklear cursor.

### DIFF
--- a/util/sokol_nuklear.h
+++ b/util/sokol_nuklear.h
@@ -2152,6 +2152,8 @@ SOKOL_API_IMPL void snk_setup(const snk_desc_t* desc) {
         nk_font_atlas_cleanup(&_snuklear.atlas);
         if (_snuklear.atlas.default_font) {
             nk_style_set_font(&_snuklear.ctx, &_snuklear.atlas.default_font->handle);
+            // This adds the default cursors into the nuklear overlay for use.
+            nk_style_load_all_cursors(&_snuklear.ctx, &_snuklear.atlas.cursors);
         }
     }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by floooh** [comment]
Looks like a nice simple fix (oh, CI fails though, can you look into that?):

MSVC warning:
```
sokol_nuklear.h(2156,55): warning C4047: 'function': 'nk_cursor *' differs in levels of indirection from 'nk_cursor (*)[7]'
```

GCC/Clang error:
```
sokol_nuklear.h:2156:55: error: incompatible pointer types passing 'struct nk_cursor (*)[7]' to parameter of type 'struct nk_cursor *' [-Werror,-Wincompatible-pointer-types]
```

dlannan:
@floooh - sorry. I just went through some of these. Will fix. Needs to be:
```k_style_load_all_cursors(&_snuklear.ctx, &_snuklear.atlas.cursors[0]);```
Running multi-platform tests now to make sure. Will update as needed.

floooh:
Thanks for the fixes! The CI error on Android is unrelated, currently investigating...

floooh:
Ok merged. Thanks!

The software mouse cursor actually feels a lot better than expected :) I tested on a 120 Hz display though.

The `sapp_show_mouse(false/true)` should probably also be hooked up to the `SAPP_EVENTTYPE_MOUSE_ENTER/LEAVE` events because at least on macOS the system mouse cursor remains hidden when moving it outside the application window (I think that's one of those platform specific behaviour in different window systems).

PR #1139, opened 2024-10-26
Add back sg_trace_hooks & sg_install_trace_hooks, not sure why those were ignored

colinbellino:


floooh:
Hmm, I don't know either. The only reason I can think of is that the function pointers in the `sg_trace_hooks` struct break the code generation scripts, but apparently it works (but might not have worked in early versions).

Some of the other language bindings also ignore the trace hooks, others don't. But I guess if it works there's no reason to not enable them.

From my pov you can remove draft status and I can merge.

floooh:
Dang yeah I see it, let me fix that in sokol-shdc and update the PR :)

colinbellino:
I used the Odin bindgen as a base, but at least in Jai it's not an issue at all :)

floooh:
Oki merged. Thanks!

PR #1150, opened 2024-11-13
sokol_nuklear.h: enable setting mouse cursor

zeozeozeo:
Add an option `enable_set_mouse_cursor` to enable setting OS mouse cursors with `sapp_set_mouse_cursor`. By default Nuklear draws its' own custom cursor.

It is the opposite of `disable_set_mouse_cursor` in sokol_imgui.h, to not break existing apps.

Currently this option doesn't actually disable loading the custom cursor icons, and still requires you to call `nk_style_hide_cursor()` to hide Nuklear's cursor.

### DIFF
--- a/util/sokol_nuklear.h
+++ b/util/sokol_nuklear.h
@@ -115,6 +115,12 @@
                 Set this to true if you don't want to use Nuklear's default
                 font. In this case you need to initialize the font
                 yourself after snk_setup() is called.
+            
+            bool enable_set_mouse_cursor
+                If true, sokol_nuklear.h will control the mouse cursor type
+                by calling sapp_set_mouse_cursor(). If using this, you should
+                probably also call nk_style_hide_cursor() to hide Nuklear's
+                custom cursor.
 
     --- At the start of a frame, call:
 
@@ -427,6 +433,7 @@ typedef struct snk_desc_t {
     int sample_count;
     float dpi_scale;
     bool no_default_font;
+    bool enable_set_mouse_cursor;
     snk_allocator_t allocator;          // optional memory allocation overrides (default: malloc/free)
     snk_logger_t logger;                // optional log function override
 } snk_desc_t;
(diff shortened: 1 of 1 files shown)


floooh:
Interesting, I never saw `for (enum ...` :)

I'll let the CI complete and if everything is green I'll merge the PR (and will probably also update one of the sokol-samples to enable mouse cursors).

floooh:
Btw, I seem to remember that there was a specific reason that I'm initializing the mouse cursor before the switch with the currently active mouse cursor, I just don't remember what that specific situation was:

https://github.com/floooh/sokol/blob/7e50f80495dbe7f38cb66e033eb02b463970cf21/util/sokol_imgui.h#L3007

(if your code somehow triggers that problem then we can still fix it in a separate PR I guess)

floooh:
Hmm, somehow I hadn't noticed that the Nuklear software mouse cursor is now always active... that definitely wasn't intended...

zeozeozeo:
> Hmm, somehow I hadn't noticed that the Nuklear software mouse cursor is now always active... that definitely wasn't intended...

Yeah, apparently `nk_style_load_all_cursors()` will always enable the software cursor. Maybe this option should be changed to `disable_set_mouse_cursor`, and disable loading cursors when it is not enabled?

floooh:
I think I will just modify your code to work the same as sokol_imgui.h: rename the flag to `disable...` and by default use the sokol_app.h mouse cursor functions, and otherwise use the Nuklear builtin mouse cursors.

floooh:
Hmm ok, not so simple, looks like calling `nk_style_load_all_cursors()` is what activates the cursor handling in the first place.

floooh:
...I think I'll go with your original code :)

floooh:
Ok, one change I'll do after playing around with it: I'll move the code from the event handler to the start of the snk_render() function. Otherwise the mouse pointer gets stuck at the end of an operation until it is moved again.

floooh:
Ok, all green now. Thanks for the PR!

floooh:
Ok, the two nuklear sokol-samples now use your code:

- https://floooh.github.io/sokol-html5/nuklear-sapp.html
- https://floooh.github.io/sokol-html5/nuklear-images-sapp.html

zeozeozeo:
nice :shipit:

Issue #1154, opened 2024-11-16
Cleanup Emscripten canvas target selector string stuff.


floooh:
See: https://github.com/floooh/sokol/issues/407#issuecomment-2480549987

TL;DR:

- clean up the confusion between DOM element id and selector strings
- change `sapp_desc.html5_canvas_name` to `sapp_desc.html5_canvas_selector`
- change the default from `canvas` to `#canvas`
- remove the special `#` prefixing
- in sokol_app.h JS code, use Emscripten's `findCanvasEventTarget()` and `findEventTarget()` instead of `getElementById()` or `querySelector()` so that the behaviour is identical with the Emscripten runtime functions.
- remove the two redundant getElementById in drag'n'drop init/shutdown code

konsumer:
This path is still using strings (with `querySelector`.)

I recommend passing canvas dom-object (with a selector on `#canvas`/`sapp_desc.html5_canvas_selector` by default, for backwards compatibility.) I think emscripten might even set it up for you, if you just use `Module.canvas`. You can see an example in [glfw](https://github.com/emscripten-core/emscripten/blob/079e2660071d5ae7d45d8e04efc09dd7e67456dd/src/library_glfw.js#L1114)

This way I can override the canvas object, even if there is no document-selector (like in the case of a web-component.)

floooh:
You can't pass a JS object across the WASM and C boundary, and you also can't pass a JS object into the `emscripten_*` functions.

floooh:
PS: e.g. see here: https://emscripten.org/docs/tools_reference/settings_reference.html#disable-deprecated-find-event-target-behavior

`Module['canvas']` is deprecated but still works for backward compatibility.

konsumer:
While I see that it's marked deprecated there, for my use case it really makes it less useful. I wonder what usecase that improves things for (seems like less control, with no payoff.) Still, it seems like a quick check once if canvas is a string (css selector) or dom element is the most consistent way to handle it, looking at other libs, even if it is deprecated behaviour. I could PR for both (so it handles new case and this deprecated thing thAt works everywhere else, and web components rely on.)

konsumer:
Like maybe this is just that I don't know how to do it better, but there are situations where `document` does not point to the DOM that the canvas is in (like in case of shadow-dom) so `document.querySelector('#whatever')` doesn't work. 

[...]

PR #1156, opened 2024-11-18
more canvas options

konsumer:
This switches all to use CSS-selector for canvas (but still support old id thing) but adds ability to use `canvas` option in setup, for advanced situations like web-components, or just having the same wasm run multiple times in a single page.

Related: #1154

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -4930,10 +4930,9 @@ _SOKOL_PRIVATE void _sapp_emsc_set_clipboard_string(const char* str) {
     sapp_js_write_clipboard(str);
 }
 
-EM_JS(void, sapp_js_add_dragndrop_listeners, (const char* canvas_name_cstr), {
+EM_JS(void, sapp_js_add_dragndrop_listeners, (const char* canvas_selector_cstr), {
     Module.sokol_drop_files = [];
-    const canvas_name = UTF8ToString(canvas_name_cstr);
-    const canvas = document.getElementById(canvas_name);
+    const canvas = Module.canvas;
     Module.sokol_dragenter = (event) => {
         event.stopPropagation();
         event.preventDefault();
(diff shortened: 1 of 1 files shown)


sokol_app.h
```diff
@@ -5005,24 +5004,31 @@ EM_JS(void, sapp_js_fetch_dropped_file, (int index, _sapp_html5_fetch_callback c
...
+        return;
+    }
+    
     // lookup and store canvas object by name
-    const target_str = UTF8ToString(c_str_target);
-    Module.sapp_emsc_target = document.getElementById(target_str);
-    if (!Module.sapp_emsc_target) {
-        console.log("sokol_app.h: invalid target:" + target_str);
+    const canvas_selector = UTF8ToString(canvas_selector_cstr);
+    Module.canvas = document.querySelector(canvas_selector);
```
floooh, line 5025: Wouldn't this overwrite the standard `Module['canvas'] that might have been set in JS code?

(apart from that `Module.canvas` might cause problems with the Emscripten Closure pass which tends to minify JS property names, that's why the Emscripten JS shims use `Module['canvas']` which is safe from minification.

konsumer: No, it returns first. Seems like magnification also does not effect this. Can you show an example of it not working?

konsumer: It should be noted, the name is semi-arbitrary, but it makes sesne to me to use the name that is used in other things (SDL, glfw, etc.) In js, if I set my selector to `!canvas` and added `Module.sapp_emsc_target` to special-selectors before loading, it would mostly work, except sokol mixes strategies a bit, in different places, using a selector, id, and the dom-reference. This PR unifies that all to point to 1 thing (the canvas, however you set that up.) In places where you use a selector, it uses `!canvas`, in places where it uses an id, it was moved to use selector, and in places where it uses `sapp_emsc_target` it uses `canvas` which all points to the same thing, now, which is the same as the js-side `canvas` option. I think this actually improves things for people who are not using `canvas` option, because it's all 1 strategy.

PR #1158, opened 2024-11-19
Fix Odin bindgen type mappings

### LINKED ISSUE #1157: Odin Bindgen Type Mismatches
Some of the primitive types don't match when generating Odin bindings. The current types seem to be matched to 64-bit platforms specifically. Namely:

* `intptr_t` is currently `i64`, but [`core:c` currently uses `int`](https://github.com/odin-lang/Odin/blob/master/core/c/c.odin#L71).
* `uintptr_t` is currently `u64`, but should map to [`uintptr`](https://github.com/odin-lang/Odin/blob/master/core/c/c.odin#L72).
* `size_t` is currently `u64`, but should map to [`uint`](https://github.com/odin-lang/Odin/blob/master/core/c/c.odin#L21).

Alternatively, perhaps their aliases from `core:c` should be used instead to ensure compatibility.

Barinzaya:
This affects `intptr_t`, `uintptr_t`, and `size_t`, which are variable-sized. The previous mappings were mapped directly to 64-bit values, and they've been changed to use the corresponding `core:c` aliases.

This closes #1157. An additional fix is required in `sokol-odin` itself to correct `helpers.Allocator.alloc_fn`, which uses `u64` for a `size_t` argument.

### DIFF
--- a/bindgen/gen_odin.py
+++ b/bindgen/gen_odin.py
@@ -105,9 +105,9 @@ prim_types = {
     'uint64_t':     'u64',
     'float':        'f32',
     'double':       'f64',
-    'uintptr_t':    'u64',
-    'intptr_t':     'i64',
-    'size_t':       'u64'
+    'uintptr_t':    'c.uintptr_t',
+    'intptr_t':     'c.intptr_t',
+    'size_t':       'c.size_t'
 }
 
 prim_defaults = {


floooh:
Thanks!

@jakubtomsu when you have a minute, can you also have a look in case you any objections?

(also at: https://github.com/floooh/sokol-odin/pull/22)

floooh:
Looks like at least one sample needs to be fixed as well:

```
/home/runner/work/sokol/sokol/examples/instancing/main.odin(126:16) Cannot assign value 'u64(state.cur_num_particles * size_of(m.vec3))' of type 'u64' to 'uint' in structure literal
	size = u64(state.cur_num_particles * size_of(m.vec3)),
	       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
```

...but that may be a chicken egg problem (e.g. first commit this MR to update the bindings, then fix the sample(s)).

jakubtomsu:
yep, looking at the code everything seems fine.

Barinzaya:
Should I update the `sokol-odin` PR to include the change the sample as well, then?

floooh:
> Should I update the `sokol-odin` PR to include the change to the sample as well, then?

I think that makes sense, even if it might not compile until the sokol PR is merged.

floooh:
Oki thanks for the PRs. I'm currently busy working an unrelated sokol PR, but I'll try to get this stuff merged tomorrow.

floooh:
...also thanks for the PR! :)

PR #1159, opened 2024-11-21
sokol_app.h html5: cleanup canvas lookup handling (see #1154)

floooh:
Proposal for #1154:

- change `sapp_desc.html5_canvas_name` to `.html5_canvas_selector`
- remove the canvas-id vs canvas-selector hack (instead sokol-app now expects a 'css selector string' to be passed in.
- uses the Emscripten helper function `findCanvasEventTarget()` to lookup the JS canvas name by canvas selector (which also considers `Module['canvas']` and `specialHMTLTargets[]`)

TODO:
- [x] test with custom canvas selectors (e.g. `<canvas id="bla">` => `.html5_canvas_selector="#bla"`
- [x] ~~test with setting canvas in index.html via `Module['canvas']`~~ this doesn't actually work any longer
- [x] test with adding canvases to `specialHTMLTargets[]` in index.html (see below)
- [x] in sokol-samples shell.html, remove the Module.canvas() and Module.postRun() functions (those are not actually called anymore)

To register canvases that can't be looked up via `document.querySelector()`, add them to the `specialHTMLTargets[]` array after the Emscripten environment is setup but before the WASM is started. For instance via `Module.preRun()` in index.html:

```js
        preRun: [
          () => {
            specialHTMLTargets['my_canvas'] = my_canvas;
          },
        ],
```

...then in sokol_main() use `.html5_canvas_selector = "my_canvas"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,33 @@
 ## Updates
 
+### 23-Nov-2024
+
+- sokol_app.h html5: Merged PR https://github.com/floooh/sokol/pull/1159 (related
+  issue https://github.com/floooh/sokol/issues/1154).
+
+  This cleans up code that is concerned about finding the HTML canvas to
+  render to by:
+
+    - removing any leftover hacks from the time when Emscripten moved
+      from `document.getElementById()` to `document.querySelector()` for
+      looking up the canvas object
+    - adding two options for canvas objects that can't be looked up via
+      `document.querySelector()`
+
+  If you don't provide a custom canvas name to sokol_app.h this change
+  is non-breaking. Otherwise:
+
+    - in sokol_main(): change `.html5_canvas_name` to `.html5_canvas_selector`
+    - change the canvas name string to a CSS selector string (e.g.
+    from `"my_canvas"` to `"#my_canvas"`)
+
+  For more options to communicate the HTML canvas object to sokol_app.h,
+  please read the new doc section `SETTING THE CANVAS OBJECT ON THE WEB PLATFORM` in sokol_app.h.
+
+  Many thanks to @konsumer for kicking off the feature and the following
+  discussion :)
+
+
 ### 19-Nov-2024
 
 - Merged PR https://github.com/floooh/sokol/pull/1155, this allows to use
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@
   For more options to communicate the HTML canvas object to sokol_app.h,
   please read the new doc section `SETTING THE CANVAS OBJECT ON THE WEB PLATFORM` in sokol_app.h.
 
+  Additionally, please also note the simplified `shell.html` in the
+  sokol-samples repository (some outdated cruft has been removed):
+
+  https://github.com/floooh/sokol-samples/blob/master/webpage/shell.html
+
   Many thanks to @konsumer for kicking off the feature and the following
   discussion :)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,7 @@
 - sokol_app.h html5: Merged PR https://github.com/floooh/sokol/pull/1159 (related
   issue https://github.com/floooh/sokol/issues/1154).
 
-  This cleans up code that is concerned about finding the HTML canvas to
-  render to by:
+  This cleans up code that is concerned about finding the WebGL/WebGPU HTML canvas by:
 
     - removing any leftover hacks from the time when Emscripten moved
       from `document.getElementById()` to `document.querySelector()` for
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -5095,8 +5095,6 @@ EM_JS(void, sapp_js_remove_dragndrop_listeners, (void), {
 
 EM_JS(void, sapp_js_init, (const char* c_str_target_selector), {
     const target_selector_str = UTF8ToString(c_str_target_selector);
-    // Special way to lookup the canvas by setting Module['canvas'] outside
-    // the WASM, this may be useful when the canvas can't be setup via document.querySelector()
     if (Module['canvas'] !== undefined) {
         if (typeof Module['canvas'] === 'object') {
             specialHTMLTargets[target_selector_str] = Module['canvas'];
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -4931,7 +4931,6 @@ _SOKOL_PRIVATE void _sapp_emsc_set_clipboard_string(const char* str) {
 
 EM_JS(void, sapp_js_add_dragndrop_listeners, (void), {
     Module.sokol_drop_files = [];
-    const canvas = Module.sapp_emsc_target;
     Module.sokol_dragenter = (event) => {
         event.stopPropagation();
         event.preventDefault();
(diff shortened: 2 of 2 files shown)


konsumer:
This doesn't seem terribly different from my solution (switch id called "name" to selector) but you need an extra step (to setup the selector in a preRun.) Additionally, does that actually work with your `Module.sapp_emsc_target`, I think it might not. What does this solve over my solution? In mine, it supports the usecase of setting your own selector (you don't have to use `!canvas` for example, you can use `my_canvas`, like here) and handles a few more usecases, but also has sensible defaults (if you set the canvas element, it will set it up for you.)

konsumer:
> I think the main difference is that it automatically works with all the emscripten_*() functions, also for canvases that can't be looked up via document.querySelector()

I'm not sure what you mean here. My functions works with selectors too.

> (it doesn't have to be !canvas but can be any string).

That's true for mine too. I just used`!canvas` as default because it's common.  You can still do it manually, as you are suggesting with my path, as a separate step, if you prefer that for some reason. it's only the default, if you passed a `canvas` option.

> I also checked, and the Emscripten functions do not find a canvas that's provided via Module['canvas'] (I guess because this option has been default-enabled for quite a while

That is exactly what  I was saying. It's no longer automatic, and that is what those options are about. The old behavior was that you need `#canvas` in the same page, and it was used automatically, and that was mega-jenky, then they allowed override with `canvas` option, then they deprecated all that as built-in to the emscripten layer, automatic, but libs kept following it, like in their own layer. You do very similar currently, just not totally the same (more like the deprecated emscripten behavior) and not in a way that can be used with web-components or multiple canvases with multiple instances of the same wasm. Even if you consider my usecase so edge that it's not worth considering, it's nice to give users a few ways to connect their canvas to sokol, and that will work as they expect. Currently, if you provide a `canvas` option, in js, the standard libs will do roughly what I am proposing, and my path doesn't disable anything that sokol did before.

For my solution, you get more options then how it currently works, and a bit more than (but including options provided by) this PR:
- you can hardcode id, as sokol already did, before
- you can use `canvas` option, and it will bind the selector to it
- you can use hardcode a query-selector, and it will also work with everything
- you can bind your own `specialHTMLTargets[]` to some other (or the same `!canvas`) selector, and juggle things in `preRun`

In each path on my solution you get the same 2 things (`Module.canvas` for js, and a string selector for the C wrappers.)

I think there is a communication breakdown somewhere here, since it seems so clearly better to me to do it my way, and I see no benefit to doing it this way. Like, it's not simpler, more elegant, and it gives users less options, and is a bit more cumbersome. There is no difference with performance or memory-usage, or anything else, and It also does not work like other libs people might be used to (SDL, glfw, raylib, etc.)

Fundamentally, it's your lib, and you can do whatever you like with that. It's not really usable to me, as it is, passing the id from C-side, in my current project, and I don't see the point of forcing users to set up that extra layer with `specialHTMLTargets` & `preRun`, when it's so simple to just do it in sokol, and support all the options, but it's your thing, and your choice to make. I can use a forked version of sokol_app.h, or wait for your PR to go in, and use it in the cumbersome fashion you suggest, or just not use it.

floooh:
> I'm not sure what you mean here. My functions works with selectors too.

I mean these:

https://github.com/floooh/sokol/blob/8c989e39c008149a8a6b89cd001a2af3dbe3a8c2/sokol_app.h#L5834-L5850

...and these:

https://github.com/floooh/sokol/blob/8c989e39c008149a8a6b89cd001a2af3dbe3a8c2/sokol_app.h#L5941-L5951

...those functions that a selector string and then call Emscripten's `findCanvasEventTarget()` function, and that first looks in `specialHTMLTargets` and if not found there do a `document.querySelector()` (but don't know about `Module['canvas']`.

floooh:
PS: there's also this detail in the WebGPU C API:

https://github.com/emscripten-core/emscripten/blob/b3edd4cdb6f8055f7b50e3cc406b8ec7b2232afb/system/include/webgpu/webgpu.h#L1136-L1140

...e.g. the WebGPU shim is also expecting a CSS selector string to find the canvas object and this also uses the `findCanvasEventTarget()` function under the hood:

https://github.com/emscripten-core/emscripten/blob/b3edd4cdb6f8055f7b50e3cc406b8ec7b2232afb/src/library_webgpu.js#L2494

floooh:
This is an interesting part in your PR:

```js
    if (Module.canvas) {
        specialHTMLTargets["!canvas"] = Module.canvas;
        stringToUTF8("!canvas", canvas_selector_cstr, 8);
        return;
    }
```

E.g. don't expect `specialHTMLTargets[]` to be initialized outside the WASM, but instead expect a canvas object to be set in a special place (like `Module['canvas']`) and patch `specialHTMLTarget` as part of the sokol_app.html setup. I would reverse the canvas name thing and simply do this:

```js
    if (Module['canvas']) {
        const canvas_selector_str = UTF8ToString(canvas_selector_cstr);
        specialHTMLTargets[canvas_selector_str] = Module['canvas'];
    }
```

I could live with that too.

I'm using `Module['canvas']` instead of `Module.canvas` because the Closure compiler usually tries to minify the `Module.canvas` form.

floooh:
Ok see:

https://github.com/floooh/sokol/pull/1159/commits/05190b6d4359eaa87b1e44f98e8460364f16fc24

konsumer:
> ...those functions take a selector string and then call Emscripten's findCanvasEventTarget() function, and that first looks in specialHTMLTargets and if not found there do a document.querySelector() (but it doesn't know about Module['canvas']),

Did you try it? As I said, my code sets the string selector for the emscripten functions, and in js you can use `Module.canvas`. It should  all work fine, allowing more options, not less.  It's not that different from how it was, but you used another name, and you mixed 3/4 strategies, where I use 2 (1 for js and 1 for C.) Using `canvas` makes it so the options work automatically (I can skip setting the cached copy of js canvas, since it's already set.)

> ...e.g. the WebGPU shim is also expecting a CSS selector string to find the canvas object and this also uses the findCanvasEventTarget() function under the hood:

Again, so do I.

> Ok see:

As I said, I think you are misreading the deprecated behavior. It's not "Don't name your copy of the canvas `Module.canvas`" (several other libraries do it this way, and it's a good way to set the option, since it stays the same name all the way through any js code.) It's saying "emscritpen used to automatically hook `#canvas` up for you" which I agree is stupid. The current behavior of sokol is more like that legacy behavior, though.

konsumer:
Is there a way to chat on discord or something? That may be a better forum for this sort of back and forth. I am happy to step through my code. Again, I'm not currently using sokol, so I'm not really sure what the point is for me, but I really don't like being misunderstood, and I do think it would help sokol to be more useful to others.

floooh:
> Is there a way to chat on discord or something?

No, tbh I'll just merge my PR and be done with it, we already wasted too much time on a trivial feature :)

One thing I'll need to fix is checking that `Module['canvas']` is actually a JS object, because it's not unusual that index.html defines it as a function that returns a canvas (must be some even older deprecated behaviour).

I'm quite convinced that `Module['canvas']` is deprecated. Library shims in the Emscripten SDK like GLFW or SDL are not a good place to look at because those are pretty much abandondend by the Emscripten devs (Emscripten platform support for SDL has moved upstream directly into SDL).

floooh:
Oki, dropping the sokol_app.h from this PR into your multicanvas example (https://github.com/konsumer/sokol-canvas-example-1154) works without changes. I'll just keep it at that.

PR #1165, opened 2024-12-06
Add Swift 6 bindgen support

kassane:
applied in: https://github.com/kassane/sokol-swift

The initial tests during the development of a ported example. Some issues appeared that are explained in reference below.
Not to mention the difficulties associated with the use of custom-modules by `.swift` or `.swiftinterface`. If not, **swiftc** may find redundancies in the translated files, e.g. multiple `struct Range'. (probably while analyzing the files in an attempt to unify all the implementations...)

### Reference
- https://github.com/kassane/sokol-swift/issues/2

cc: @floooh

### DIFF
--- a/bindgen/.gitignore
+++ b/bindgen/.gitignore
@@ -8,3 +8,4 @@ sokol-odin/
 sokol-rust/
 sokol-d/
 sokol-jai/
+sokol-swift/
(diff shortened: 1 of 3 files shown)


floooh:
FYI: I'm keeping an I on this, but if I should do something (like merging PRs) please give me an extra ping. The next thing on the list is probably the C3 bindings though (not sure yet when those will be ready for merging).

PR #1167, opened 2024-12-11
Allow to have both WinMain and main defined on Windows

Querijn:
My library uses Sokol under the hood and requires you to define 

```
add_executable(example WIN32 main.cpp)
target_link_libraries(example MyLib) # MyLib uses Sokol
```

when on Windows by default. If you were to use `add_executable(example main.cpp)` instead, it would not link due to the missing main entrypoint. I could define SOKOL_WIN32_FORCE_MAIN for MyLib to force main(), but then I could not create a WinMain unless I removed it again. I considered doing NO_ENTRY and defining both locally in the library but that seemed like more work (and less in the spirit of open-source) than making a PR.

This PR addresses this problem, creating a definition for `SOKOL_WIN32_FORCE_WINMAIN`. This allows you to force either or both, where the default behaviour is the original behaviour.

- Defining neither is effectively the same as defining SOKOL_WIN32_FORCE_WINMAIN, which creates a WinMain entry point.
- Defining only SOKOL_WIN32_FORCE_MAIN will create just the main entry point, and not the WinMain entry point.
- Defining both will create both.

I'm not happy about the length of the extra line I added in the documentation, and I'm open to modify it if you have a better idea.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -49,6 +49,9 @@
     On Windows, SOKOL_DLL will define SOKOL_APP_API_DECL as __declspec(dllexport)
     or __declspec(dllimport) as needed.
 
+    if SOKOL_WIN32_FORCE_MAIN and SOKOL_WIN32_FORCE_WINMAIN are both defined,
+    it is up to the developer to define the desired subsystem.
+
     On Linux, SOKOL_GLCORE can use either GLX or EGL.
     GLX is default, set SOKOL_FORCE_EGL to override.
 
@@ -2186,8 +2189,11 @@ inline void sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
     #include <windows.h>
     #include <windowsx.h>
     #include <shellapi.h>
-    #if !defined(SOKOL_NO_ENTRY)    // if SOKOL_NO_ENTRY is defined, it's the applications' responsibility to use the right subsystem
-        #if defined(SOKOL_WIN32_FORCE_MAIN)
+    #if !defined(SOKOL_NO_ENTRY)    // if SOKOL_NO_ENTRY is defined, it's the application's responsibility to use the right subsystem
+
+        #if defined(SOKOL_WIN32_FORCE_MAIN) && defined(SOKOL_WIN32_FORCE_WINMAIN)
+            // If both are defined, it's the application's responsibility to use the right subsystem
+        #elif defined(SOKOL_WIN32_FORCE_MAIN)
             #pragma comment (linker, "/subsystem:console")
         #else
             #pragma comment (linker, "/subsystem:windows")
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -27,13 +27,14 @@
 
     Optionally provide the following defines with your own implementations:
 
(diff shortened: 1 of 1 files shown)


floooh:
I wonder if it works to just remove all the FORCE defines and just always have both main() and WinMain() in the code (unless  SOKOL_NO_ENTRY is defined which would remove both main() and WinMain()). Did you try that already? I'll need to tinker with this idea a bit, if it works I would prefer that since it removes a bit of configuration complexity.

Querijn:
I did! That is what I did before I decided to make this PR. It worked fine on my machines, Windows 10/Windows 11 were tested at the time.

Only concern I can think is that maybe existing systems depend on the existence of either, not both and are therefore configured on that. Defining both would break this very hypothetical case.

floooh:
> Oh, and the subsystem has to be explicitly defined...

...does the pragme-comment magic at that place actually work with your PR when both `SOKOL_WIN32_FORCE_WINMAIN` and `SOKOL_WIN32_FORCE_MAIN` are defined?

I also just saw that there's this outdated thingie (outdated because UWP / WinRT is no longer supported):

https://github.com/floooh/sokol/blob/a94f66a4823cabe63374c93c502331e83642b70f/sokol_app.h#L2002-L2008

Querijn:
> > Oh, and the subsystem has to be explicitly defined...
> 
> ...does the pragma-comment magic at that place actually work with your PR when both `SOKOL_WIN32_FORCE_WINMAIN` and `SOKOL_WIN32_FORCE_MAIN` are defined? E.g. I guess you need to override that in your build script's linker setting to pick one or the other?

CMake + the Visual Studio generator solves this correctly _somehow_, I did not need any changes to this and the correct subsystem was used..
 
> PS: maybe it makes sense to not set any subsystem at all via pragma-comment when both `SOKOL_WIN32_FORCE_MAIN` and `SOKOL_WIN32_FORCE_WINMAIN` is defined?

Maybe removing the force altogether is the better call, but yeah, that makes sense.

Querijn:
I've checked again, enabling both will open up a console window even when starting through WinMain. Which might be undesired behaviour. So I'm convinced removing these is a better call.

floooh:
Yeah, my current thinking is: keep the idea to have both FORCE defines (e.g. what your PR does), but change this block so that no subsystem is set when **both** are defined:

```
     #if defined(SOKOL_WIN32_FORCE_MAIN) 
         #pragma comment (linker, "/subsystem:console") 
     #else 
         #pragma comment (linker, "/subsystem:windows") 
     #endif
```

Can you update the PR for that?

Also, while you're at it, can you remove this leftover UWP snippet? Thx :)

https://github.com/floooh/sokol/blob/a94f66a4823cabe63374c93c502331e83642b70f/sokol_app.h#L2002-L2008

Querijn:
This should be it. Would you like me to squash it or is this OK?

floooh:
Looks good now. I'll play around with it a bit though, will most likely merge on Saturday.

floooh:
Ok merged, thanks! I'll also write a little changelog entry.

PR #1172, opened 2024-12-28
gen_odin.py: Use shared linux libraries if USE_DLL

AlexanderArvidsson:
This adds support for shared libraries on linux, enabling hot reloading without having to mess with the sokol app state. Hot reloading via static lib causes issues which can only be solved by persisting the internal sokol app state, which isn't available from Odin.
With a shared library, hot reloading works out of the box and makes development super fast.

This requires https://github.com/floooh/sokol-odin/pull/24 to be merged first.


floooh:
One second, need to fix the D GH Action script first....

floooh:
...and merged, thanks!

PR #1182, opened 2025-01-11
Include comments in Jai bindgen

colinbellino:
I saw what was added in https://github.com/floooh/sokol/pull/1176 so i thought i'd do a quick update for the Jai bindgen too :)


floooh:
...and merged. Thanks!

PR #1187, opened 2025-01-12
D bindgen: add memtrack and fix fetch update

kassane:
* fix fetch update
* add memtrack

cc: @floooh

**Note**: Tried add `nuklear`, `spine` and `gfx_imgui`, but get struct error!!

```console
ERROR gen_struct: sspine_t * spine;
ERROR gen_struct: sgimgui_buffer_t * slots;
```

### DIFF
--- a/bindgen/gen_d.py
+++ b/bindgen/gen_d.py
@@ -26,6 +26,7 @@ module_names = {
     'sglue_':   'glue',
     'sfetch_':  'fetch',
     'simgui_':  'imgui',
+    'smemtrack_': 'memtrack',
 }
 
 c_source_paths = {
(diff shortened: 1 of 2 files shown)


bindgen/gen_d.py
```diff
@@ -91,6 +93,7 @@
 prim_defaults = {
     'int':          '0',
     'bool':         'false',
+    'char':         '0',
```
kassane, line 96: Has required in `gfx_imgui` bindgen test

floooh:
For some of the headers under util, the hassle to effort to make them bindings compatible probably isn't worth the effort (especially when they depend on 3rd party libs like sokol_spine.h).

floooh:
...looks like sokol_memtrack.h also has issues. For sokol_memtrack.h I also wonder if that's all that useful, since it's such a trivial thing to do natively in the target language.

kassane:
> For sokol_memtrack.h I also wonder if that's all that useful, since it's such a trivial thing to do natively in the target language.

In D case, users could use this wrapper to access `malloc`/`free`  with no interference from GC. Unfortunately, not all users feel comfortable using `malloc`/`free` directly.
This avoids creating your own allocation solution.

floooh:
Oki, is the PR ready, or do you want to do more changes? :)

kassane:
Done!

PR #1200, opened 2025-01-25
[WIP] compute shader support

floooh:
Related sokol-shdc PR: https://github.com/floooh/sokol-tools/pull/173

- [ ] after merge:
    - [ ] fix blogpost links
- [ ] merge:
    - [x] https://github.com/floooh/sokol-zig/pull/100
    - [x] https://github.com/floooh/sokol-odin/pull/28
    - [x] https://github.com/floooh/sokol-nim/pull/35
    - [x] https://github.com/floooh/sokol-rust/pull/33
    - [x] https://github.com/floooh/sokol-c3/pull/1
    - [x] https://github.com/kassane/sokol-d/pull/56
    - [x] https://github.com/colinbellino/sokol-jai/pull/5
    - [x] https://github.com/floooh/sokol-tools/pull/173
    - [ ] https://github.com/floooh/sokol-samples/pull/162
- [x] test new samples on:
    - [x] macOS+Metal
    - [x] iOS+Metal
    - [x] WebGPU
    - [x] Windows+D3D11
    - [x] Windows+GL
    - [x] Linux
- [x] bindings:
    - [x] zig
        - [x] sokol-shdc
        - [x] example
    - [x] odin
        - [x] sokol-shdc
        - [x] example
    - [x] c3
        - [x] sokol-shdc
        - [x] example
    - [x] nim
        - [x] sokol-shdc
        - [x] example
    - [x] rust
        - [x] sokol-shdc
        - [x] example
    - [x] d
        - [x] sokol-shdc
        - [x] example
    - [x] jai
        - [x] sokol-shdc
        - [x] ~~example~~
    - [x] yaml
- [ ] general stuff:
    - [x] sokol-shdc documentation
    - [x] sokol-gfx documentation
    - [x] blog post
    - [x] sokok-shdc changelog
    - [ ] sokol changelog
    - [x] sample webpage
        - [x] webgl2
        - [x] webgpu
    - [x] GL backend: ifdef-compute-not-available
    - [x] GL backend: use glMemoryBarrier (see: https://arm-software.github.io/opengl-es-sdk-for-android/compute_intro.html)
        - [x] track a 'gpu_dirty' bool in sg_buffer.gl, in sg_apply_bindings issue a memory barrier if any of the storage buffers in the bindings is 'gpu_dirty', and reset the dirty flag, do this across passes and frames
    - [x] ~~Metal backend: remove the managed-buffer-synchronization and use the same granular approach as the memory barrier code in GL~~ => this wouldn't work because the buffer sync needs to happen outside a compute pass
    - [x] ~~change the internal names of the various `*_info` structs to not be abbreviated so much~~
    - [x] rename `sg_features.storage_buffer` to `sg_features.compute`
    - [x] ~~write a compute demo which samples from regular textures (just to check if texture sampling works in compute passes)~~
    - [x] port wgpu-compute-boids
    - [x] ~~drive-by: remove return value from `_sg_*_apply_bindings()`, all implementations always return true anyway~~ => it's actually used in the wgpu backend
    - [x] investigate D3D11 Refcounting Leaks on shutdown (doh, missing `d3d11_shutdown()`
- [x] update buffer init behaviour:
  - [x] no longer initialize desc.data size from desc.size (and stricter validation for data.size) - code cleanup, not directly related to compute shader
  - [x] GL specific hazard tracking:
      - [x] track a per-buffer flag when a storage buffer was bound for read/write (cleared
             cleared when bound for read-only
      - [x] if a storage buffer is bound in any pass-type and no matter if read/only or read/write and the read/write flag in the buffer was set, issue a glMemoryBarrier()) - don't need other glMemoryBarrier types since storage buffers cannot currently be re-used as vertex- or index-buffer.
      - [x] remove the other unneeded current hazard-tracking code
  - [x] allow creating immutable buffers without content (will be initialized to zero)
      - [x] metal
      - [x] d3d11
      - [x] gl
      - [x] wgpu
- [x] compute shader objects
    - [x] metal
    - [x] d3d11
        - [x] ~~on `sg_shader_storage_buffer` rename `hlsl_register_t_n` to `hlsl_register_t_or_u_n`~~
    - [x] gl
    - [x] wgpu
- [x] compute pipelines
    - [x] metal
        - [x] investigate: setThreadgroupMemoryLength (this is only need for dynamically sized threadgroup memory, not for statically allocated `threadgroup` variables (e.g. SPIRVCross turns GLSL `shared` variables into statically allocated threadgroup variables)
        - [x] semi-related: set bound buffer immutability:
             - in render pipelines: all bound vertex- and storage-buffers are MTLMutabilityImmutable
             - in compute pipelines: read-only storage buffers are MTLMutabilityImmutable
    - [x] d3d11
    - [x] gl
    - [x] wgpu
- [x] compute passes
    - [x] postponed ~~NOTE: concurrent compute pass encoders require bumping the minimal supported macOS base version to 10.14~~
    - [x] track gpu-written storage-buffers in compute passes
    - [x] metal
        - [x] synchronize written storage buffers (managed buffers only - this blits the gpu-written data back into the cpu-side managed buffer, do this at end of a compute pass)
    - [x] d3d11
    - [x] gl
    - [x] wgpu
- [x] apply_bindings
    - [x] metal
        - [x] figure out if samplers are supposed to work in compute passes, e.g. MTLComputeCommandEncoder allows to bind samplers but how does that work without the partial-derivatives?
        - [x] apply images and samplers in compute passes
    - [x] d3d11
        - [x] ??? - works without d3d11 validation layer warnings, maybe doesn't apply to "raw buffers"? hmm.. ~~"Buffer resources used for output from the compute shader must be created with the D3D11_BIND_RENDER_TARGET flag. Such resources may be read from, however."~~
        - [x] ??? works without d3d11 validation layer warnings, maybe doesn't apply to "raw buffers"? ~~"Buffer resources created with the D3D11_BIND_SHADER_RESOURCE flag may only be used as inputs to the compute shader."~~
    - [x] gl
    - [x] wgpu
- [x] dispatch
    - [x] metal
        - [x] ~~assume iPhone8 (e.g. non-uniform thread size feature)~~ (probably don't need this with the below changes)
        - [x] rewrite: take workgroup size from shader reflection
        - [x] rewrite: dispatch takes num_groups, not num_threads
    - [x] d3d11
    - [x] gl
    - [x] wgpu 
- [x] validation layer / assert:
    - [x] render/compute shader vs SG_SHADERSTAGE_COMPUTE
    - [x] validate sg_shader_desc.mtl_threads_per_threadgroup (must be > 0, question is just: what's the upper bound?)
    - [x] don't call `sg_dispatch()` in render passes
    - [x] check that the `num_groups_*` in the dispatch call doesn't exceed 0xFFFF (see D3D11 functional spec)
    - [x] don't call `sg_draw()`, `sg_apply_viewport`, `sg_apply_scissor_rect` in compute passes
    - [x] storage buffer bindings on vertex- or fragment-stage must be readonly
    - [x] in `sg_begin_pass()` don't set attachments or swapchain in compute passes
    - [x] in `sg_apply_bindings()` don't bind vertex/index buffers, ~~textures and samplers~~ (textures and samplers should be fine)
    - [x] ~~in `sg_make_shader()`: don't mix texture, sampler and writable storage buffer bindings~~
    - [x] don't use read/write storage resource bindings in render pipeline objects
    - [x] ...?
- [x] trace hooks and stats tracking
    - [x] stats: num_dispatch
    - [x] update sokol_gfx_imgui.h
    - [x] ...?

Questions and open problems:

- [YES] support thread-shared memory? (requires special CPU side support in Metal: ~~https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/setthreadgroupmemorylength(_:index:)?language=objc)~~ => setThreadGroupMemoryLength is only needed when the length isn't known to the shader (e.g. the shared memory isn't statically allocated in the shader itself, we don't need that since SPIRVCross always statically allocates)
- [YES] should we just allow multiple writes to the same resource and insert the necessary barriers in the APIs which require it?
- [DONE] ||in HLSL, GLSL and WGSL the shader defines the workgroup size, but in Metal that's defined on the CPU side in the dispatch call.
    - GLSL: `layout(local_size_x = X​, local_size_y = Y​, local_size_z = Z​) in;`
    - HLSL: `numthreads[x, y, z];`
    - WGSL: `@workgroup_size(x,y,z)`
    - then the dispatch call differs by either providing the number of threads, or the number of workgroup invocation in each dimension (the latter implies that number of threads is an integer multiple of number of workgroups)
    - Solution:
        - follow WebGPU convention
        - need to provide workgroup size as shader reflection information in sg_shader_desc, and this will be passed into the Metal dispatch call... (but needs to be evaluated against the max allowed sizes)
        - need to figure out the max-safe-workgroup-size values (again, what does WebGPU do?)
        - ok, WebGPU does the `dispatchWorkgroups` thing, e.g. if workgroup_size is 64, then: `passEncoder.dispatchWorkgroups(Math.ceil(numParticles / 64));`
        - ...still need to figure the max safe values for workgroup size though.
 
- [LATER] storage textures now or later?
- [LATER] split SG_BUFFERTYPE into a bool-flags struct? (to allow storage+vertex buffers), alternatively: just do vertex pulling from storage buffers...?


septag:
Congrats on compute-shaders for the upcoming support. Unfortunately, I don't have any projects to use sokol_gfx right now, but great news nevertheless!

PR #1204, opened 2025-01-25
Make Odin bindings respect common vetting rules + easier to use WASM

karl-zylinski:
Odin binding improvements:

1. Easier to use with web builds.
2. Compile cleanly with `-vet -strict-style`

(1) Is done by linking to `env.o` (the wasm environment) in case of wasm build. This assumes you use emscripten and link the sokol libs using `emcc`. But that's the only way to currently use sokol on the web, so I guess that's an OK assumption.

(2) Is done to avoid it will saying that the import `c` is unused.

I'll do a separate PR with the WASM lib build scripts (sokol-odin is missing them).


floooh:
Okidoki. 

The `wasm32` vs `wasm64p32` in Odin is quite confusing tbh, since everywhere else wasm32 means `32-bit pointers and 64-bit integers` (e.g. there is no version of WASM which restricts integers to 32 bits).

Does it even make sense to support Odin's idea of `wasm32` instead of going `wasm64p32` all the way?

floooh:
Ok, my question was exhaustively answered on the Odin Discord ;)

floooh:
...and merged... lots of stuff in the pipeline today, I may need to spread this over the next few days, just FYI.

PR #1205, opened 2025-01-30
sokol_args: Fix Emscripten extra semi-colon

RobLoach:
In Emscripten, the `EM_JS` function should not end with a `;`. This change fixes that for both instances.

This is the error that you get when using the `;`...
```
sokol_args.h:688:63: error: extra ';' outside of a function [-Werror,-Wextra-semi]
  688 | EM_JS_DEPS(sokol_audio, "$withStackSave,$stringToUTF8OnStack");
      |                                                               ^
sokol_args.h:732:3: error: extra ';' outside of a function [-Werror,-Wextra-semi]
  732 | });
      |   ^
2 errors generated.
```

### DIFF
--- a/sokol_args.h
+++ b/sokol_args.h
@@ -685,7 +685,7 @@ extern "C" {
 #endif
 
 #if defined(EM_JS_DEPS)
-EM_JS_DEPS(sokol_audio, "$withStackSave,$stringToUTF8OnStack");
+EM_JS_DEPS(sokol_audio, "$withStackSave,$stringToUTF8OnStack")
 #endif
 
 EMSCRIPTEN_KEEPALIVE void _sargs_add_kvp(const char* key, const char* val) {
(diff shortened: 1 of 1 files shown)


floooh:
Oh, that's a nice warning! Thanks for the hint (I'll wait with the merge until I have a bit of time to also add that warning to the compilation test that run in CI).

floooh:
...plenty of similar warnings in other headers too, mostly after EM_JS... I'm cleaning those up now (and will that that `-Wextra-semi` to the CI warning set).

Interestingly, GCC doesn't seem to have a similar warning option...

floooh:
Interesting, GCC only accepts -Wextra-semi in C++ mode, but not in C mode:

```
error: command-line option ‘-Wextra-semi’ is valid for C++/ObjC++ but not for C [-Werror]
```

Issue #1208, opened 2025-02-06
sokol_gfx.h: Missing SG_BLENDOP_MIN/SG_BLENDOP_MAX?


jdah:
It looks like all backends support using MIN and MAX as blend operations:

* GL: https://registry.khronos.org/OpenGL-Refpages/gl4/html/glBlendEquation.xhtml
* D3D: https://learn.microsoft.com/en-us/windows/win32/api/d3d11/ne-d3d11-d3d11_blend_op
* Metal: https://developer.apple.com/documentation/metal/mtlblendoperation
* WGPU: https://docs.rs/wgpu/latest/wgpu/enum.BlendOperation.html (not sure if these are the right docs?)

is there any reason these have been excluded from sokol_gfx? If not, I'll go ahead and add them in :)

jdah:
candidate (untested on everything except GL) PR: https://github.com/floooh/sokol/compare/master...jdah:sokol:master

Is there a quick way to run cross-platform compile tests from a Mac?

floooh:
I as wondering if WebGL2 is the problem, but seems like GLES 3.0 supports it (so WebGL2 should too): https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glBlendEquation.xhtml).

It's probably an oversight from after WebGL1 / GLES2 support was kicked out.

I'll also need to check on macOS, macOS GL seems to be a weird mix of features between 3.2 and 4.1, but not a complete 4.1 implementation.

Can you put your fix into a PR?

What's your intended use case for this btw?

floooh:
> Is there a quick way to run cross-platform compile tests from a Mac?

Not really, best bet would be the Zig toolchain, but I don't have any build.zig setup for the sokol samples.

jdah:
> I as wondering if WebGL2 is the problem, but seems like GLES 3.0 supports it (so WebGL2 should too):

From what I can glean from some old stack overflow questions, it looks like some older GLES 3 devices didn't have support for it.

> I'll also need to check on macOS, macOS GL seems to be a weird mix of features between 3.2 and 4.1, but not a complete 4.1 implementation.

It works on my M1 Pro with macOS 15.3, at least

>  What's your intended use case for this btw?

Nice lines that look (kind of) like they were drawn on a vector display :-)

<img width="472" alt="Image" src="https://github.com/user-attachments/assets/9043c55c-9de8-4308-9fd5-3ed8de201a92" />

[...]

PR #1209, opened 2025-02-07
Add SG_BLENDOP_MIN/MAX

jdah:
Adds SG_BLENDOP_MIN/MAX

See #1208 and https://github.com/floooh/sokol-samples/pull/160

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -2452,6 +2452,8 @@ typedef enum sg_blend_op {
     SG_BLENDOP_ADD,
     SG_BLENDOP_SUBTRACT,
     SG_BLENDOP_REVERSE_SUBTRACT,
+    SG_BLENDOP_MIN,
+    SG_BLENDOP_MAX,
     _SG_BLENDOP_NUM,
     _SG_BLENDOP_FORCE_U32 = 0x7FFFFFFF
 } sg_blend_op;
@@ -7364,6 +7366,8 @@ _SOKOL_PRIVATE GLenum _sg_gl_blend_op(sg_blend_op op) {
         case SG_BLENDOP_ADD:                return GL_FUNC_ADD;
         case SG_BLENDOP_SUBTRACT:           return GL_FUNC_SUBTRACT;
         case SG_BLENDOP_REVERSE_SUBTRACT:   return GL_FUNC_REVERSE_SUBTRACT;
+        case SG_BLENDOP_MIN:                return GL_MIN;
+        case SG_BLENDOP_MAX:                return GL_MAX;
         default: SOKOL_UNREACHABLE; return 0;
     }
 }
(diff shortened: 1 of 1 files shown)


floooh:
Interesting, WebGPU puts restrictions on the combination of blend factors:

Screenshot 2025-02-09 at 11 51 11

I'll check what the restrictions are, and will fix the sample (but not enforce the restrictions in the sokol-gfx valiation layer, since at least WebGL2 doesn't complain).

floooh:
...I guess since the formulas for blend-op-min/max are `min(RGBAsrc, RGBAdst)` and `max(RGBAsrc, RGBAdst)` (e.g. no blend-factor involved) I guess that other APIs just implicitly use a blend-factor of one.

I will use ONE/ONE as defaults when using blend-op-min/max and the blend factors are not explicitly set, and add a validation layer error if other blend factors are set in the sg_pipeline_desc.

jdah:
> ...
> ...ok, so in WebGPU, all blend-factors must be ONE in combination with MIN/MAX blend-op, I'll change the defaults and the valiation layer accordingly (e.g. when only setting the blend-op but not the blend-factors in the pipeline-desc, the blend-factors will be set to ONE, and the validation layer will check that the blend-factors are ONE).

Interesting - looks like this is documented in webgpu (https://gpuweb.github.io/gpuweb/#fragment-state) but implicit in at least the OpenGL docs (https://registry.khronos.org/OpenGL-Refpages/gl4/html/glBlendEquation.xhtml). don't recall that D3D mentions it anywhere either.

That sounds like a reasonable approach though - so all platforms will give validation layer errors when using MIN/MAX and blend factors that aren't ONE? or just webgpu?

floooh:
> That sounds like a reasonable approach though - so all platforms will give validation layer errors when using MIN/MAX and blend factors that aren't ONE? or just webgpu?

Yes, but if you create the pipeline object without explicitly defining blend-factors, it will work, e.g.:

```c
sg_make_pipeline(&(sg_pipeline_desc){
    .colors[0].blend = {
        .op_rgb = SG_BLENDOP_MIN,
    },
});
```
...in that case I just use `SG_BLENDFACTOR_ONE` as the default for both blend factors.

...the validation layer will only complain when you do something like this:

```c
sg_make_pipeline(&(sg_pipeline_desc){
    .colors[0].blend = {
        .op_rgb = SG_BLENDOP_MIN,
        .src_factor_rgb = SG_BLENDFACTOR_SRC_ALPHA,
        .dst_factor_rgb = SG_BLENDFACTOR_ONE_MINUS_SRC_ALPHA,
    },
});
```

...I patched up your sample code in a merge branch btw, no need for you to do any changes.

floooh:
...now just testing on D3D11 and then I'm ready to merge...

PR #1216, opened 2025-03-06
bindgen: add option to use tiger style for zig

adamrt:
This commit adds an option to generate function names using `snake_case` instead of `camel_case`, per Tiger Style.

I don't mind maintaining this in my own fork if this is too niche for your taste, but figured I would offer it up. Thanks for all your work!

-- Commit Message:

Tiger Style comes from TigerBeetle's Zig style guide that they used for their popular database (same name).

"Use snake_case for function, variable, and file names. The underscore is the closest thing we have as programmers to a space, and helps to separate words and encourage descriptive names."

https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md


floooh:
Thanks for the PR! On one hand I'm not a fan of the Zig ecosystem already developing different naming convention so early in its life, OTH the changes are not very intrusive. I'll need to think about it for a day or two :)

PR #1237, opened 2025-03-30
Expose X11 Window and Display

bullno1:
The handle of X11 Window and Display should be exposed to the end user like every other platform.

This allows doing platform-specific operations such as resizing or setting attributes.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -12388,7 +12388,7 @@ SOKOL_API_IMPL int sapp_gl_get_minor_version(void) {
 }
 
 SOKOL_API_IMPL const void* sapp_x11_get_window(void) {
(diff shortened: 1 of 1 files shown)


floooh:
Thanks!

I think the check for `_SAPP_GLX` should be removed though. This define is used to differentiate between GLX vs EGL for the GL context creation, but doesn't affect the validity of of `_sapp.x11.window` or `_sapp.x11.display`.

Did you run into any trouble before adding that check?

floooh:
...I'll just go on and remove the `_SAPP_GLX` during merge. If you stumbled over an issue with that we can still fix it later :)

floooh:
And merged! I removed the _SAPP_GLX check, and added a changelog blurb.

Many thanks!

PR #1238, opened 2025-03-30
Fix memory leak when doing dragndrop in X11

bullno1:


floooh:
...and merged. Many thanks!

I did a minor coding-style fix and added a changelog blurb.

Now looking into that other PR, I think I will just remove that _SAPP_GLX check (see my comment there).

bullno1:
> Good catch, thanks! How did you stumble over the issue, did you run with a mem-leak detection library or tool?

I always do `-fsanitize=address,undefined` during development and it just shows up.

PR #1240, opened 2025-04-03
fix: mismatch function with cimgui

n67094:
The generated binding name for Cimgui has apparently changed. This update should resolve the mismatch issue.

### DIFF
--- a/util/sokol_imgui.h
+++ b/util/sokol_imgui.h
@@ -2380,7 +2380,7 @@ SOKOL_API_IMPL void simgui_setup(const simgui_desc_t* desc) {
         #if defined(__cplusplus)
             ImGuiPlatformIO* pio = &ImGui::GetPlatformIO();
         #else
-            ImGuiPlatformIO* pio = _SIMGUI_CFUNC(GetPlatformIO)();
+            ImGuiPlatformIO* pio = _SIMGUI_CFUNC(GetPlatformIO_Nil)();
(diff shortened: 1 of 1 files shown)


floooh:
Hmm, looks like the original cimgui made that change (https://github.com/cimgui/cimgui/blob/205107640d70aeffc9cd37a1e7a8d240708a55e5/cimgui.h#L3637), but the new Dear Bindings didn't (which I switched over to via https://github.com/floooh/dcimgui).

So far the two bindings seem to have been mostly compatible, at least when using the `ig` prefix for the Dear Bindings generation, but now it looks there are bigger incompatibilities.

Going forward I would recommend switching over to the Dear Bindings (https://github.com/dearimgui/dear_bindings), since transparently supporting two incompatible C bindings will be too much trouble.

floooh:
PS: I'll see if I can figure out whether the original cimgui or the Dear Bindings are included, if it's only one or a handful functions that differ then I guess sokol_imgui.h can be kept compatible (much harder for sokol_gfx_imgui.h though and probably not worth it there).

floooh:
Ok, see my 'hack' here: https://github.com/floooh/sokol/commit/fc82d9b086f7d28dbc7f14841b4e51e26e072884

Can you check if that works for you? I'll probably not add the 'old' cimgui.h back into the CI tests to ensure compatibility.

PR #1247, opened 2025-04-11
sokol_gl.h: set default write_mask to RGBA

jrachele:
Setting the write_mask from RGB to RGBA allows for using sokol_gl to draw to a render target (when creating a pipeline with make_context_pipeline), and retain alpha values. 

Without this change,  render textures written to with sokol_gl will truncate the alpha information, and will appear invisible when blending is enabled.

### DIFF
--- a/util/sokol_gl.h
+++ b/util/sokol_gl.h
@@ -3148,7 +3148,7 @@ static void _sgl_init_pipeline(sgl_pipeline pip_id, const sg_pipeline_desc* in_d
     }
     desc.colors[0].pixel_format = ctx_desc->color_format;
     if (desc.colors[0].write_mask == _SG_COLORMASK_DEFAULT) {
-        desc.colors[0].write_mask = SG_COLORMASK_RGB;
+        desc.colors[0].write_mask = SG_COLORMASK_RGBA;
     }
 
     _sgl_pipeline_t* pip = _sgl_lookup_pipeline(pip_id.id);


floooh:
I don't quite remember why the default color mask is RGB but I'm hesitant changing the default.

In general, if you need render state that differs from the default, it's best to create create your own pipeline objects via `sgl_make_pipeline()` or `sgl_make_context_pipeline()` (but it sounds like you're alreading doing that, is there a reason why you can't set the color mask RGBA so that the default RGB isn't used?).

If you just need 1.0 in your render target's alpha channel, and assuming you're clearing the render target before rendering, it would be more efficient to just set the clear color alpha value to 1.0 though.

jrachele:
That is true, setting the write_mask to RGBA for my offscreen renderer works fine as well, but I figured that the default ought to retain alpha information if not specified, as it can be difficult to understand why the texture is drawing invisible, provided you have standard blending enabled on the render pipeline, which is the default. 

In my case I am actually not clearing the render target before rendering, I understand this is a bit unorthodox, but I do this to avoid redrawing the entire texture each frame, and instead use LOAD to keep the same texture data (I am dealing with a lot of quads for a very large 2D tilemapped world). Since the default writes it as RGB, the alpha values are truncated, and only new data that I write shows up for exactly one frame.

Considering the default pixel_format of an sg_Image is SG_PIXELFORMAT_RGBA8, I feel like it makes more sense here for the write_mask to include the alpha channel by default, but I understand being hesitant to change it as well

floooh:
> but I figured that the default ought to retain alpha information if not specified, as it can be difficult to understand why the texture is drawing invisible, provided you have standard blending enabled on the render pipeline, which is the default.

...I've also seen the 'opposite confusion' though, for instance on WebGL the canvas is alpha-blended with the web page background, and this can be confusing if your alpha-blended drawing modifies the render target alpha channel, e.g. you'd want the alpha-blending to happen between the src texture and the render target content, but not affect the composition of the render target with the webpage background, and for this case the default settings are 'safe' (e.g. default clear color in begin-pass clears the alpha to 1, and the rendering is guaranteed to not change that value).

The same could be achieved of course by setting the write mask to RGBA and the alpha blend factors to src=ZERO and dst=ONE (e.g. it will write the original alpha value of the render target back to the render target - but this would also require to create a separate pipeline object to change the default behaviour if you actually want to have the output alpha value of the fragment shader to be written into the render target).

E.g. I've seen a lot of confusion about the RGB and Alpha blend factors (often setting srcAlpha and dstAlpha to the same blend factors as srcRgb and dstRgb - which is almost never what you'd actually want), and then wondering why the page background (in WebGL) peeks through in places where they were rendering alpha-blended pixels.

jrachele:
Fair enough! That makes sense to me. I will go ahead and close the PR. Thanks for taking a look at it

PR #1259, opened 2025-05-02
Replace GetIO with GetIO_Nil in sokol_imgui.h

areynaldo:
I'm new to contributing to open source, so please feel free to correct me if needed.

The `sokol_imgui.h` file still uses `igGetIO`, but cimgui replaced it with `GetIO_Nil` about two months ago, which causes linker errors during the build.

This PR updates `sokol_imgui.h` to use `GetIO_Nil` instead of `igGetIO` to resolve the issue.

### DIFF
--- a/util/sokol_imgui.h
+++ b/util/sokol_imgui.h
@@ -2365,7 +2365,7 @@ SOKOL_API_IMPL void simgui_setup(const simgui_desc_t* desc) {
     #else
         _SIMGUI_CFUNC(CreateContext)(NULL);
         _SIMGUI_CFUNC(StyleColorsDark)(_SIMGUI_CFUNC(GetStyle)());
-        ImGuiIO* io = _SIMGUI_CFUNC(GetIO)();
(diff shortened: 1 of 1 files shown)


floooh:
Similar issue as this one: https://github.com/floooh/sokol/pull/1240, it's an incompatibility between the 'old' cimgui bindings and the new DearBindings generated C bindings which I'm maintaining via https://github.com/floooh/dcimgui, until quite recently the new and old bindings were mostly compatible when using the `ig` prefix for the new bindings, but now it looks like the old cimgui bindings are starting to diverge.

For the `GetPlatformIO()` function I fixed it like this to preserve compatibility with bot binding flaviours, but for the GetIO() function this is starting to become too much hassle:

https://github.com/floooh/sokol/commit/fc82d9b086f7d28dbc7f14841b4e51e26e072884

Is it an option for you to switch to the bindings provided by the dcimgui repository?

PR #1262, opened 2025-05-07
Fix missalign sokol app

roig:
This fix solves the issue I was having about a missalignment crash when dereferencing raw_input_data.
I don't know why you used an uint8_t array so maybe I'm missing something here.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2053,7 +2053,6 @@ inline void sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
 #include <stdlib.h> // malloc, free
 #include <string.h> // memset, strncmp
 #include <stddef.h> // size_t
-#include <stdalign.h> //alignas
 #include <math.h>   // roundf
 
 // helper macros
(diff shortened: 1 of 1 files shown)


floooh:
> I don't know why you used an uint8_t array so maybe I'm missing something here.

IIRC the problem is that `GetRawInputData()` may return additional data for 'non-standard input devices'.

I just checked GLFW, and this first pools for the required size, and then does a heap allocation.

I'll merge your fix so that it unblocks you, but will then go ahead and implement GLFW's solution of heap-allocating the data on demand, since this seems like a more robust solution:

```c
            GetRawInputData(ri, RID_INPUT, NULL, &size, sizeof(RAWINPUTHEADER));
            if (size > (UINT) _glfw.win32.rawInputSize)
            {
                _glfw_free(_glfw.win32.rawInput);
                _glfw.win32.rawInput = _glfw_calloc(size, 1);
                _glfw.win32.rawInputSize = size;
            }

            size = _glfw.win32.rawInputSize;
            if (GetRawInputData(ri, RID_INPUT,
                                _glfw.win32.rawInput, &size,
                                sizeof(RAWINPUTHEADER)) == (UINT) -1)
            {
                _glfwInputError(GLFW_PLATFORM_ERROR,
                                "Win32: Failed to retrieve raw input data");
                break;
            
```

Thanks for reporting the issue and the PR!

floooh:
PS: ignore the breaking CI, that's the sokol-d CI test, tends to break from time to time.

PR #1265, opened 2025-05-14
Set activation policy before making window fullscreen

danielchasehooper:
On macOS 15.4.1 `sapp_desc.fullscreen = true` didn't work because `NSWindow -toggleFullScreen:` doesn't seem to work when the `NSApp.activationPolicy` is still `NSApplicationActivationPolicyProhibited`. Setting the activationPolicy earlier fixes the issue. 

(noticed the bug on macOS 15.4.1 (24E263), with SOKOL_METAL, in a command line app, not an app bundle)

Side note: this exposed a bug in `sapp_is_fullscreen` where its internal state could differ from the real window state. To eliminate the possibility of that bug on macOS it could be implemented as 
```c
SOKOL_API_IMPL bool sapp_is_fullscreen(void) {
    #if defined(_SAPP_MACOS)
    return (_sapp.macos.window.styleMask & NSWindowStyleMaskFullScreen) != 0;
    #else
    return _sapp.fullscreen;
    #endif
}
```

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -4097,11 +4097,11 @@ _SOKOL_PRIVATE void _sapp_macos_frame(void) {
     #endif
     [_sapp.macos.window center];
     _sapp.valid = true;
+    NSApp.activationPolicy = NSApplicationActivationPolicyRegular;
     if (_sapp.fullscreen) {
         /* ^^^ on GL, this already toggles a rendered frame, so set the valid flag before */
         [_sapp.macos.window toggleFullScreen:self];
     }
-    NSApp.activationPolicy = NSApplicationActivationPolicyRegular;
     [NSApp activateIgnoringOtherApps:YES];
     [_sapp.macos.window makeKeyAndOrderFront:nil];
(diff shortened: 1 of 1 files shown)


floooh:
Hi @danielchasehooper I'm currently travelling, will be back on the weekend and check your prs and tickets :)

danielchasehooper:
I figured, hope you have a good trip! Thanks for letting me know and sorry for spamming you while traveling

floooh:
I actually stumbled over this a couple of days ago (e.g. fullscreen only works when building as a macOS bundle).

I'll take care of your PRs after the sokol-gfx `compute-ms2` merge (which hopefully happens tomorrow evening).

floooh:
Change of plans, looking into this today. Since the sokol-merges need to be done in one go it's better if I move those to the weekend.

floooh:
Ok, working nicely in pacman.c (which is built as a cmdline tool on macOS).

floooh:
...and merged. Thanks! I also added a short blurb to the changelog.

PR #1266, opened 2025-05-14
Support Alpha blending in sokol_gl.h

danielchasehooper:
sokol_gl's api accepts colors with alpha, so blending should probably be enabled

### DIFF
--- a/util/sokol_gl.h
+++ b/util/sokol_gl.h
@@ -3331,6 +3331,9 @@ static void _sgl_init_context(sgl_context ctx_id, const sgl_context_desc_t* in_d
     sg_pipeline_desc def_pip_desc;
     _sgl_clear(&def_pip_desc, sizeof(def_pip_desc));
     def_pip_desc.depth.write_enabled = true;
+    def_pip_desc.colors[0].blend.enabled = true;
+    def_pip_desc.colors[0].blend.src_factor_rgb = SG_BLENDFACTOR_SRC_ALPHA;
+    def_pip_desc.colors[0].blend.dst_factor_rgb = SG_BLENDFACTOR_ONE_MINUS_SRC_ALPHA;
     ctx->def_pip = _sgl_make_pipeline(&def_pip_desc, &ctx->desc);
     if (!sg_add_commit_listener(_sgl_make_commit_listener(ctx))) {
         _SGL_ERROR(ADD_COMMIT_LISTENER_FAILED);


util/sokol_gl.h
```diff
@@ -3331,6 +3331,9 @@ static void _sgl_init_context(sgl_context ctx_id, const sgl_context_desc_t* in_d
     sg_pipeline_desc def_pip_desc;
     _sgl_clear(&def_pip_desc, sizeof(def_pip_desc));
     def_pip_desc.depth.write_enabled = true;
```
floooh, line 3333: It probably makes sense to remove this `depth.write_enabled = true` line when alpha-blending is enabled by default, and also remove this code in `_sgl_init_pipeline()` since it wouldn't have any purpose then:

https://github.com/floooh/sokol/blob/202f41b1ce892e1a3fcd25b5c9497a5871ac57a3/util/sokol_gl.h#L3146-L3148

The rest of the default pipeline state is configured for 2D rendering anyway, which IMHO makes it ok to switch the default pipeline state to alpha blended (e.g. the common setup for rendering alpha-blended 2D sprites).

floooh (changes requested):

util/sokol_gl.h
KTRosenberg: Is there still a plan to enable alpha blending for the gl-style API? It seems needed for most 2D applications.

floooh:
> Is there still a plan to enable alpha blending for the gl-style API? It seems needed for most 2D applications.

Tbh I'm going back and forth on this (otherwise I'd have this merged already). I (sometimes) no longer think that enabling blending by default is the right way, and the alternatives all explode the internal number of render pipeline one way or another.

The user already has control over alpha blending via `sgl_push_pipeline()`, it's a bit more hassle but more flexible than the alternatives.

KTRosenberg:
Ah, I see. I think it then might be useful to have a couple default pipelines named after the use cases (maybe)? Regular back to front for normal sprites and additive for particle systems are probably the most useful. I was thinking of introducing this api to beginners so I wasn’t that comfortable with the idea of having them create pipelines.

KTRosenberg:
More specifically, I think it would be nice to have a few more built-ins that could be pushed for a command. Then you wouldn't have to interact with shaders to have basic sprites going. (I'm actually designing a few lessons on 2D games and am considering using this gl-esque API for that.). `sgl_enable_blending_premultiply_alpha(true)`, `sgl_enable_blending_additive(true)`, or something like this. Built-ins would be nice to avoid more compilation steps.

floooh:
That wouldn't change much unfortunately with the internal state explosion (basically for each state variation, sokol_gl.h needs to create 5 pipeline objects (one for each primitive type), so with two builtin alpha blending variations we would already be at 15 pipeline objects.

At that point it might make sense to first rewrite the internal pipeline handling code to a 'hash-and-cache' approach (e.g. create pipeline objects on demand instead of all pipeline objects upfront - which tbh generally might make sense, even with the current system).

KTRosenberg:
I can understand that. Is there an example for how to get this use case working for a couple of the pipelines that I need? I was hoping to reuse whatever is internal to sokol since I’m less familiar with it at this point. Also, I don’t want to clutter this issue, so is there’s a better avenue for support? I’d really like to get a higher-level api working for sprite graphics.

floooh:
(argh, I had pressed Save to early... editing...

> Also, I don’t want to clutter this issue, so is there’s a better avenue for support?

Issues are the right place currently for such discussions :)

> Is there an example for how to get this use case working for a couple of the pipelines that I need?

The basisu-sapp sample has an example for a regular alpha blending pipeline.

How to create one:

```c
    // a sokol-gl pipeline object for alpha-blended rendering
    state.alpha_pip = sgl_make_pipeline(&(sg_pipeline_desc){
        .colors[0] = {
            .write_mask = SG_COLORMASK_RGB,
            .blend = {
                .enabled = true,
                .src_factor_rgb = SG_BLENDFACTOR_SRC_ALPHA,
                .dst_factor_rgb = SG_BLENDFACTOR_ONE_MINUS_SRC_ALPHA,
            },
        }
    });
```
https://github.com/floooh/sokol-samples/blob/8afa83928ce1870efeb0d513e7c4dce4f5db7b3e/sapp/basisu-sapp.c#L87-L97

...and then to 'load' this pipeline call `sgl_load_pipeline(pip)` before an `sgl_begin/sgl_end`, like here:

```c
static void draw_quad(quad_params_t params) {
    sgl_texture(params.view, state.smp);
    if (params.pip.id) {
        sgl_load_pipeline(params.pip);
    }
    else {
        sgl_load_default_pipeline();
    }
    sgl_push_matrix();
    sgl_translate(params.pos.x, params.pos.y, 0.0f);
    sgl_scale(params.scale.x, params.scale.y, 0.0f);
    sgl_rotate(params.rot, 0.0f, 0.0f, 1.0f);
    sgl_begin_quads();
    sgl_v2f_t2f(-1.0f, -1.0f, 0.0f, 0.0f);
    sgl_v2f_t2f(+1.0f, -1.0f, 1.0f, 0.0f);
    sgl_v2f_t2f(+1.0f, +1.0f, 1.0f, 1.0f);
    sgl_v2f_t2f(-1.0f, +1.0f, 0.0f, 1.0f);
    sgl_end();
    sgl_pop_matrix();
}
```
https://github.com/floooh/sokol-samples/blob/8afa83928ce1870efeb0d513e7c4dce4f5db7b3e/sapp/basisu-sapp.c#L109-L126

...there's also a 'pipeline stack' similar to the 'matrix stack' which can be manipulated vis `sgl_push_pipeline()` and `sgl_pop_pipeline()`.

floooh:
Finally... there's also the concept of 'sokol header extensions'... e.g. you can write your own header which basically looks like this:

`sokol_gl_ext.h`:
```c
#pragma once
#include "sokol_gl.h"

// your own sgl extension functions declaration
void sgl_my_extension_func(void);

#if defined(SOKOL_GL_IMPL)

// ...and the implementation:
void sgl_my_extension_func(void) {
    // implementation code here
}
#endif 
```

The implementation block sees the private implementation code and data of sokol_gl.h (but tbh, for this specific problem you don't even need access to that, the public sokol_gl.h API should be enough).

KTRosenberg:
Thanks. By the way, is there a reason not to use premultiplied alpha in this case, if it’s related to how sokol configures things?

floooh:
> is there a reason not to use premultiplied alpha in this case

It depends what your input textures look like (e.g. the rgb values should already be premultiplied). The default pixel shader basically multiplies the texture color with the vertex color and passes this back (e.g. there's no adhoc-premultiply happening in the pixel shader):

https://github.com/floooh/sokol/blob/2e75443dbd4940b5aa8d76a8e479f8e4b270b9a3/shdgen/sokol_gl.glsl#L22-L31

...you *can* inject your own pixel shader with the `sg_pipeline_desc` though, and as long as the shader has the same interface as the builtin shader that should work.

KTRosenberg:
I see. I'd probably just import regular png files though like in the sample you provide, so maybe it's fine. (EDIT: actuallty pngs don't have premultiplied values do they, so I'd need to do that myself?) For custom shaders, I suspect I'd need to compile for all platforms in advance? I was trying to work-out what the workflow for this might be. I'd be open to just using your sokol glsl dialect if there's a runtime compilation step I missed that would handle all of this for me on whichever platform.

floooh:
For cross-platform shaders, there's no runtime compilation. An integrated shader compiler would bloat the code too much (by about 10 MBytes) and pull in a lot of really large and ugly C++ dependencies (glslang, SPIRVTools, SPIRVCross, Tint).

There's ways to integrate sokol-shdc into the build process though so that working on shader code feels like regular source code, but that mostly only makes sense when the shader code changes a lot, otherwise just compiling manually and including the resulting C header file (or other-language source files) into the build.

KTRosenberg:
Thanks again for clarifying. That makes sense that you’d probably want to keep that separation. So from what I can tell, for my specific case, I’m trying implement premultiplied alpha blending, most likely in the shader. Your examples from before work for the blend state, but I’d like to modify the existing sokol gl shaders to multiply the final color output with its own alpha to do premultiplied alpha in the shader instead of on the CPU. What would be the easiest way to clone the built in pipeline to support this one change? I suppose just run the shader compiler on the original and push it at start?
https://github.com/floooh/sokol/blob/master/shdgen/sokol_gl.glsl
It would just be color *= color.a.

floooh:
> I suppose just run the shader compiler on the original and push it at start?

Yeah, but it's a bit of a hassle to copy-paste the output into sokol_gl.h.

There's a helper script here which automates the whole thing: https://github.com/floooh/sokol/blob/master/shdgen/shdgen.ts, but that's only for my own use (must be run on macOS and will attempt to run a CI pipeline on the sokol Github repo which will fail for anybody else than me). But you could maybe modify that helper script to do what you want.

KTRosenberg:
I thought I could just push my own shader at the start instead of aiming to replace the one embedded. Is that not possible for sokol_gl? Maybe the problem I’m not seeing is that I might have to redo all the bindings and setup myself.

I’m on macos, so that helps.
Maybe this is an argument in favor of having a nicer way to modify the builtin-shader and keep all the same data inputs? I think the gl-ish api is useful enough to warrant that.

Also, I figure at this point I could start a new issue since I don’t want to clutter this one.
Thanks again for the support.

PR #1284, opened 2025-06-13
Update C3 language to version 0.7.2

radekm:
Update syntax of C3 binding from 0.6.* to 0.7.*. There's also a pull request https://github.com/floooh/sokol-c3/pull/3 to sokol-c3 repo.


floooh:
Thanks!

PR #1288, opened 2025-06-20
sokol_audio.h: remove extra semicolons

fuhsnn:
These trigger `-Werror,-Wextra-semi` with clang 19 when building `tests/CMakeLists.txt`:

```
In file included from /home/user/sokol/tests/compile/sokol_audio.c:2:
/home/user/sokol/tests/compile/../../sokol_audio.h:1483:2: error: extra ';' outside of a function [-Werror,-Wextra-semi]
 1483 | };
      |  ^
/home/user/sokol/tests/compile/../../sokol_audio.h:1492:2: error: extra ';' outside of a function [-Werror,-Wextra-semi]
 1492 | };
      |  ^
```

### DIFF
--- a/sokol_audio.h
+++ b/sokol_audio.h
@@ -1480,7 +1480,7 @@ error:
         _saudio.backend.device = 0;
     }
     return false;
-};
+}
 
 _SOKOL_PRIVATE void _saudio_alsa_backend_shutdown(void) {
     SOKOL_ASSERT(_saudio.backend.device);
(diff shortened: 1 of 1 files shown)


floooh:
Thanks! I wonder why GCC doesn't complain about these during CI builds... there is some special case for GCC but it should at least complain in C++ mode:

https://github.com/floooh/sokol/blob/07a022ec7b24a7983217199c63fa894ac6397733/tests/CMakeLists.txt#L53-L57

...guess I should also try building with Clang in the Linux CI builds...

Anyway. Thanks a lot for catching that!

Issue #1252, opened 2025-04-18
Planning: Introduce a view object type


floooh:
Just to get this out of my head for now:

Currently sokol-gfx implicitely manages resource-view objects in some backend APIs, this often simplifies the API but limits flexibility, and especially with the new compute shader resources those limitations are becoming more painful (for instance in D3D11 an offset into a storage buffer is baked into the UAV object, while the offset for index- and vertex-buffer bindings are fully dynamic).

While at it:

- generalize cube map faces to 'slice' everywhere in the API (esp `sg_image_data`)
- get rid of baked `sg_attachments` objects, instead `sg_attachments` becomes a transient struct of views, just like `sg_bindings`

```c
// storage buffer view
const sg_view b0 = sg_make_view(&(sg_view_desc){
    .storage_buffer = {
        .resource = buf,
        .offset = 1234,
    },
});

// regular texture view
const sg_view b1 = sg_make_view(&(sg_view_desc){
    .texture = {
        .image = img,
        .sampler = smp,  // for validation purposes only, samplers must still be separately bound
        // FIXME: check if the following 'reinterpretation cast' would work across all backend APIs
        // (WebGL2 is probably a no-go)
        .dimension = SG_IMAGEDIMENSION_1D/2D/ARRAY/CUBEMAP, // renamed SG_IMAGETYPE_x
        .pixel_format = ..., // optional
        .mip_levels = { .base = ..., .num = ... }, // optional (is this possible in GL?)
        .slices = { .base = ..., .num = ...}, // optional       
    },
});

// render attachment view
const sg_binding b2 = sg_make_binding(&(sg_binding_desc){
    .attachment = {
        .image = img,
        .pixel_format = ..., // optional (also: WebGL2???)
        .mip_level = ..., // optional
        .slice = ..., // optional
    },
});

// a storage texture binding
const sg_binding b3 = sg_make_binding(&sg_binding_desc){
    .storage_image = {
        .image = img,
        .pixel_format = ..., // optional
        .mip_level = ..., // optional
        .slice = ..., // optional
    },
});
```

Q: should there also be vertex- and index-buffer views? Technically not needed, but it would make the `sg_bindings` struct smaller since the offsets would be baked into the views (and they rarely need to be fully dynamic anyway).

```c
// vertex buffer binding
const sg_binding b0 = sg_make_binding(&(sg_binding_desc){
    .vertex_buffer = {
        .buffer = buf,
        .offset = 1234,
    },
});

// index buffer binding
[...]

PR #1290, opened 2025-06-25
sokol_gfx.h: fix sg_query_attachments_desc not filling in storage attachments

kcbanner:
I think this might be made redundant by https://github.com/floooh/sokol/issues/1252, so it may not be worth merging.


floooh:
Yeah, tbh I don't remember if I just forgot that part or skipped it because it will be obsolete anyway.

Unless you need that data I would prefer skipping the merge (let me know).

floooh:
PS: ignore the CI errors, looks like they switched to a more recent MSVC version which doesn't play well with Nuklear.

floooh:
Ok, I'll close this PR :)

kcbanner:
All good! I figured I would post this in case anyone needed the fix in the mean time.

PR #1293, opened 2025-06-27
sokol_gfx.h: allow #include directive for d3d11 shaders

rdunnington:
This PR adds some minor plumbing to allow HLSL shaders to #include other HLSL files, using the default D3D11 include handler.

This change is motivated by my use-case, where I'm writing HLSL directly and avoid the `sokol-shdc` workflow so that I can make direct changes to HLSL files and immediately hot-reload them. I'm getting to the point in my project where I have enough shaders that it's starting to make sense to share functionality and it would be nice to not have to copy+paste between files all the time.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -3443,6 +3443,8 @@ typedef struct sg_sampler_desc {
     source code, you can provide an optional target string via
     sg_shader_stage_desc.d3d11_target, the default target is "vs_4_0" for the
     vertex shader stage and "ps_4_0" for the pixel shader stage.
+    You may optionally provide the file path to enable the default #include handler
+    behavior when compiling source code.
 */
 typedef enum sg_shader_stage {
     SG_SHADERSTAGE_NONE,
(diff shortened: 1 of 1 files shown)


floooh:
Looks like a good fix, mostly because it is so simple :)

I'll try to take care of the merge tomorrow.

floooh:
Please ignore the CI problem, the latest MSVC clashes with Nuklear. I just fixed that (weird though that the CI didn't pick up that change, let me check...)

rdunnington:
Thank you! My project is also using zig, once this goes in I can open a PR to update the zig bindings as well. (assuming those are updating manually)

floooh:
...ok I forgot to fix the Nuklear vs MSVC issues in the sokol repository master branch. Don't worry about the CI, I'll take care of everything during the merge.

floooh:
> (assuming those are updating manually)

Nope, all automated ;) The merge commit will trigger a CI pipeline to update the bindings.

rdunnington:
Very slick. Thanks!

floooh:
Ok merged. Thanks!

floooh:
PS: the zig bindings will be uptodate when this pipeline has finished:

https://github.com/floooh/sokol/actions/runs/15943794814

PR #1296, opened 2025-06-30
gen_d: Add sokol_nuklear and sokol_args support

kassane:
### Sync
- https://github.com/kassane/sokol-d/pull/80

### DIFF
--- a/.github/workflows/gen_bindings.yml
+++ b/.github/workflows/gen_bindings.yml
@@ -61,7 +61,6 @@ jobs:
       - uses: actions/checkout@main
         with:
           repository: kassane/sokol-d
-          ref: nuklear-example
           path: bindgen/sokol-d
       - uses: actions/checkout@main
         with:
@@ -323,6 +322,7 @@ jobs:
           dub build :instancingcompute
           dub build :mrt
           dub build :noninterleaved
+          dub build :nuklear
           dub build :offscreen
           dub build :quad
           dub build :shapes
--- a/.github/workflows/gen_bindings.yml
+++ b/.github/workflows/gen_bindings.yml
@@ -61,6 +61,7 @@ jobs:
       - uses: actions/checkout@main
         with:
           repository: kassane/sokol-d
+          ref: nuklear-example
           path: bindgen/sokol-d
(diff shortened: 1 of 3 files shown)


kassane:
Here, done.
cc: @floooh

floooh:
I'll be travelling until Thursday, will check when I'm back!

floooh:
Did some testing and there seems to be a chicken-egg problem with the nuklear.h header. What I did:

1. in `sokol/bindgen/sokol-d` switch to the `nuklear-example` branch to get the required changes for the bindings
2. then in `sokol/bindgen` run `python3 gen_all.py`

Step (2) fails because `sokol_nuklear.c` doesn't find `nuklear.h` when clang ast-dump is called:

```
sokol-d/src/sokol/c/sokol_nuklear.c:15:10: fatal error: 'nuklear.h' file not found
   15 | #include "nuklear.h"
      |          ^~~~~~~~~~~
1 error generated.
```

The difference to sokol_imgui.h is that sokol_nuklear.h currently needs to include nuklear.h not only when building the implementation but also when only including the declaration, because of the types `nk_handle`, `nk_flags`, `nk_context` and `nk_plugin_filter` in the public API.

I currently don't have a great idea how to fix this (ideally the sokol_nuklear.h API shouldn't have any Nuklear types, but at least for that `snk_edit_string()` function that's tricky (not sure how useful that is though, since it only calls `sapp_show/hide_keyboard()` and that isn't supported on most platforms anyway).

...OTH, the sokol repo even includes a copy of `nuklear.h` (as `tests/ext/nuklear.h`), so one option could be to just copy this into the D bindings directory before running clang.

kassane:
Maybe this hack solve: https://github.com/floooh/sokol/pull/1296/commits/ad15fcd5c28acae9e97061d510a5d87dba8c692b
in my local-test works!

**Edit**: check https://github.com/floooh/sokol/pull/1296/commits/22a75482433de85492cd7c7d63a69671b1c86108

kassane:
I think that's a good way to go. It could be improved by adding support for other languages later.

floooh:
...and merged. Thanks!

PR #1300, opened 2025-07-25
sokol_gfx.h: Adapt SG_MAX_COLOR_ATTACHMENTS per backend.

TristanWellman:
I noticed when using Sokol that the maximums for a lot of things were limited to the web capabilities, which is fine for most applications, but if you want to use more unique buffers outside the usual color, position, normal, etc. You need more than 4 color buffers which is supported in all the other OpenGL, DX, and Metal backends.

I added a small check to see if it is GLES/WGPU, I did not research to see limits on the other buffer objects and samplers, but I would assume SG_MAX_STORAGE_ATTACHMENTS can also be 8 for the other backends.

I am using OpenGL 4.1 and tested it on MacOS 26, Arch Linux, and Windows 11 23H2. It should be tested with the other backends.

Thanks!

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -1965,7 +1965,11 @@ typedef struct sg_range {
 enum {
     SG_INVALID_ID = 0,
     SG_NUM_INFLIGHT_FRAMES = 2,
+#if defined(SOKOL_GLES3) || defined(SOKOL_WGPU)
     SG_MAX_COLOR_ATTACHMENTS = 4,
+#else
+    SG_MAX_COLOR_ATTACHMENTS = 8,
+#endif
     SG_MAX_STORAGE_ATTACHMENTS = 4,
     SG_MAX_UNIFORMBLOCK_MEMBERS = 16,
     SG_MAX_VERTEX_ATTRIBUTES = 16,


floooh:
Thanks for the PR, but I unfortunately need to close it :)

The resource-view update I'm currently working on will get rid of the idea of storage attachments (and pre-baked attachment objects alltogether).

At first there will still be the same restriction though (the constant will just be called `SG_MAX_STORAGEIMAGE_BINDINGS_PER_STAGE`) but I hope to turn this into a runtime-limit in a follow up update (along with all other "max-per-stage-bindings" for the other resource types.

floooh:
PS: there's also a sort-of unwritten rule that the public API should have the same 'shape' no matter what backend is selected (I'm currently actually not sure if that's guaranteed though), but the idea is that you can compile sokol into different DLLs (one per backend), and then select at runtime which backend-DLL to load, and this will only work when no backend-specific constants are baked into the executable which loads the DLLs.

TristanWellman:
Ah that's great, Can't wait to see it!

floooh:
This is now handled via https://github.com/floooh/sokol/pull/1330

PR #1301, opened 2025-07-25
remove references to sg_begin_default_pass

stianhj:
Was looking at some old code I had and noticed the sg_begin_default_pass had been [removed](https://floooh.github.io/2024/02/26/sokol-spring-cleaning-2024.html#detailed-change-list), but there are still a few references to it in the docs.


floooh:
Good catch! Thanks a alot!

Issue #1269, opened 2025-05-21
Idea for improving sokol gfx library


roig:
Hello!

Just wanted to share with you an idea to improve sokol library. 
I know that maybe it will be a big breaking change but I think it's worth at least to think about it.

Given that it has now support for OpenGL 4.3, there are two functions that can ease the setup of bindings and also reduce gl attribute calls sent to the gpu:
glDrawArraysInstancedBaseInstance and glDrawElementsInstancedBaseVertexBaseInstance

I implemented these like this:
```cpp
typedef struct {
    unsigned int count;
    unsigned int instance_count;
    unsigned int first;
    unsigned int base_instance;
} DrawArraysIndirectCommand;

typedef struct {
    unsigned int count;
    unsigned int instance_count;
    unsigned int first_index;
    int base_vertex;
    unsigned int base_instance;
} DrawElementsIndirectCommand;

SOKOL_API_IMPL void sg_ext_gl_draw_arrays_indirect(DrawArraysIndirectCommand cmd) {
    #ifdef SOKOL_GLCORE
    const GLenum p_type = _sg.gl.cache.cur_primitive_type;
    glDrawArraysInstancedBaseInstance(p_type, cmd.first, cmd.count, cmd.instance_count, cmd.base_instance);
    _sg_stats_add(num_draw, 1);
    //int base_element, int num_elements, int num_instances,
    _SG_TRACE_ARGS(draw, cmd.first,cmd.count,cmd.instance_count);
    #else
    SOKOL_ASSERT(false);
    #endif
}

SOKOL_API_IMPL void sg_ext_gl_draw_elements_indirect(DrawElementsIndirectCommand cmd) {
    #ifdef SOKOL_GLCORE
    const GLenum i_type = _sg.gl.cache.cur_index_type;
    const GLenum p_type = _sg.gl.cache.cur_primitive_type;
    const int i_size = (i_type == GL_UNSIGNED_SHORT) ? 2 : 4;
    GLvoid* indices = (GLvoid*)(GLintptr)(cmd.first_index * i_size);
    glDrawElementsInstancedBaseVertexBaseInstance(p_type, cmd.count, i_type, indices, cmd.instance_count, cmd.base_vertex, cmd.base_instance);
    _sg_stats_add(num_draw, 1);
    //int base_element, int num_elements, int num_instances,
    _SG_TRACE_ARGS(draw, cmd.first_index,cmd.count,cmd.instance_count);
    #else
    SOKOL_ASSERT(false);
[...]

PR #1307, opened 2025-08-06
sokol_gfx.h: added sg_draw_from_base (#1269)

CorbinSteele:
I needed this and it seemed extremely straightforward to add. I took the liberty of making the function call `sg_draw_from_base` because it seemed less confusing than merely calling it `sg_draw_base` which makes "base" sound too much like a drawn object. Other than that, I think I covered everything suggested in the linked feature request.

### WARNING
Assume that this basically hasn't been tested at all. I don't _think_ it could break anything, unless I'm blind and somehow screwed up the Windows GL loader signatures or something.

**_IF_** it compiles on everything, then it should work without ever breaking any existing functionality.

### NOTE
I did slightly modify/reuse the internal _sg_draw function, so if anyone is ignoring the API and relying on that directly then... well, they kinda deserve what they get. I also chose not to distinguish between sg_draw_from_base and sg_draw call stats because I can't imagine why anyone would ever want them to be tracked separately.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -2155,7 +2155,8 @@ typedef struct sg_features {
     bool mrt_independent_write_mask;    // multiple-render-target rendering can use per-render-target color write masks
     bool compute;                       // storage buffers and compute shaders are supported
     bool msaa_image_bindings;           // if true, multisampled images can be bound as texture resources
-    bool separate_buffer_types;         // cannot use the same buffer for vertex and indices (onlu WebGL2)
+    bool separate_buffer_types;         // cannot use the same buffer for vertex and indices (only WebGL2)
+    bool base_instance_only_zero;       // cannot use base_vertex or base_instance in draw_from_base (only WebGL2)           
 } sg_features;
 
 /*
@@ -3819,6 +3820,7 @@ typedef struct sg_trace_hooks {
     void (*apply_bindings)(const sg_bindings* bindings, void* user_data);
     void (*apply_uniforms)(int ub_index, const sg_range* data, void* user_data);
     void (*draw)(int base_element, int num_elements, int num_instances, void* user_data);
+    void (*draw_from_base)(int base_element, int num_elements, int num_instances, int base_vertex, int base_instance, void* user_data);
     void (*dispatch)(int num_groups_x, int num_groups_y, int num_groups_z, void* user_data);
     void (*end_pass)(void* user_data);
     void (*commit)(void* user_data);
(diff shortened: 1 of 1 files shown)


floooh:
Thanks! Apart from the CI problems (from quick glance, looks like some functions are not available on macOS, e.g. only had been added after GL 4.1?) I'm not sure yet when to merge the PR (definitely not before the pending resource view update I'm currently working on), but I'll keep it 'parked' for at least when other people might want it.

CorbinSteele:
That was my main intent in submitting it, yeah.

It looks like I did mess up the wgpu call signature though: apparently base_vertex is a signed int even though every other parameter is unsigned.

I'm going to leave it unfixed because I'm now unsure whether passing a _negative_ base_vertex ever makes sense and needs to be allowed for some bizarre use case.

cloudwu:
This feature works for me too. 

My workaround is adding an uniform `baseinst` to the shader :  https://github.com/cloudwu/soluna/blob/master/src/texquad.glsl#L31

I also thought about adding an API like `sg_draw_from_base`, but I found that the lower version of opengl does not seem to support it :(

floooh:
Btw you can also use the buffer offsets in `sg_bindings` to achieve a similar effect, but of course that means you'll need to rebind between drawcalls whenever the base-offsets change - but that's what I do in sokol_imgui.h, and it doesn't require storage buffer support.

PS: hmm, GL 4.2... https://registry.khronos.org/OpenGL-Refpages/gl4/html/glDrawElementsInstancedBaseVertexBaseInstance.xhtml - but that's actually okay-ish... macOS is stuck at 4.1, but GL on macOS isn't a great idea anyway (instead Metal is recommended). And WebGL also doesn't have the function, but WebGPU is coming "really soon now". That leaves Android as unsupported, since GLES 3.2 only has `glDrawElementsInstancedBaseVertex` but not `glDrawElementsInstancedBaseVertexBaseInstance`.

cloudwu:
> Btw you can also use the buffer offsets in `sg_bindings` to achieve a similar effect, but of course that means you'll need to rebind between drawcalls whenever the base-offsets change - but that's what I do in sokol_imgui.h, and it doesn't require storage buffer support.

base vertex and base instance are different , IMO. Changing buffer offsets in `sg_bindings` can only affect  base vertex, but I need both.

https://github.com/cloudwu/soluna/blob/cc6c34f492da5dd4d370bc0ba37238fc15cfcbd1/src/material_default.c#L89-L90

btw, This may be because I've used the step function `SG_VERTEXSTEP_PER_INSTANCE`

https://github.com/cloudwu/soluna/blob/cc6c34f492da5dd4d370bc0ba37238fc15cfcbd1/src/material_default.c#L101

I wrote a [blog (in Chinese, translate into english by google translate)](https://blog-codingnow-com.translate.goog/2025/02/2d_pipeline_design.html?_x_tr_sl=auto&_x_tr_tl=en)  to record my 2d pipeline design.

floooh:
> Changing buffer offsets in sg_bindings can only affect base vertex, but I need both.

You can also define an offset into the instance data via sg_bindings.

E.g. let's say you have one buffer for the per-vertex data and one for the per-instance data, and the vertex layout is setup like in the sokol instancing sample:

```c
.layout = {
    // vertex buffer at slot 1 must step per instance
    .buffers[1].step_func = SG_VERTEXSTEP_PER_INSTANCE,
    .attrs = {
        [ATTR_instancing_pos]      = { .format=SG_VERTEXFORMAT_FLOAT3, .buffer_index=0 },
        [ATTR_instancing_color0]   = { .format=SG_VERTEXFORMAT_FLOAT4, .buffer_index=0 },
        [ATTR_instancing_inst_pos] = { .format=SG_VERTEXFORMAT_FLOAT3, .buffer_index=1 }
    }
},
```

...e.g. position and color is loaded from the vertex buffer at index 0, and the per-instance data is loaded from the vertex buffer at buffer index 1, and the buffer at index 1 uses `SG_VERTEXSTEP_PER_INSTANCE`.

The sokol instancing sample doesn't use any offsets into those buffers, so the `sg_apply_bindings` just looks like this:

```c
sg_apply_bindings(&(sg_bindings){
    .vertex_buffers = {
        [0] = buffer_with_per_vertex_data,
        [1] = buffer_with_per_instance_data
    },
    .index_buffer = buffer_with_index_data,
});
```

If you want offsets both into the vertex- and instance-data it would look like this (also for completeness: an offset into the index buffer):

```c
sg_apply_bindings(&(sg_bindings){
    .vertex_buffers = {
        [0] = buffer_with_per_vertex_data,
        [1] = buffer_with_per_instance_data
    },
    .vertex_buffer_offsets = {
        [0] = byte_offset_into_vertex_data,
        [1] = byte_offset_into_instance_data,
    },
    .index_buffer = buffer_with_index_data,
    .index_buffer_offset = byte_offset_into_instance_data,
});
```

You just have to be aware that those offsets are in number of bytes and not in number of buffer items (like in the glDrawElementsBaseVertexBaseInstance call). Functionality-wise it should be the same, it's just not quite as efficient since the buffers need to be 're-bound' between drawcalls with different offsets.

In the upcoming resource view update it will also be possible to bind storage buffers with an offset via a 'storage buffer view object' (this was a feature that's currently missing).

cloudwu:
> > Changing buffer offsets in sg_bindings can only affect base vertex, but I need both.
> 
> You can also define an offset into the instance data via sg_bindings.
 
Thanks for pointing this out. I just modified my code in this way, and the code has become simpler.

cloudwu:
> You just have to be aware that those offsets are in number of bytes and not in number of buffer items (like in the glDrawElementsBaseVertexBaseInstance call). Functionality-wise it should be the same, it's just not quite as efficient since the buffers need to be 're-bound' between drawcalls with different offsets.

Do you have any plans to support indirect draw for multiple draw commands? Let each draw command specify a base.

floooh:
> Do you have any plans to support indirect draw for multiple draw commands? Let each draw command specify a base.

Plans yes, but no time window yet (e.g. don't hold your breath). It would mirror WebGPU's draw-indirect feature and would mostly be useful when generating draw commands in a compute shader. 

https://developer.mozilla.org/en-US/docs/Web/API/GPURenderPassEncoder/drawIndirect

floooh:
NOTE: I'm closing this PR in favour of https://github.com/floooh/sokol/pull/1339 (yeah I know, `sg_draw_ex()` isn't very creative, but once the GL backends will be dropped, `sg_draw_ex()` will become the regular `sg_draw()`...)

PR #1309, opened 2025-08-10
sokol_app: support for custom mouse cursor images

Seb-degraff:
This PR adds custom cursor images for macos, windows, linux and macos.

Two new functions are added to make and destroy cursor images:

```c
/* returns a custom cursor image for use with sapp_set_mouse_cursor_image */
sapp_mouse_cursor_image sapp_make_mouse_cursor_image(sapp_image_desc* desc, int hotspot_x, int hotspot_y);
/* frees resource associated with the cursor image */
void sapp_destroy_mouse_cursor_image(sapp_mouse_cursor_image image);
```

`sapp_mouse_cursor_image` is just an opaque handle to the platform's cursor representation. I deliberately choose not to use a pool and generational handles because I think the typical use case for custom cursor images is pretty trivial – just make a bunch of cursors in setup() and optionaly destroy them in shutdown() – and it makes the cost of added code not worth it imo.

It does mean that the user should be careful not to destroy a cursor image and then pass it to sapp_set_mouse_cursor_image, of course.

To set a custom cursor, `sapp_set_mouse_cursor_image()` can be called instead of `sapp_set_mouse_cursor()`. Calling `sapp_set_mouse_cursor` will reset the cursor to the specified "system" cursor. `sapp_set_mouse_cursor_image()` and `sapp_set_mouse_cursor()` therefore work as "sibling" functions, both overriding the internal cursor state. This is the simplest api I could think of that doesn't break existing code.

Note that after calling `sapp_set_mouse_cursor_image()`, `sapp_get_mouse_cursor()` returns a new `sapp_mouse_cursor` enum value: `SAPP_MOUSECURSOR_CUSTOM_IMAGE`. This was done to simplify the tracking of the cursor stat in sokol_app, and also so external code that uses `sapp_get_mouse_cursor()` to check if `sapp_set_mouse_cursor_image()` should be called continues to work.

Implementation was rather straightforward on all platforms except the web which was quite a bit more thrilling!
To make a cursor, sokol_app will first take the raw rgba pixels and write them to a bmp image in memory. I picked bmp because it seemed to be the simplest format to write that browser support. Transparency is poorly specified in the bmp format, and different browser understand different versions, but after some experimentation I found a format that works on safari, chromium and firefox (for the record, it's the version that has a bip header of 124bytes). The bmp is then passed to the JS side (along with hotspot coordinates), put into a blob, which itself is made into a URL. The url is then stored in a global array, along with the hotspot coordinates. The index into this array is stuffed in the sapp_mouse_cursor_image opaque handle, which can be used later to retrive the JS object and update the CSS with the proper url and hotspot. The array will never shrink, but again I think this is not a concern with reasonable usage of the api.

Finally, I also made a new sample to test the behaviour on different platforms. I'll do a separate PR on sokol-samples. It doesn't have to be merged, but will be useful to test this PR.

I hope this is a helpful PR, happy to do any required changes of course!

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -10914,12 +10914,12 @@ _SOKOL_PRIVATE sapp_mouse_cursor_image _sapp_x11_make_mouse_cursor_image(sapp_im
     img->yhot = (XcursorDim) hotspot_y;
     const size_t dest_num_bytes = (size_t)(img->width * img->height) * sizeof(XcursorPixel);
     SOKOL_ASSERT(dest_num_bytes == desc->pixels.size);
-    // Convert RGBA -> ARGB
+    // Copy RGBA -> BGRA
     for (size_t i = 0; i < dest_num_bytes; i += 4) {
-        ((uint8_t*) img->pixels)[i+3] = ((uint8_t*) desc->pixels.ptr)[i+0];
-        ((uint8_t*) img->pixels)[i+2] = ((uint8_t*) desc->pixels.ptr)[i+3];
-        ((uint8_t*) img->pixels)[i+1] = ((uint8_t*) desc->pixels.ptr)[i+2];
-        ((uint8_t*) img->pixels)[i+0] = ((uint8_t*) desc->pixels.ptr)[i+1];
+        ((uint8_t*) img->pixels)[i+0] = ((uint8_t*) desc->pixels.ptr)[i+2];
+        ((uint8_t*) img->pixels)[i+1] = ((uint8_t*) desc->pixels.ptr)[i+1];
+        ((uint8_t*) img->pixels)[i+2] = ((uint8_t*) desc->pixels.ptr)[i+0];
+        ((uint8_t*) img->pixels)[i+3] = ((uint8_t*) desc->pixels.ptr)[i+3];
     }
     Cursor cursor = XcursorImageLoadCursor(_sapp.x11.display, img);
     ret.opaque = (uint64_t) cursor;
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -4258,7 +4258,7 @@ _SOKOL_PRIVATE sapp_mouse_cursor_image _sapp_macos_make_mouse_cursor_image(sapp_
         bitsPerPixel:32];
     
     if (rep != nil) {
-        memcpy([rep bitmapData], desc->pixels.ptr, desc->width * desc->height * 4);
+        memcpy([rep bitmapData], desc->pixels.ptr, (size_t) (desc->width * desc->height * 4));
 
         NSImage* native = [[NSImage alloc] initWithSize:NSMakeSize(desc->width, desc->height)];
         [native addRepresentation:rep];
@@ -10906,13 +10906,13 @@ _SOKOL_PRIVATE void _sapp_x11_destroy_cursors(void) {
 _SOKOL_PRIVATE sapp_mouse_cursor_image _sapp_x11_make_mouse_cursor_image(sapp_image_desc* desc, int hotspot_x, int hotspot_y) {
     sapp_mouse_cursor_image ret = {};
     XcursorImage* img = XcursorImageCreate(desc->width, desc->height);
-    SOKOL_ASSERT(img && (img->width == desc->width) && (img->height == desc->height) && img->pixels);
-    img->xhot = hotspot_x;
-    img->yhot = hotspot_y;
+    SOKOL_ASSERT(img && ((int) img->width == desc->width) && ((int) img->height == desc->height) && img->pixels);
+    img->xhot = (XcursorDim) hotspot_x;
+    img->yhot = (XcursorDim) hotspot_y;
     const size_t dest_num_bytes = (size_t)(img->width * img->height) * sizeof(XcursorPixel);
     SOKOL_ASSERT(dest_num_bytes == desc->pixels.size);
     // Convert RGBA -> ARGB
-    for (int i = 0; i < dest_num_bytes; i += 4) {
+    for (size_t i = 0; i < dest_num_bytes; i += 4) {
         ((uint8_t*) img->pixels)[i+3] = ((uint8_t*) desc->pixels.ptr)[i+0];
         ((uint8_t*) img->pixels)[i+2] = ((uint8_t*) desc->pixels.ptr)[i+3];
         ((uint8_t*) img->pixels)[i+1] = ((uint8_t*) desc->pixels.ptr)[i+2];
@@ -12322,7 +12322,7 @@ SOKOL_API_IMPL void sapp_set_mouse_cursor(sapp_mouse_cursor cursor) {
     SOKOL_ASSERT((cursor >= 0) && (cursor < _SAPP_MOUSECURSOR_NUM));
     SOKOL_ASSERT(cursor != SAPP_MOUSECURSOR_CUSTOM_IMAGE); // call sapp_make_mouse_cursor_image instead.
     if (_sapp.mouse.current_cursor != cursor) {
-        _sapp_update_cursor(cursor, (sapp_mouse_cursor_image) {}, _sapp.mouse.shown);
+        _sapp_update_cursor(cursor, (sapp_mouse_cursor_image) {0}, _sapp.mouse.shown);
     }
 }
 
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -5608,7 +5608,6 @@ EM_JS(int, sapp_js_make_mouse_cursor_image, (uint8_t* bmp_ptr, int bmp_size, int
 })
 
 // Only used by the emscriten backend right now. Returned memory must be freed by caller.
-// Only used by the emscriten backend right now. Returned memory must be freed by caller.
 sapp_range _sapp_bitmap_from_image_desc(sapp_image_desc* desc) {
     SOKOL_ASSERT(desc->width * desc->height * 4 == (int) desc->pixels.size);
     size_t bmp_header_size = 14;
@@ -5634,6 +5633,7 @@ sapp_range _sapp_bitmap_from_image_desc(sapp_image_desc* desc) {
     write_int32(0); // reserved
     write_int32(bmp_header_size+dib_header_size); // offset to pixel data
     SOKOL_ASSERT((size_t)(bmp_write - bmp_header_start) == bmp_header_size);
+    SOKOL_UNUSED(bmp_header_start);
     // DIB Header
     uint8_t* dib_header_start = bmp_write;
     write_int32(dib_header_size); // header size
@@ -5654,12 +5654,14 @@ sapp_range _sapp_bitmap_from_image_desc(sapp_image_desc* desc) {
     write_int32('sRGB'); // color space type 'Win ' 'sRGB' or 0 for RGB
     bmp_write += 64; // color space stuff, unused for 'Win ' or 'sRGB'
     SOKOL_ASSERT((size_t)(bmp_write - dib_header_start) == dib_header_size);
+    SOKOL_UNUSED(dib_header_start)
     #undef write_byte
     #undef write_int16
     #undef write_int32
     // copy the pixel data row by row (bmp is bottom to top)
     ptrdiff_t remain = (bmp_data + bmp_size) - bmp_write;
     SOKOL_ASSERT(remain == (int) desc->pixels.size);
+    SOKOL_UNUSED(remain);
     size_t row_size = (size_t)desc->width * 4;
     for (int y = 0; y < desc->height; y++) {
         memcpy(bmp_write + (size_t)(desc->height - y - 1) * row_size, (uint8_t*)desc->pixels.ptr + (size_t)y * row_size, row_size);
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -5588,7 +5588,7 @@ _SOKOL_PRIVATE void _sapp_emsc_update_cursor(sapp_mouse_cursor cursor, sapp_mous
 }
 
 EM_JS(int, sapp_js_make_mouse_cursor_image, (uint8_t* bmp_ptr, int bmp_size, int hotspot_x, int hotspot_y), {
-    const copy = new Uint8Array(Module.HEAPU8.buffer, bmp_ptr, bmp_size);
+    const copy = new Uint8Array(HEAPU8.buffer, bmp_ptr, bmp_size);
     const blob = new Blob([copy.slice()], { type: 'image/bpm' });
     const url = URL.createObjectURL(blob);
     const cursor = {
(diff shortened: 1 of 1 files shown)


floooh:
Wow big one :) It might take bit until I get around properly reviewing and merging (I want to get the sokol-gfx resource view update out of the door first, which will be at least one more week).

I'll try to add a couple of code review comments in the meantime...

I also need to think about the handle type... a proper 'pool' like for sokol-gfx resources might be overkill, but I wonder if it makes sense to have a fixed number of custom mouse cursor slots, e.g. smth like `SAPP_MOUSECURSOR_CUSTOM_IMAGE_1` to `SAPP_MOUSECURSOR_CUSTOM_IMAGE_16` or so... and use those instead of handles - and you simply associate an `sapp_image_desc` with one of those custom slots.

That way the `sapp_set_mouse_cursor_image()` function could be removed and `sapp_set_mouse_cursor()` would continue to work.

...e.g.

```c
sapp_mouse_cursor_image sapp_make_mouse_cursor_image(sapp_image_desc* desc, int hotspot_x, int hotspot_y);
```
...would change into:

```c
void sapp_make_mouse_cursor_image(sapp_mouse_cursor cursor, const sapp_image_desc* desc);
```

(this also contains the code review changes from below, e.g. moving the hotspot into `sapp_image_desc`)

Trying to override one of the standard mouse cursors would log an error and otherwise be a no-op - alternatively (but maybe for some later time), we could allow overriding the standard cursors, but this would then need another function to restore the original cursor).

PS: the `_make_` function name is a bit poor in this case... maybe set/define/attach or smth....

PPS: I think `sapp_bind_mouse_cursor_image(...)` might be the best name, and later this could have an associated `sapp_unbind_mouse_cursor_image()` - which in case of overriding the standard cursor slots would restore the 'default system cursor'.

Seb-degraff:
Thanks for the quick feedback!

I really like the idea of a fixed numbers of user cursor slots, and how it simplifies the api. Couple of questions/clarifications:
1. I suppose it also means we don't need a destroy function for the cursor image, all "bound" user cursors can be released on sokol_app shutdown?
2. what should happen if sapp_bind_mouse_cursor_image is called a second time on the same slot? Error + no op as well? Or free the associated platform resources and bind to the new one?
3. should the number of user cursor slots be configurable?
I could imagine some games needing a lot of cursors (perhaps a game with a ton of interaction types, or one with animated cursors). I don't know how we could achieve that; the first thing that comes to my mind would be to keep the enum as is and let the user pass values equal or higher than _SAPP_MOUSECURSOR_NUM, up until some configurable constant (e.g. _SAPP_MOUSECURSOR_NUM  + SAPP_MAX_USERCURSORIMAGE). Or use the negative range perhaps? Both sound less clean than explicit enum entries though, and I could imagine it wouldn't work well with bindings to other langs...

> Wow big one :) It might take bit until I get around properly reviewing and merging (I want to get the sokol-gfx resource view update out of the door first, which will be at least one more week).

Of course, no hurry :)

floooh:
> I suppose it also means we don't need a destroy function for the cursor image, all "bound" user cursors can be released on sokol_app shutdown?

I think there should be a `sapp_unbind_mouse_cursor_image(sapp_mouse_cursor cursor)` function which destroys any underlying operating system objects and puts that slot back into the 'unbound' state. Calling that function with a cursor that's currently unbound should be a silent no-op (similar to how the sokol-gfx destroy functions can be called with an invalid resource handle).

Similar, calling `sapp_unbind_mouse_cursor_image()` on a 'standard cursor' should be a no-op, at least as long as we don't allow overriding standard cursors - in that case the `unbind` operation should restore the standard cursor image for that slot.

When sokol-app shuts down it should just call `sapp_unbind_mouse_cursor_image()` on all cursor slots.

> what should happen if sapp_bind_mouse_cursor_image is called a second time on the same slot? Error + no op as well? Or free the associated platform resources and bind to the new one?

The most convenient behaviour would be to allow calling bind on an already bound cursor slot and it will simply 'rebind' (e.g. internally call `unbind` and then `bind`.

> should the number of user cursor slots be configurable?

No I wouldn't go that far, we should provide a big enough number of custom slots (but I think 16 should be plenty, at most 32), and each custom slot should have an explicit enum item (e.g. not `SAPP_MOUSECURSOR_NUM + x` tricks).

If a game needs many different custom cursors (e.g. maybe in an inventory where the cursor should become some random inventory item during a drag'n'drop operation) the game can just bind a new image at the start of a drag'n'drop). Same with animated cursors, ideally the unbind/bind combo is fast enough that this can be done once per frame without hickups.

Seb-degraff:
Sorry, just saw you had answered.

Thanks for the details, that clears it up perfectly!

I absolutely agree that a fixed enum is better, and it's a good point that an app can dynamically rebind cursor slots in case it needs more than 16 – I kind of lost sight of that by focusing on the "create all cursors at startup" scenario.

I'll do the changes soon, probably today or tomorrow.

Thanks again!

Seb-degraff:
Ok, moved to the better api!

I think it's getting close to done, at least from my pov. I need to check the current diff for the PR since I went back on some changes – there might be some cleanup left to do.

Internally I kept the concept of opaque handle, but removed the wrapper type. The handles are stored in _sapp.custom_mouse_cursors. It's nice to have a single array shared by all backends as it allows for quickly checking if a custom cursor is bound to a particular sapp_mouse_cursor.

I choose to directly go for allowing binding custom images to system cursors. It was easy to do, and I think it's cool that you can bind images to the various system cursors at startup and have e.g. sokol_imgui.h use them automatically.

One other small improvement I did was to generate cursor css properties for the web backend at cursor_bind time rather than at cursor_set time, to avoid allocations on the js side.

floooh:
I'll try to do a proper code review "soon-ish" :)

floooh:
Ok, I'll start looking into the PR now, might take a few days (don't worry about the conflict, I'll take care of that)

floooh:
Tested and reviewed:

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] Web

floooh:
On Windows I stumbled over something strange:

In `_sapp_win32_destroy_custom_mouse_cursor()` the `DestroyCursor` call returns zero which indicates failure. My first hunch was that the correct function is `DestroyIcon`, but that also returns zero. Investigating...

PS: hmm, GetLastError() returns 0, which is `ERROR_SUCCESS`... the `DestroyCursor()` doc specifically says to call GetLastError to get error information:

https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-destroycursor

> If the function fails, the return value is zero. To get extended error information, call [GetLastError](https://learn.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror).

PPS: what I *will* do though is change the `DestroyCursor` call to `DestroyIcon`, that's also what GLFW does (it creates the cursor also with `CreateIconIndirect` and uses `DestroyIcon` to destroy the cursor.

PPPS: the doc say though "the icon must not be in use"... maybe the cursor must be 'unused' before the call... I'll try tinkering in that direction.

floooh:
Yes, it's about the cursor icon currently in use on Win32, the DestroyIcon() function fails in that case. I changed the `_sapp_unbind_mouse_cursor_image()` function like this (if the active cursor is unbound, hide it first, then destroy the icon, then show it again, this breaks the dependency):

```c
SOKOL_APP_API_DECL void sapp_unbind_mouse_cursor_image(sapp_mouse_cursor cursor) {
    SOKOL_ASSERT((cursor >= 0) && (cursor < _SAPP_MOUSECURSOR_NUM));
    uint64_t* slot = &_sapp.custom_mouse_cursors[(int) cursor];
    if (*slot) {
        // at least on win32 attempting to destroy a cursor that's currently
        // active is an error, so when trying to destroy the active cursor
        // first make the cursor invisible which 'breaks' the dependency
        const bool shown = _sapp.mouse.shown;
        if (_sapp.mouse.current_cursor == cursor) {
            _sapp_update_cursor(cursor, false);
        }
        #if defined(_SAPP_MACOS)
        _sapp_macos_destroy_custom_mouse_cursor(*slot);
        #elif defined(_SAPP_EMSCRIPTEN)
        _sapp_emsc_destroy_custom_mouse_cursor(cursor);
        #elif defined(_SAPP_WIN32)
        _sapp_win32_destroy_custom_mouse_cursor(*slot);
        #elif defined(_SAPP_LINUX)
        _sapp_x11_destroy_custom_mouse_cursor(*slot);
        #endif
        *slot = 0;
        // update the displayed cursor in case the current cursor is the one we just unbound
        if (_sapp.mouse.current_cursor == cursor) {
            _sapp_update_cursor(cursor, shown);
        }
    }
}
```

...in the `_sapp_win32_destroy_custom_mouse_cursor()` function I log a warning when DestroyIcon returns with a failure (this shouldn't happen anymore, but I'm not sure if all edge cases are covered):

```c
SOKOL_API_IMPL void _sapp_win32_destroy_custom_mouse_cursor(uint64_t opaque_handle) {
    HCURSOR cursor_handle = (HCURSOR) opaque_handle;
    // NOTE: DestroyIcon() may return zero (failure) if the cursor is currently in
    // use. Normally that shouldn't happen since when attempting to unbind the
    // current cursor it will be hidden first, but since there might be other edge
    // cases we just log a warning but don't fail hard
    BOOL res = DestroyIcon(cursor_handle);
    if (!res) {
        _SAPP_WARN(WIN32_DESTROYICON_FOR_CURSOR_FAILED);
    }
}
```

...I'm doing those changes in a merge branch btw: https://github.com/floooh/sokol/tree/Seb-degraff-sapp-custom-mouse-cursor-image

floooh:
PS: I just thought of a simpler solution one sec...

This is better :) (basically just switching back to the default cursor before destroying the custom cursor).

```c
SOKOL_APP_API_DECL void sapp_unbind_mouse_cursor_image(sapp_mouse_cursor cursor) {
    SOKOL_ASSERT((cursor >= 0) && (cursor < _SAPP_MOUSECURSOR_NUM));
    uint64_t* slot = &_sapp.custom_mouse_cursors[(int) cursor];
    if (*slot) {
        uint64_t platform_cursor_handle = *slot;
        // if this is the active cursor, first restore it to its default image,
        // this must be done before attempting to destroy any cursor image
        // resources which at least on win32 would fail if the cursor is still in use
        *slot = 0;
        if (_sapp.mouse.current_cursor == cursor) {
            _sapp_update_cursor(cursor, _sapp.mouse.shown);
        }
        #if defined(_SAPP_MACOS)
        _sapp_macos_destroy_custom_mouse_cursor(platform_cursor_handle);
        #elif defined(_SAPP_EMSCRIPTEN)
        _sapp_emsc_destroy_custom_mouse_cursor(platform_cursor_handle);
        #elif defined(_SAPP_WIN32)
        _sapp_win32_destroy_custom_mouse_cursor(platform_cursor_handle);
        #elif defined(_SAPP_LINUX)
        _sapp_x11_destroy_custom_mouse_cursor(platform_cursor_handle);
        #endif
    }
}
```

floooh:
Ah, found a problem with ARC on macOS, good thing I thought of testing the ARC build configs.

The code immediately crashes in this bridged cast when ARC is enabled:

```c
ns_cursor = (__bridge NSCursor*) (void*) custom_cursor;
```

...thankfully the solution is quite simple, the `NSCursor*` object ids need to be pinned into memory by writing the ids to the already existing array `_sapp_macos_t.cursors[]`.

The situation is really quite similar to Javascript :)

I think I will just use the same idea as in JS anywhere else. E.g. get rid of the "uint64_t hack", and just keep a boolean array to indicate whether a custom image has been defined for a cursor slot.

All platform-specific state structs (except `_sapp_emsc_t` already keep an array of cursor handles, and I guess we can just use this for the custom cursor handles too, e.g.:

```c
_sapp_win32_t =>  Cursor cursors[];
_sapp_macos_t => NSCursor* cursors[];
_sapp_x11_t => Cursor cursors[];
```

PS: except just using the same array won't work since we need to preserve the original system cursor handles... so maybe just a separate per-platform `custom_cursors[]` array.

floooh:
Here's what I have in mind (get rid of the uint64_t handles) for macOS:

https://github.com/floooh/sokol/pull/1321/commits/f64afdcdeec8779d3db189ddbb3324c5348eca8c

E.g. the top-level array of type-erased custom cursor handles is now a bool array which just says whether that slot has a custom cursor, and down in the platform-specific state the previous `cursors[]` array becomes `system_cursors[]`, and there's a new array of `custom_cursors[]`.

On macOS this fixes the ARC problems because the platform-specific cursor handles are pinned into memory, but IMHO this idea generally is a bit cleaner also for the other platforms (and JS is also no longer a special case). I'll implement that same fix for the other platforms next :)

floooh:
Ok, I have changed the macOS, Linux and Windows code to use a separate array of platform-specific cursor handles which replaces the uint64_t handle code (instead this array has been replaced with a bool array which just stores whether a custom cursor image has been bound to that slot). It's a surprisingly small change and it even simplified the code here and there.

Tomorrow I'll take care of web version (should need even fewer changes) and then I'll merge either tomorrow evening or Monday morning.

Big kudos btw for the clean and complete PR (especially also the new sample), it's really nice to work with the code :)

You can keep track of my changes here:

https://github.com/floooh/sokol/pull/1321

...hopefully merging this 'followup' PR should also automatically close this PR as merged (it should since I normally create such intermediate merge branches to apply smaller fixes, the only difference is that I usually don't create a separate PR from that intermediate branch)

Seb-degraff:
Thank you for looking at the PR and for the changes!

Definitely looks cleaner without the opaque handles, which in retrospect stopped making sense from the moment they weren't exposed to the user anymore. I was hesitant to make per-backend arrays as it felt like more code and less shared logic, but it's not actually much code and the custom_cursor_bound array solves the shared logic part nicely. And it cleans up the special case of JS. Perfect!

> Big kudos btw for the clean and complete PR (especially also the new sample), it's really nice to work with the code :)
Thanks, I try! Likewise to you for sokol :)

By the way, I just made [another PR](https://github.com/floooh/sokol/pull/1323) to fix a cursor-related bug on macos. If you'd rather merge and test it together with this pr, feel free to cherry pick the commit!

floooh:
I also just moved the BMP encoding code from the C side to the JS side 😬

Not entirely sure if it makes sense, but at least it gets rid of the temporary allocation on the WASM side (in exchange for producing some short-lived garbage on the JS side)...

I'll merge the custom cursor PR tomorrow (Monday), together with the macOS fix.

Not sure yet about the fullscreen PR, I'll need to do more testing.

~~Are you planning to use fullscreen on web for anything?~~ nvm, just saw the link to your game :)

floooh:
Ok, I'm kinda impressed that Github recognized the branch ancestry and automatically marked this PR as merged as well ;)

PR #1310, opened 2025-08-13
Avoid warnings in mingw

cloudwu:
mingw gcc don't accept `#pragma comment`, and some warnings would be generated.
```
3rd/sokol/sokol_app.h:2230: warning: ignoring '#pragma comment ' [-Wunknown-pragmas]
 2230 |             #pragma comment (linker, "/subsystem:windows")
      |
3rd/sokol/sokol_app.h:2236: warning: ignoring '#pragma comment ' [-Wunknown-pragmas]
 2236 |     #pragma comment (lib, "kernel32")
      |
3rd/sokol/sokol_app.h:2237: warning: ignoring '#pragma comment ' [-Wunknown-pragmas]
 2237 |     #pragma comment (lib, "user32")
      |
3rd/sokol/sokol_app.h:2238: warning: ignoring '#pragma comment ' [-Wunknown-pragmas]
 2238 |     #pragma comment (lib, "shell32")    /* CommandLineToArgvW, DragQueryFileW, DragFinished */
      |
3rd/sokol/sokol_app.h:2239: warning: ignoring '#pragma comment ' [-Wunknown-pragmas]
 2239 |     #pragma comment (lib, "gdi32")
      |
3rd/sokol/sokol_app.h:2241: warning: ignoring '#pragma comment ' [-Wunknown-pragmas]
 2241 |         #pragma comment (lib, "dxgi")
      |
3rd/sokol/sokol_app.h:2242: warning: ignoring '#pragma comment ' [-Wunknown-pragmas]
 2242 |         #pragma comment (lib, "d3d11")`
```

There doesn't seem to be any mention of any equivalent pragmas https://gcc.gnu.org/onlinedocs/gcc/Pragmas.html
This patch checks the macro `_MSC_VER` to turn on these progama comment, but I'm not sure if this is exclusive to msvc.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2220,6 +2220,12 @@ inline void sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
     #include <windows.h>
     #include <windowsx.h>
     #include <shellapi.h>
+
+    #if defined(__GNUC__)
+        #pragma GCC diagnostic push
+        #pragma GCC diagnostic ignored "-Wunknown-pragmas"
+    #endif
+
     #if !defined(SOKOL_NO_ENTRY)    // if SOKOL_NO_ENTRY is defined, it's the application's responsibility to use the right subsystem
 
         #if defined(SOKOL_WIN32_FORCE_MAIN) && defined(SOKOL_WIN32_FORCE_WINMAIN)
@@ -2240,6 +2246,10 @@ inline void sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
     #if defined(SOKOL_D3D11)
         #pragma comment (lib, "dxgi")
         #pragma comment (lib, "d3d11")
+    #endif
+
+    #if defined(__GNUC__)
+        #pragma GCC diagnostic pop
     #endif
 
     #if defined(SOKOL_D3D11)
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2224,22 +2224,26 @@ inline void sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
 
         #if defined(SOKOL_WIN32_FORCE_MAIN) && defined(SOKOL_WIN32_FORCE_WINMAIN)
             // If both are defined, it's the application's responsibility to use the right subsystem
-        #elif defined(SOKOL_WIN32_FORCE_MAIN)
-            #pragma comment (linker, "/subsystem:console")
-        #else
-            #pragma comment (linker, "/subsystem:windows")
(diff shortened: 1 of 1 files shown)


floooh:
Thanks!

I would prefer to silence the warning via `#pragma GCC diagnostic ignored "-Wunknown-pragmas"` though, because Clang on Windows supports those `#pragma comment (lib, ...)` pragmas (see: https://github.com/llvm/llvm-project/commit/e43f0fea153bd67e39f085d92d9e12ada915bb99), but I don't think that Clang on Windows can be identified via `_MSC_VER`.

I think it should look like this:

```c
#if defined(__GNUC__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunknown-pragmas"
#endif

// all the pragma-comment-lib statements which trigger this warning...

#if defined(__GNUC__)
#pragma GCC diagnostic pop
#endif
```

Could you try if that works and change the PR like that?

Tbh though, this 'unknown-pragma' warning is so common (also with GCC on other platforms) that I usually just disable that globally in the build system.

In general, I'm maintaining clang build files for Windows (e.g. https://github.com/floooh/sokol-samples/blob/master/fips-files/configs/d3d11-win64-clang-vscode-debug.yml), but currently those are not wired up to the various CI pipelines so testing is only very sporadic.

Clang is also indirectly used when using the sokol-zig bindings...

OTH GCC on Windows isn't tested at all so far, but as long as it isn't too much hassle we should keep it working.

PS: don't worry about the broken `test-zig` CI step, that's unrelated (some bug in the current Zig nightly when fetching packages)

cloudwu:
> 
> Could you try if that works and change the PR like that?
> 

I changed the PR, it works.

floooh:
Ok thanks! ~~Gitlab~~ Github CI had another hickup in the Android build but that's also unrelated. I'll just merge.

Other headers (like sokol_gfx.h and sokol_audio.h) might run into the same problem, but I can take care of those separately (I'll write a ticket for that to myself).

cloudwu:
I'm using sokol_gfx.h but I'm not seeing the warnings.

I just looked at it , it checks the macro `_MSC_VER`.

https://github.com/floooh/sokol/blob/master/sokol_gfx.h#L5153

You may want to change it to the same method.

floooh:
Oh interesting... that probably means that Clang actually detects _MSC_VER...

But I think the solution to suppress those warnings on GCC is the better solution... I wrote a ticket here, I'll take care of that after the big upcoming resource-view merge.

https://github.com/floooh/sokol/issues/1311

PR #1317, opened 2025-08-23
sokol-d: add `gfx-imgui` support

kassane:
# Sync
- https://github.com/kassane/sokol-d/pull/85

cc: @floooh

### DIFF
--- a/bindgen/gen_all.py
+++ b/bindgen/gen_all.py
@@ -52,6 +52,7 @@ d_tasks = [
     [ '../sokol_fetch.h', 'sfetch_', [] ],
     [ '../util/sokol_memtrack.h', 'smemtrack_', [] ],
     [ '../util/sokol_imgui.h', 'simgui_',   ['sg_', 'sapp_'] ],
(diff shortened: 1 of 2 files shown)


floooh:
FYI, I'm currently trying to catch up on smaller issues and PRs, it might take a few days until I get around merging.

kassane:
No worries! Currently, I've finished any changes to this feature in PR in both repositories.

> [!NOTE]
> Also, note that it is intended to do something similar to https://github.com/floooh/sokol-zig/issues/129.

floooh:
@kassane ok merged, sorry for the delay.

floooh:
PS: note though that I'm trying to come up with a more flexible way to add the sokol util headers to the sokol-zig bindings without having the sokoll-zig package depend on the 3rd party libs like Dear ImGui (or for instance Spine)... instead I want to improve the 'bring your own dependency' idea by splitting up the monolithic sokol C library into a core library and then maybe one library per util header. It's not super high priority though...

kassane:
@floooh, please try re-run CI/CD.

floooh:
Working now. Thanks!

PR #1319, opened 2025-08-26
sokol_app.h: wait for X11 VisibilityNotify event when mapping window

### LINKED ISSUE #1318: sokol_app.h: wait for X11 VisibilityNotify event before starting app
Some X11 actions require the window to be visible. For example, when I try to set mouse lock to `true` in the application's `init` function, `XGrabPointer` returns `BadWindow`. Looking into the GLFW source code, after mapping the window, they wait for the `VisibilityNotify` event with a timeout. The following proof of concept works on my machine, but it never times out.
```c
_SOKOL_PRIVATE void _sapp_x11_show_window(void) {
    if (!_sapp_x11_window_visible()) {
        XMapWindow(_sapp.x11.display, _sapp.x11.window);
        XEvent dummy;
        XWindowEvent(_sapp.x11.display, _sapp.x11.window, VisibilityChangeMask, &dummy);
        XRaiseWindow(_sapp.x11.display, _sapp.x11.window);
        XFlush(_sapp.x11.display);
    }
}
```
Writing a better implementation with a time-out requires getting the file descriptor for the display and using `select` or `poll` to wait for any event and then check if the event is of the correct type.

Joan-rv:
Fixes #1318

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -10644,6 +10644,19 @@ _SOKOL_PRIVATE void _sapp_x11_send_event(Atom type, int a, int b, int c, int d, 
                &event);
 }
 
+_SOKOL_PRIVATE bool _sapp_x11_wait_for_event(int event_type, double timeout_sec, XEvent* out_event) {
+    _sapp_timestamp_t ts;
+    _sapp_timestamp_init(&ts);
+    while (!XCheckTypedWindowEvent(_sapp.x11.display, _sapp.x11.window, event_type, out_event)) {
+        struct pollfd fd = { ConnectionNumber(_sapp.x11.display), POLLIN };
+        poll(&fd, 1, timeout_sec * 1000);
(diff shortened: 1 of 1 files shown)


floooh:
Thanks for the PR! I'll try to take care of it with the current 'batch'.

Don't worry about the failing CI, that's some unrelated problem when pulling the Nim toolchain.

floooh (changes requested):
A couple of small change requests, thanks!

floooh:
Thanks for the changes! I'll try to give it a whirl and merge within the next few days :)

floooh:
Ok, gave it a whirl on my Linux desktop (via XWayland) and it's working fine. Merging now...

floooh:
Ok merged. I also added a little blurb to the changelog. Thanks!

PR #1322, opened 2025-08-31
sokol_app: added fullscreen support for web platform

Seb-degraff:
As the title says!

Pretty straightforward change. Tested in the three major browsers (on mac). Also tested on an android phone and on an iPad.

To quickly test, add these lines to the sample `events-sapp.cc` below `ImGui::Text("dropped files (%d/%d):", sapp_get_num_dropped_files(), max_dropped_files);`

```
bool is_fullscreen = sapp_is_fullscreen();
if (ImGui::Checkbox("Fullscreen", &is_fullscreen)) {
    sapp_toggle_fullscreen();
}
```

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -133,7 +133,7 @@
     IME                 | TODO    | TODO? | TODO  | ???   | TODO    |  ???
     key repeat flag     | YES     | YES   | YES   | ---   | ---     |  YES
     windowed            | YES     | YES   | YES   | ---   | ---     |  YES
-    fullscreen          | YES     | YES   | YES   | YES   | YES     |  ---
+    fullscreen          | YES     | YES   | YES   | YES   | YES     |  YES(3)
     mouse hide          | YES     | YES   | YES   | ---   | ---     |  YES
     mouse lock          | YES     | YES   | YES   | ---   | ---     |  YES
     set cursor type     | YES     | YES   | YES   | ---   | ---     |  YES
@@ -147,6 +147,7 @@
 
     (1) macOS has no regular window icons, instead the dock icon is changed
     (2) supported with EGL only (not GLX)
+    (3) fullscreen in the browser not supported on iphones
 
     STEP BY STEP
     ============
@@ -863,6 +864,10 @@
 
     To check if the application window is currently in fullscreen mode,
     call sapp_is_fullscreen().
+
+    On the web, sapp_desc.fullscreen will have no effect, and the application
+    will always start in non-fullscreen mode. Call sapp_toggle_fullscreen() to
+    switch to fullscreen programatically, if the user allows it.
 
     WINDOW ICON SUPPORT
     ===================


floooh:
Thanks! In the past I always had one or another problem with fullscreen mode (implemented with Emscripten's fullscreen wrapper functions), mainly on iOS Safari (which didn't support fullscreen at all back then), and then another problen that the popups for fullscreen and mouselock got in the way of each other. At that point I basically gave up :) But maybe this stuff has improved in recent years.

floooh:
Regarding this:

```
bool is_fullscreen = sapp_is_fullscreen();
if (ImGui::Checkbox("Fullscreen", &is_fullscreen)) {
    sapp_toggle_fullscreen();
}
```

...I'm surprised this works, since normally such functionality needs to be called from within a 'short-lived input event handler', I would have expected that `sapp_toggle_fullscreen()` would only set a flag, and the actual requestFullscreen call needs to be done inside the input event callbacks (similar to how pointerlock works: https://github.com/floooh/sokol/blob/42640f062344e1df3b2b32d570e88be9bdf05757/sokol_app.h#L5503-L5511

...also see here: https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen#security

This might be also one of those web features which may work fine from localhost, but then fail on a real web server :)

...IIRC this was basically why I gave up with fullscreen support, since both mouse-lock and fullscreen require this 'transient user activation', and getting both to work together was a bit of a challenge.

Seb-degraff:
I think it changed quite a bit, browser seem to have found less intrusive ways to show the switch to fullscreen, or are just more lenient!

It worked surprisingly well on all platforms I tested, except iphone that doesn't support it (I'm guessing because apple don't want game devs shipping games outside of their walled garden, skipping their 30% cut).

I was surprised that it worked outside of a user input event too. But I can confirm it works, and not only locally. See my game here, for example: https://sebdegraff.com/games/game_3/2025-08-31/game_3.html

I vaguely recall that when unity implemented their html5 player, they hit the same issue and I think they mentioned it was timing-based e.g. the request to go fullscreen had to follow a user input event within a few milliseconds. They ended up "queuing" the fullscreen request to be applied on the next user input event. Perhaps the performance and framerate of the game has an impact? I'll try to put a sleep() (or equivalent) in there and see if it still works.

EDIT: you have to go to the pause menu to enable fullscreen in the linked game

floooh:
Yep found the fullscreen checkbox :) Interesting that it works! I guess even if a browser makes trouble users could still call `sapp_toggle_fullscreen()` from within the sokol-app event handler (which is called from within a JS event handler).

Maybe the 'must be called from event handler' is the reason it doesn't work on iOS? I seem to remember that iPad and iOS behaved differently there.

Seb-degraff:
> Maybe the 'must be called from event handler' is the reason it doesn't work on iOS? I seem to remember that iPad and iOS behaved differently there.

I don't think so, it says here that it's explicitly not supported on iPhone, sadly. Personally I'm pretty sure it's about the app store $$$ !
Screenshot 2025-08-31 at 18 07 08
https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen#browser_compatibility

Also I missed this sentence before: "Swiping down exits fullscreen mode, making it unsuitable for some use cases like games". That's sad. Not a problem for my particular game, but I can imagine it being an issue indeed... 

Perhaps fullscreen on iOS safari can just be not supported by sokol_app then. Desktop browsers and android support is already very nice for sending demos to people, for example, or for game jam games which people play on PC.

A way to query for fullscreen support would be nice then, so the app can show a "Would you like to play in fullscreen" popup at startup if supported. Perhaps a new dedicated function sapp_fullscreen_supported(), or a mechanism similar to the sg_features struct in sokol_gfx ?

floooh:
> A way to query for fullscreen support would be nice then

I would leave that to the application side for now. If we do this in sokol_app.h it should be done like querying for optional features in sokol-gfx, e.g. an `sapp_features` struct of booleans, and an `sapp_features sapp_query_features(void)` function.

If the feature can be detected by checking if some JS function exists I think it's fine, but if it involves looking at the user-agent string I'd rather delegate that to the API user ;)

> "Would you like to play in fullscreen" popup at startup if supported...

I think ideally a webgame should allow to toggle between fullscreen and windowed at any time, maybe with a little fullscreen/windowed icon like on YouTube videos:

image

One thing I need to check/fix is that the `sapp_is_fullscreen()` function always reflects the current mode (I haven't looked too detailed yet at your PR), for instance in the `imgui-highdpi-sapp` web sample (https://floooh.github.io/sokol-html5/imgui-highdpi-sapp.html) the code thinks that its running in fullscreen mode (because of the `Switch to windowed` button. This is because the demo starts in fullscreen mode via `sapp_desc.fullscreen = true` in sokol_main() - but this won't work in the web version. But currently `sapp_is_fullscreen()` still returns true.

I don't know yet whether I'll fix that in a separate PR or while merging your fullscreen PR :)

Seb-degraff:
> If the feature can be detected by checking if some JS function exists I think it's fine, but if it involves looking at the user-agent string I'd rather delegate that to the API user ;)

It's simpler than that: instead of being about whether the browser supports the fullscreen api, it's about whether or not the browser will allow the page to switch to fullscreen. There's an api to check that:

> The fullscreenEnabled property tells you whether or not it is possible to engage fullscreen mode. This is false if fullscreen mode is not available for any reason (such as the "fullscreen" feature not being allowed, or fullscreen mode not being supported).

(From here https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API)

> I think ideally a webgame should allow to toggle between fullscreen and windowed at any time, maybe with a little fullscreen/windowed icon like on YouTube videos

Good point, that'd be a great UX. But for some cases were the experience is much better in fullscreen, I think it can also make senses to also ask the user at startup (in my case it also has the nice side effect to delay the start of the game until the player clicks/taps, both focussing the canvas and allowing audio to be played).
Anyway that popup is implementable fully in the app. An improvement would be to not show it if the browser won't allow fullscreen anyway (so on iPhone currently), which implies querying the aforementioned web api (fullscreenEnabled). I'm totally fine doing that on my side, and living that out of that PR :)

> This is because the demo starts in fullscreen mode via sapp_desc.fullscreen = true in sokol_main() - but this won't work in the web version. But currently sapp_is_fullscreen() still returns true.

Ah! Well that's perfect, I had added a line to force fullscreen to false at startup on the web, so that should be solved by that PR I believe.

floooh:
Ok, there's one thing I need to find a solution for: while the `imgui-highdpi-sapp` now works (since the fullscreen flag is initially set to false) I can still confuse it by calling `sapp_toggle_fullscreen()` in the `init()` function. This is (correctly) rejected by the browser because it hasn't been called anywhere near an input event handler.

However, the sokol-app fullscreen flag remains set (I guess the Emscripten callback isn't called when setting fullscreen mode has been rejected) so the preliminary flag change that happens here isn't 'undone':

```c
_SOKOL_PRIVATE void _sapp_emsc_toggle_fullscreen(void) {
    sapp_js_toggle_fullscreen();
    // NOTE: When the fullscreen JS promise completes, _sapp.fullscreen will be updated
    //       (see _sapp_emsc_fullscreenchange_cb()).
    //       We still toggle _sapp.fullscreen here so from the sokol_app user's perspective,
    //       the switch is instant.
    _sapp.fullscreen = !_sapp.fullscreen;
}
```


I will try this instead: not rely on the Emscripten callback, but instead check the Promise that's returned by requestFullscreen() and exitFullscreen() (this is rejected when the operation fails) - and 'post-hoc' set the _sapp.fullscreen flag from the Promise result (e.g. it may happen that the `_sapp.fullscreen` flag is first set to true inside `_sapp_emsc_toggle_fullscreen()`, but then shortly after is corrected to false if switching to fullscreen failed.

floooh:
FYI this is what I came up with:

https://github.com/floooh/sokol/pull/1324/files

The request/exit fullscreen calls now check the returned promise for the actual success/failure, and 'correct' the _sapp.fullscreen flag via a callback to the WASM side. This seems to work nicely also for the case when the fullscreen switch is denied for any reason.

I also removed the Emscripten callback, and I fixed some of the prefix-things (on Firefox, exitFullscreen was actually called `mozCancelFullScreen`, and the `document.fullscreenElement` is also called differently.

I'm not sure though if we should keep the Mozilla special case, since Firefox switched to the standard names in 2018. For Safari it's still important though since they only switched in 2023.

floooh:
PS: ahhh, I still need the callback too... for the case where the user actively leaves the fullscreen mode via Esc or pressing the button in the popup banner...

floooh:
Ok, this should fix it: https://github.com/floooh/sokol/pull/1324/commits/b51c63d23e6db1ed211f67deac14ae26ba2fa8a1

floooh:
Merged via https://github.com/floooh/sokol/pull/1324

PR #1323, opened 2025-08-31
sokol_app, macos: fix not showing the correct cursor sometimes

Seb-degraff:
e.g when hovering the frame of the window (which changes the cursor), then coming back inside the window, the cursor will be incorrect.

I suppose this issue wasn't caught earlier because most apps that change the cursor will frequently change it when hovering different things, which calls _sapp_macos_update_cursor. However, if you set the cursor once in init, or even if you repeatedly set the same cursor every frame, the proper cursor will never be shown. 

Calling _sapp_macos_update_cursor() in `- (void)cursorUpdate` fixes it.

Loosely related to https://github.com/floooh/sokol/pull/1309 since showing a single custom cursor is a usecase that would be hit by this bug.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -4857,6 +4857,9 @@ static void _sapp_gl_make_current(void) {
             _sapp_macos_mods(event));
     }
 }
+- (void)cursorUpdate:(NSEvent *)event {
+    _sapp_macos_update_cursor(_sapp.mouse.current_cursor, _sapp.mouse.shown);
+}
 @end
 
 #endif // macOS


Seb-degraff:
Just a thought: to reproduce easily you can run the cursors sample from the [custom image PR](https://github.com/floooh/sokol-samples/pull/176), and click on the button at the bottom to override the default cursor. Then move the cursor to a side/corner of the macos window and back inside (not hovering a cursor panel). The mouse should get stuck on a resizing cursor or default system cursor (if I'm correct on my understanding of the bug, I didn't try that particular test!)

floooh:
> Just a thought: to reproduce easily you can run the cursors sample from the https://github.com/floooh/sokol-samples/pull/176

Yep I can reproduce...

floooh:
Ok, tested and merged. Thanks! Next I'll look into the web fullscreen PR, might take a bit longer since I also want to check what Emscripten is doing under the hood nowadays (e.g. if it's the same it might make sense to replace the JS code with an Emscripten API call... but I suspect that the Emscripten code will look a lot more complicated, so probably not a good idea...

PR #1324, opened 2025-09-01
Merge branch for #1322 (web fullscreen support)

floooh:
Merge branch with followup fixes for https://github.com/floooh/sokol/pull/1322:

- catch rejected promise when attempting to switch in and out of fullscreen and reset the `_sapp.fullscreen` flag accordingly
- some console.warn() warnings when switching in and out of fullscreen is rejected
- fixed moz/webkit prefixed names

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -5718,37 +5718,65 @@ _SOKOL_PRIVATE void _sapp_emsc_destroy_custom_mouse_cursor(sapp_mouse_cursor cur
     sapp_js_destroy_custom_mouse_cursor((int) cursor);
 }
 
+// will be called after the request/exitFullscreen promise resolves or rejects
+// to set the actual state of fullscreen mode
+EMSCRIPTEN_KEEPALIVE void _sapp_emsc_set_fullscreen_flag(int f) {
+    _sapp.fullscreen = (bool)f;
+}
+
 EM_JS(void, sapp_js_toggle_fullscreen, (void), {
     const canvas = Module.sapp_emsc_target;
     if (canvas) {
-        if (!document.fullscreenElement) {
-            let promise;
-            if(canvas.requestFullScreen) {
-                canvas.requestFullScreen();
-            } else if(canvas.webkitRequestFullScreen) {
-                canvas.webkitRequestFullScreen();
-            } else if(canvas.mozRequestFullScreen) {
-                canvas.mozRequestFullScreen()
+        // NOTE: Safari had the prefix until 2023, Firefox until 2018
+        const fullscreenElement = document.fullscreenElement || document.webkitFullscreenElement || document.mozFullScreenElement;
+        let p = undefined;
+        if (!fullscreenElement) {
+            if (canvas.requestFullscreen) {
+                p = canvas.requestFullscreen();
+            } else if (canvas.webkitRequestFullscreen) {
+                p = canvas.webkitRequestFullscreen();
+            } else if (canvas.mozRequestFullScreen) {
+                p = canvas.mozRequestFullScreen();
+            }
+            if (p) {
+                p.then(() => {
+                    __sapp_emsc_set_fullscreen_flag(1);
+                }).catch((err) => {
+                    console.warn('sapp_js_toggle_fullscreen(): failed to enter fullscreen mode with', err);
+                    __sapp_emsc_set_fullscreen_flag(0);
+                });
+            } else {
+                console.warn('sapp_js_toogle_fullscreen(): browser has no [webkit|moz]requestFullscreen function');
+                __sapp_emsc_set_fullscreen_flag(0);
             }
         } else {
             if (document.exitFullscreen) {
-                document.exitFullscreen();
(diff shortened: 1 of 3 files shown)


floooh:
Ok, I'm going to merge this now @Seb-degraff, if I messed up anything with my changes let me know and I'll provide followup fixes, but testing on macOS across Chrome, FF and Safari appears to look good.

floooh:
Fullscreen can be tested in the imgui-highdpi-sapp sample via the Switch to fullscreen/windowed button: (https://floooh.github.io/sokol-html5/imgui-highdpi-sapp.html):

image

sokol_app.h
```diff
@@ -5718,37 +5718,65 @@ _SOKOL_PRIVATE void _sapp_emsc_destroy_custom_mouse_cursor(sapp_mouse_cursor cur
...
     if (canvas) {
-        if (!document.fullscreenElement) {
-            let promise;
-            if(canvas.requestFullScreen) {
-                canvas.requestFullScreen();
-            } else if(canvas.webkitRequestFullScreen) {
-                canvas.webkitRequestFullScreen();
-            } else if(canvas.mozRequestFullScreen) {
-                canvas.mozRequestFullScreen()
+        // NOTE: Safari had the prefix until 2023, Firefox until 2018
```
Seb-degraff, line 5743: Just an additional comment: strangely, in my tests, latest firefox on mac (142.0.1) would not work without the moz prefixed version. Not sure why!
I should double-check, maybe I had messed up something

floooh:
Hmm, when I check manually in Firefox on macOS, both methods exist, and I get a deprecation warning for `mozRequestFullscreen` (that `Module.Rb` is the Closure-minified name for `Module.sapp_emsc_target`, you'll need to look that up in the browser console because it may change with each build):

image

...e.g. this was the triangle-sample: https://floooh.github.io/sokol-html5/triangle-sapp.html

PS: this is also 142.0.1

PPS: there *were* some slight incosistencies in your code for upper-vs-lower-case, e.g. `requestFullScreen` versus the correct `requestFullscreen`, while `mozRequestFullScreen` is correct but inconsistent with `requestFullscreen` and `webkitRequestFullscreen`. That would explain why it would fallthrough to `mozRequestFullScreen`, but not why Chrome works...

PPPS: aha in Chrome it worked because it has two overloads for webkit - and Safari also has those two overloads. So the original code was probably always falling through to either `webkitFullScreen` or `mozRequestFullScreen`, but never actually called `requestFullscreen` :) (f*cking browser APIs man...)

image

Seb-degraff:
I get the same depreciation warning, and indeed requestFullscreen exists. I assumed I was simply misremembering, but then I went and tried removing the moz prefix branch in my old code (before your changes, and it stopped working on Firefox!

I don't understand why removing the moz branch changes anything, it should not be reached if requestFullscreen exists... but commenting out the branch breaks it.

```js
const canvas = Module.sapp_emsc_target;
if (canvas) {
    if (!document.fullscreenElement) {
        if(canvas.requestFullScreen) {
            canvas.requestFullScreen();
        } else if(canvas.webkitRequestFullScreen) {
            canvas.webkitRequestFullScreen();
        } /* else if (canvas.mozRequestFullScreen) {
            canvas.mozRequestFullScreen();
        } */
    } else {
        if (document.exitFullscreen) {
            document.exitFullscreen();
        } else if (document.webkitExitFullscreen) {
            document.webkitExitFullscreen();
        } /* else if (document.mozExitFullscreen) {
            document.mozExitFullscreen();
        } */
    }
}
```

Adding some logs and stepping with the debugger, I see that canvas.requestFullscreen is falsy, e.g. console.log(canvas.requestFullScreen ? 'truthy' : 'falsy'); prints falsy. That seems completely crazy to me.

Seb-degraff:
I'm having trouble debugging this, because now Firefox hangs when trying to evaluate console.log(Module.sapp_emsc_target.requestFullscreen ? "yes" : "no") in the console, in the context of my game.

`console.log(canvas.requestFullScreen ? "yes" : "no");` in sapp_js_toggle_fullscreen prints no. 
Interestingly, `document.body.requestFullscreen ? "yes" : "no")` in the browser console in a blank page prints "yes".

Could it be emscripten that messes with the functions of the canvas, and somehow makes requestFullscreen falsy? Or is it a quirk of firefox specific to canvases elements, for some weird back-compatibility reason?

floooh:
Note that you're mixing `requestFullscreen` (lower-case 's') and `requestFullScreen` (upper-case 'S').

One exists (`requestFullscreen`), the other doesn't (`requestFullScreen`).

Only `webkitRequestFull[s|S]creen` has two overloads (on Chrome and Safari).

And `mozRequestFullScreen` only exists in the upper-case-S version.

It's confusing as hell, but I guess there are 'hysterical reasons' for this mess ;)

Seb-degraff:
Oh wow, thank you! I probably accepted the autocomplete suggestion in the firefox console and didn't spot the capital S.
(or vice versa)

PR #1331, opened 2025-09-19
fix query_state bug

RJ-Infinity:
this fixes a bug that i noticed when on android but i belive could effect any platform.
Resources could become invalid with no way to check if they were valid or not, as when accsessing them it would crash with a failed assert.

To fix i make `_sg_lookup_<resource>` return 0 if the slot index is less than the pool size.
This causes `sg_query_<resource>_state` to return SG_RESOURCESTATE_INVALID on those cases instead of failing an assert.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -7103,7 +7103,10 @@ _SOKOL_PRIVATE _sg_view_t* _sg_view_at(uint32_t view_id) {
 
 // returns pointer to resource with matching id check, may return 0
 _SOKOL_PRIVATE _sg_buffer_t* _sg_lookup_buffer(uint32_t buf_id) {
-    if (SG_INVALID_ID != buf_id) {
+    if (
+        SG_INVALID_ID != buf_id &&
+        _sg_slot_index(buf_id) < _sg.pools.buffer_pool.size
+    ) {
         _sg_buffer_t* buf = _sg_buffer_at(buf_id);
         if (buf->slot.id == buf_id) {
             return buf;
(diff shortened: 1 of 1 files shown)


floooh:
Hmm, this is actually illegal behaviour. You're only supposed to call the query-state functions with a handle you got from one of the `sg_make_*` or `sg_alloc_*` functions, and those cannot return handles which go beyond the pool size.

In debug mode such an invalid handle is caught in range checks inside the `_sg_*_at()` functions (but you already mentioned those).

> Resources could become invalid with no way to check if they were valid or not, as when accsessing them it would crash with a failed assert.

This I don't understand... can you share more details?

I would be interested in how you got the handles where the index is greater than the pool size since this shouldn't be possible (except those handles had been overwritten by memory corruption, but in this case the 'soft' range checks would just hide the symptoms - e.g. in case of memory corruption you'd actually *want* an assert as close to the corruption happening as possible).

RJ-Infinity:
I appolagise as i belive this is acctualy my fault. All the handles i used were created with sg_make_* functions however when running under android if you minimise the app it might force you reinitislise the graphics as it deinitialises the egl surface. however I now realise that when it destoys the surface i wasnt calling sg_shutdown only sgl_shutdown so the handles were carrying over when i reinitialise the graphics. as i say thats my bad for not shutting down sokol gfx.
I now do agree with you that the asserts were probably correct as if it had happend to fall within the valid range i probably would have had some very unusual bugs.
Thank you for considering the report anyway but feel free to close it as it was a bug in my code not yours.

floooh:
> under android if you minimise the app it might force you reinitislise the graphics as it deinitialises the egl surface

Ouch, is that new? On my Google Pixel 4a I can switch the samples into the background and then come back later without having to reinitialize everything.

RJ-Infinity:
im acctualy not sure if its new. It does only happen sometimes (about 3 in 4) but I get a APP_CMD_TERM_WINDOW command (according to the [docs](https://developer.android.com/reference/games/game-activity/group/android-native-app-glue#group__android__native__app__glue_1ggab1238b3017ceefa7fdfbcc79eee6c289ae4454163b5148d7feb2c1398e87f1bd5) it sets `android_app->window` to `NULL`) when i minimise the app followed by an APP_CMD_INIT_WINDOW when it is reopend. it might be that im on a less powerfull device that needs to save resources?

Issue #1332, opened 2025-09-20
querying statistics about resident object counts (image, other buffers)


nurpax:
Does sokol have some way of querying the # of allocated buffers?  I'd like to have some light statistics in my app about these, so that I can eyeball that there are no obvious leaks.

I did see the per-frame stats, but that doesn't seem to include information about resource allocations.

floooh:
Currently not, you'd have use the trace-hooks thingie to keep track of object creation and destruction yourself.

It would be trivial to add those to the frame-stats though, ideally for each object type:

- number of 'live objects' (technically, not a 'frame stat', but better than creating a separate stats-tracker)
- created and destroyed in current frame

(since the objects live in sparse arrays, there's currently no internal value which could simply be exposed, it would need to be a count which is incremented on object creation and decremented on destruction).

PS: ok, technically there's an `sg_pool.queue_top` value which could be used to compute the number of live objects (pool_size - queue_top) - might be more robust to simply use this one for the total count.

nurpax:
The new frame stats fields would accomplish what I had i mind.  

W.r.t `sg_pool.queue_top`, I wonder if this is readable in the Zig bindings?

floooh:
> W.r.t sg_pool.queue_top, I wonder if this is readable in the Zig bindings?

No, it would only be accessible via the global `_sg` struct but this is only visible in the C source file which compiles the sokol-gfx implementation (e.g. in the sokol-zig bindings here: https://github.com/floooh/sokol-zig/blob/master/src/sokol/c/sokol_gfx.c).

Adding the values to the frame-stats would be much cleaner.

floooh:
I'll give it a shot.

floooh:
~~Heh, looks like this is already uncovering some strange behaviour... When resizing the mrt-sapp window on macOS with the Metal backend I see that 7 image objects are destroyed and created in a frame during the resize, but it should only be 3.~~

~~Could be that the resize-event is triggered multiple times per frame (also even then it should be a multiple of 3, e.g. 6 but not 7)... will need to investigate.~~

Nvm, I didn't count the resolve-images and -views.

PPS: it still seems to be possible that the resize callback may be called multiple times, because sometimes I see 14 images being created instead of 7... I'll write a separate ticket for that...

floooh:
The sokol_gfx_imgui.h frame-stats-panel is getting a bit crowded, but the information is there.

[...]

PR #1334, opened 2025-09-23
Add resource-pool stats to sg_frame_stats

floooh:
Fixes: https://github.com/floooh/sokol/issues/1332

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -7354,6 +7354,18 @@ _SOKOL_PRIVATE void _sg_update_resource_stats(sg_resource_stats* stats, const _s
     stats->total_free = (uint32_t) pool->queue_top;
 }
 
+_SOKOL_PRIVATE void _sg_update_frame_stats(void) {
+    _sg.stats.frame_index = _sg.frame_index;
+    _sg_update_resource_stats(&_sg.stats.buffers, &_sg.pools.buffer_pool);
+    _sg_update_resource_stats(&_sg.stats.images, &_sg.pools.image_pool);
+    _sg_update_resource_stats(&_sg.stats.views, &_sg.pools.view_pool);
+    _sg_update_resource_stats(&_sg.stats.samplers, &_sg.pools.sampler_pool);
+    _sg_update_resource_stats(&_sg.stats.shaders, &_sg.pools.shader_pool);
+    _sg_update_resource_stats(&_sg.stats.pipelines, &_sg.pools.pipeline_pool);
+    _sg.prev_stats = _sg.stats;
+    _sg_clear(&_sg.stats, sizeof(_sg.stats));
+}
+
 _SOKOL_PRIVATE uint32_t _sg_align_u32(uint32_t val, uint32_t align) {
     SOKOL_ASSERT((align > 0) && ((align & (align - 1)) == 0));
     return (val + (align - 1)) & ~(align - 1);
@@ -20544,12 +20556,6 @@ SOKOL_API_IMPL int sg_query_surface_pitch(sg_pixel_format fmt, int width, int he
 
 SOKOL_API_IMPL sg_frame_stats sg_query_frame_stats(void) {
     SOKOL_ASSERT(_sg.valid);
-    _sg_update_resource_stats(&_sg.prev_stats.buffers, &_sg.pools.buffer_pool);
-    _sg_update_resource_stats(&_sg.prev_stats.images, &_sg.pools.image_pool);
-    _sg_update_resource_stats(&_sg.prev_stats.views, &_sg.pools.view_pool);
-    _sg_update_resource_stats(&_sg.prev_stats.samplers, &_sg.pools.sampler_pool);
-    _sg_update_resource_stats(&_sg.prev_stats.shaders, &_sg.pools.shader_pool);
-    _sg_update_resource_stats(&_sg.prev_stats.pipelines, &_sg.pools.pipeline_pool);
     return _sg.prev_stats;
 }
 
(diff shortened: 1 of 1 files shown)


nurpax:
Very cool!  I tried this and it seems to work.  However, on frames when I know I'm doing `sg.makeImage()`, I don't see any change any of allocated, deallocated, inited, uninited.  Should I be seeing any stats changes when using `sg.makeImage()`?  I'm on the latest zig-sokol a4e26184eb4266c235f122c6d0bc78c52877da4b (which I believe is on latest sokol).

floooh:
sg_query_frame_stats() is basically one frame behind to make sure that the caller always sees the stats for a complete frame, e.g. this _sg_update_frame_stats() helper function is called from sg_commit():

https://github.com/floooh/sokol/blob/1dd76395283c8ba42b89224942325d83887f3a64/sokol_gfx.h#L7365-L7366

nurpax:
That's how I thought it probably works.  I think I even had an assert like `allocated==0` to check if it ever reports anything else than zero.  This was after all initial loading was done though.  My app was doing on makeImage/destroyImage per frame and I didn't see any allocated or inited in this case.  I admit I don't know exactly how the allocs work but could there be some sort of pool slot re-use happening?

I'll try in a more isolated case.

floooh:
> I admit I don't know exactly how the allocs work but could there be some sort of pool slot re-use happening?

Hmm.... even if slot reuse happens the alloc/dealloc/init/uninit counters should still show something since the counts are not 'derived' from some change in the allocated pool slot count, but directly counted. E.g. for images it's here and that's called directly from `sg_make_image()`:

https://github.com/floooh/sokol/blob/1dd76395283c8ba42b89224942325d83887f3a64/sokol_gfx.h#L20587

...maythe the counts are reset somewhere unexpected, that should only happen here in `_sg_update_frame_stats()` which is called from `sg_commit()`:

https://github.com/floooh/sokol/blob/1dd76395283c8ba42b89224942325d83887f3a64/sokol_gfx.h#L7357-L7367

But I only checked the alloc counts in the mrt-sapp sample which recreates images and views on window resize, maybe there's an edge case I'm not aware of.

nurpax:
I made a simpler repro.  Basically drop the below code into f.ex. the `sokol-zig/examples/texcube.zig` file:

```
var image: sg.Image = .{};

export fn frame() void {
    const std = @import("std");

    const dt: f32 = @floatCast(sapp.frameDuration() * 60);

    state.rx += 1.0 * dt;
    state.ry += 2.0 * dt;
    const vs_params = computeVsParams(state.rx, state.ry);

    if (image.id != 0) {
        sg.destroyImage(image);
    }

    image = sg.makeImage(.{
        .width = 4,
        .height = 4,
        .data = init: {
            var data = sg.ImageData{};
            data.mip_levels[0] = sg.asRange(&[4 * 4]u32{
                0xFFFFFFFF, 0xFF000000, 0xFFFFFFFF, 0xFF000000,
                0xFF000000, 0xFFFFFFFF, 0xFF000000, 0xFFFFFFFF,
                0xFFFFFFFF, 0xFF000000, 0xFFFFFFFF, 0xFF000000,
                0xFF000000, 0xFFFFFFFF, 0xFF000000, 0xFFFFFFFF,
            });
            break :init data;
        },
    });

    const stats = sg.queryFrameStats();
    std.debug.print("stats images: {any}\n", .{stats.images});

    sg.beginPass(.{ .action = state.pass_action, .swapchain = sglue.swapchain() });
    sg.applyPipeline(state.pip);
    sg.applyBindings(state.bind);
    sg.applyUniforms(shd.UB_vs_params, sg.asRange(&vs_params));
    sg.draw(0, 36, 1);
    sg.endPass();
    sg.commit();
}
```

This prints the below log:

```
$  ~/dev/sokol-zig git:(master) ✗ zig build run-texcube
stats images: .{ .total_alive = 0, .total_free = 0, .allocated = 0, .deallocated = 0, .inited = 0, .uninited = 0 }
stats images: .{ .total_alive = 2, .total_free = 126, .allocated = 0, .deallocated = 0, .inited = 0, .uninited = 0 }
stats images: .{ .total_alive = 2, .total_free = 126, .allocated = 0, .deallocated = 0, .inited = 0, .uninited = 0 }
stats images: .{ .total_alive = 2, .total_free = 126, .allocated = 0, .deallocated = 0, .inited = 0, .uninited = 0 }
stats images: .{ .total_alive = 2, .total_free = 126, .allocated = 0, .deallocated = 0, .inited = 0, .uninited = 0 }
stats images: .{ .total_alive = 2, .total_free = 126, .allocated = 0, .deallocated = 0, .inited = 0, .uninited = 0 }
stats images: .{ .total_alive = 2, .total_free = 126, .allocated = 0, .deallocated = 0, .inited = 0, .uninited = 0 }
stats images: .{ .total_alive = 2, .total_free = 126, .allocated = 0, .deallocated = 0, .inited = 0, .uninited = 0 }
```

The fields `allocated`, `deallocated`, `inited` and `uninited` are always zero.  My expectation is that a call to `sg.makeImage()` would show up somewhere in the stats.  The field `total_alive` is already quite useful, just wanted to also observe makeImage calls in the stats, esp. as I'm working on a "image suballocator" for throw-away render targets.

I'm on macOS.

floooh:
Hmm I'm currently having trouble debugging Zig code on my Mac laptop... let me try that in one of the C samples one sec...

floooh:
Ah right, I'm an idiot... the stats are only counted in the `sg_alloc_()` and `sg_init_()` calls, but not the `sg_make_*()` calls.

Fix on the way...

floooh:
Ok better: https://github.com/floooh/sokol/commit/cf15469d568da3d1dfbb0d0f8b8b1334aa482609

I moved the stats-counter updates one level lower so that all creations and destructions are caught.

Thanks for reporting this! If you notice anything else weird, let me know :)

nurpax:
It's working now, thanks a lot!

PR #1346, opened 2025-10-09
tvOS build fixes, handle remote press events

tomasandrle:
- Fix build for tvOS, which is an "offspring" of iOS
 - Handle Apple TV remote button presses by translating them into key press events
 
Request for feedback:
 - are the `#if` conditions OK like this?
 - matching key constants (btw in my code I'm handling `_MENU` the same way as `_ESC`)
 - callback code could probably be extracted into c function(s) like the touch handlers below?

So far I tested this on the tvOS simulator, will try on device next week.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -5344,36 +5344,30 @@ _SOKOL_PRIVATE void _sapp_ios_app_event(sapp_event_type type) {
     }
 }
 
-#if defined(_SAPP_TVOS)
-_SOKOL_PRIVATE void _sapp_tvos_press_event(UIPressType press, sapp_event_type type) {
-    sapp_keycode key;
-    switch (press) {
-        case UIPressTypeUpArrow:    key = SAPP_KEYCODE_UP; break;
-        case UIPressTypeDownArrow:  key = SAPP_KEYCODE_DOWN; break;
-        case UIPressTypeLeftArrow:  key = SAPP_KEYCODE_LEFT; break;
-        case UIPressTypeRightArrow: key = SAPP_KEYCODE_RIGHT; break;
-        case UIPressTypeSelect:     key = SAPP_KEYCODE_ENTER; break;
-        case UIPressTypeMenu:       key = SAPP_KEYCODE_MENU; break;
-        case UIPressTypePlayPause:  key = SAPP_KEYCODE_MENU; break;
-        default:                    key = SAPP_KEYCODE_INVALID; break;
-    }
-    if (key != SAPP_KEYCODE_INVALID) {
-        _sapp_init_event(type);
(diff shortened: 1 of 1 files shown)


floooh:
Thanks! I'll do a thorough code review later, but in the meantime:

Please replace the ifdef checks for `TARGET_OS_IPHONE` and `TARGET_OS_TV` with the already existing `_SAPP_IOS` and create a new define `_SAPP_TVOS` (see here for how `_SAPP_IOS` is defined: https://github.com/floooh/sokol/blob/a99d4f344bf314a9ec3968fd22307d003579c745/sokol_app.h#L2139-L2151).

`_SAPP_TVOS` should be defined on top of `_SAPP_IOS` (e.g. on tvOS both should be defined).

I think in most places you can also remove the check for iOS, since the iOS specific code is already wrapped in one big check for `_SAPP_IOS` starting here: https://github.com/floooh/sokol/blob/a99d4f344bf314a9ec3968fd22307d003579c745/sokol_app.h#L5221

floooh:
Ping @tomasandrle :) Do you have time to do the define-changes I requested in the above comment? (I should probably have done a proper Github code review, but anyway...).

Otherwise I can do that too, but the PR might be pushed back a bit.

tomasandrle:
@floooh sorry for the delay, I was busy with other things.
Don't waste your time on it, I'll fix it.

floooh:
No rush :)

tomasandrle:
Pushed tweak with _SAPP_TVOS `#define` and moved handling code into C function, more like the other handlers.
Haven't added a custom Play/pause key yet.

I'm not sure if it's necessary to put the handlers (`_sapp_tvos_press_event` and API callbacks) under the `_SAPP_TVOS` define:

It can save a few bytes in iOS binaries but if there's no separate tvOS test pipeline on Github, things could break in that part of code without getting noticed... I'm slightly leaning towards leaving it as part of _SAPP_IOS - compiles fine and doesn't break anything AFAIK.

floooh:
Found one problem, there's now two functions called `_sapp_tvos_press_event`, can you fix that? See my inline comment above.

...strange though that the CI compilation tests don't complain... function overloading like this should only work in ObjC++ mode... hmm.

tomasandrle:
@floooh I inlined it. CI tests didn't complain because they didn't have `_SAPP_TVOS` defined. And when I was building for tvOS on my computer I didn't get an error because I'm including `sokol_app.h` from a .mm file, so it was in C++ mode as you suspected.

I also removed that `_SAPP_TVOS` condition to include the callback code in all iOS builds, if that's OK.

floooh:
...and merged. Many thanks!

PR #1349, opened 2025-10-11
Support for bindgen C2 language

Angluca:
https://github.com/Angluca/sokol-c2
Hope it can join floooh/sokol-c2, I'll add more example later :)

### DIFF
--- a/bindgen/README.md
+++ b/bindgen/README.md
@@ -27,6 +27,7 @@ To update the Zig bindings:
 > git clone https://github.com/kassane/sokol-d
 > git clone https://github.com/colinbellino/sokol-jai
 > git clone https://github.com/floooh/sokol-c3
+> git clone https://github.com/angluca/sokol-c2
 > python3 gen_all.py
 ```
 
--- a/.github/workflows/gen_bindings.yml
+++ b/.github/workflows/gen_bindings.yml
@@ -70,6 +70,10 @@ jobs:
         with:
           repository: floooh/sokol-c3
           path: bindgen/sokol-c3
+      - uses: actions/checkout@main
+        with:
+          repository: angluca/sokol-c2
+          path: bindgen/sokol-c2
       - name: generate
         run: |
           cd bindgen
@@ -118,6 +122,13 @@ jobs:
           name: ignore-me-c3
           retention-days: 1
           path: bindgen/sokol-c3/sokol.c3l
+      - name: upload-c2-artifact
+        uses: actions/upload-artifact@main
+        with:
+          name: ignore-me-c2
+          retention-days: 1
+          path: bindgen/sokol-c2/sokol
+          path: bindgen/sokol-c2/c
 
   test-zig:
     needs: gen-bindings
--- a/bindgen/gen_all.py
+++ b/bindgen/gen_all.py
@@ -1,4 +1,4 @@
-import os, argparse, gen_nim, gen_zig, gen_odin, gen_rust, gen_d, gen_jai, gen_c3, shutil
+import os, argparse, gen_nim, gen_zig, gen_odin, gen_rust, gen_d, gen_jai, gen_c3, gen_c2, shutil
 
 parser = argparse.ArgumentParser()
 parser.add_argument("--zig-tiger-style", action="store_true", help="Enable zig tiger style mode.")
@@ -78,3 +78,9 @@ gen_c3.prepare()
 for task in tasks:
     [c_header_path, main_prefix, dep_prefixes] = task
     gen_c3.gen(c_header_path, main_prefix, dep_prefixes)
+
+# C2
+gen_c2.prepare()
+for task in tasks:
+    [c_header_path, main_prefix, dep_prefixes] = task
+    gen_c2.gen(c_header_path, main_prefix, dep_prefixes)
--- a/bindgen/gen_c2.py
+++ b/bindgen/gen_c2.py
@@ -1,0 +1,463 @@
+#-------------------------------------------------------------------------------
+#   gen_c2.py
+#
+#   Generate c2 bindings.
+#-------------------------------------------------------------------------------
+import gen_ir
+import gen_util as util
+import os, shutil, sys
+
+bindings_root = 'sokol-c2'
+c_root = f'{bindings_root}/c'
+module_root = f'{bindings_root}/sokol'
+
+# TODO: Consider changing module names to something shorter.
+#       For example we could C prefixes, for example `sg` instead of current `gfx`.
+module_names = {
+    'slog_':    'sokol_log',
+    'sg_':      'sokol_gfx',
+    'sapp_':    'sokol_app',
+    'stm_':     'sokol_time',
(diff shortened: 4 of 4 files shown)


floooh:
Can you add me as collaborator to https://github.com/angluca/sokol-c2? Makes it easier for me to prepare the repository so that the gen-bindings CI jobs can directly push updates, and it's also easier for me to directly fix samples on breaking sokol-gfx updates.

Should be possible on this page:

https://github.com/angluca/sokol-c2/settings/access

floooh:
Btw, for future PRs I would recommend to first create a dedicated branch on your fork and create the PR from this branch instead of master :)

floooh:
PPS: there are changes in gen_nim.py which shouldn't be there and which undo more recent changes in the gen-bindings-script for Nim... were you on an outdated sokol version when you created the PR?

I would recommend starting from scratch:

- either undo all changes to the master branch in your fork and bring the fork to the latest state of the sokol master branch (e.g. see: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork) so that your fork is on the same state as the sokol master branch
- ...or alternatively (might be easier): delete your fork repository and create a new fork with the latest sokol state (make sure to backup your c2-bindings changes before you do that!)
- create a new branch named `sokol-c2-bindings` from the master branch
- implement the c2 bindings changes in this branch
- create a new PR from this branch

Angluca:
I have setting Manage access to you just
Finally no nim files in pull request:)
How to add floooh/sokol-c2, I see more language add in floooh/sokol-xxx

floooh:
> How to add floooh/sokol-c2, I see more language add in floooh/sokol-xxx

There's two options. In case of D and Jai, the bindings repository is not under `floooh`, but I have push-access to those repos:

- https://github.com/kassane/sokol-d
- https://github.com/colinbellino/sokol-jai

...we can do the same thing with the C2 bindings and keep the bindings at `angluca/sokol-c2`, or if you prefer we can also move the bindings to `floooh/sokol-c2`, your choice :)

Angluca:
Please be sure to port it to floooh/sokol-c2, because your home is the Sokol official genuine version, which will let more people see the niche language of sokol and c2, and the key is to allow more people to maintain and participate :)
Hopefully Sokol will get better and better, It feels too suitable for writing games, so small

floooh:
Okidoki, then I'll put the stuff into `floooh/sokol-c2`.

I will try to take care of all this around next weekend... for the next couple of days I'd like to focus on the vulkan backend branch I started... (at least to a point where I have a working render loop).

floooh:
Note @Angluca I have forked your sokol-c2 repo and added you as collaborator (so you have push rights: https://github.com/floooh/sokol-c2).

I'll try to work on the C2 stuff over the next couple of days. Please also see my change requests for sokol-shdc: https://github.com/floooh/sokol-tools/pull/195

Angluca:
Very cool ;)

bindgen/README.md
```diff
@@ -27,6 +27,7 @@ To update the Zig bindings:
 > git clone https://github.com/kassane/sokol-d
 > git clone https://github.com/colinbellino/sokol-jai
 > git clone https://github.com/floooh/sokol-c3
+> git clone https://github.com/angluca/sokol-c2
```
floooh: Change to `git clone https://github.com/floooh/sokol-c2`

floooh (changes requested):
Only two change requests to update the bindings repository URL.

And while at it, can you also update the list of official bindings in the readme?

E.g. here: https://github.com/floooh/sokol?tab=readme-ov-file#official-language-bindings

Angluca:
@floooh  use floooh/sokol-c2 better
And I think floooh/sokol-c2 should be an independent project rather than being inherited from me. Because who knows? Maybe I might use another language at some point. Maybe it will need to be maintained by someone else.

floooh:
> And I think floooh/sokol-c2 should be an independent project rather than being inherited from me.

I created `floooh/sokol-c2` as fork of `Angluca/sokol-c2`, so the 'heritage' is visible, but I'm going to maintain it, no worries. I still added you as collaborator to the fork since that makes things easier for quick changes. But no worries, you're not "tied" to sokol-c2, if you move on that's not a problem :)

Let me take over the PR from here on, I'll try to get it merged later today or maybe tomorrow.

floooh:
One important missing thing is a Github Action test script like this:

https://github.com/floooh/sokol-c3/blob/main/.github/workflows/build.yml

...the main problem is installing the C2 compiler. Can you find out what's the recommended way to use C2 in a CI script and add such a script to https://github.com/floooh/sokol-c2?

For instance for C3 there's a Github Action like this: https://github.com/radekm/setup-c3

...and after that we also need to add a new CI job to sokol gen_bindings.yml like this one for C3:

https://github.com/floooh/sokol/blob/f0e8e1d563fe7519a723e7eb9eb2c01c3b1b9d66/.github/workflows/gen_bindings.yml#L357-L389

Angluca:
I don't know how to write workflows/gen_bindings.yml, haven't this sokol path, must download in that server?
```zsh
# can build libsokol.a and example/clear
make example=clear LIB_SOURCE=.../this_sokol_path
...
```

floooh:
The main problem is basically how to install the C2 compiler inside the Github Action.

The only instructions I have found how to install the C2 compiler is to compile it in place, like here:

https://github.com/c2lang/c2compiler?tab=readme-ov-file#bootstrap

...or here in their CI script:

https://github.com/c2lang/c2compiler/blob/master/.github/workflows/main.yml

I can look into this, but it will take some time.

Is there a C2 Discord or something? Maybe somebody else already has a solution we can copy.

floooh:
FYI I'll experiment a bit with this today (bootstrapping a C2 toolchain for CI).

PS: nvm, some other things got in the way...

Angluca:
If you use the official c2compiler, you will have to modify the code, because the c2c default variable name is up to 31 bytes long,
Must modify c2compiler/ast_utils/constants.c2:line 21: MaxIdentifierLen = 96 when you want build sokol-c2-example

floooh:
> If you use the official c2compiler, you will have to modify the code, because the c2c default variable name is up to 31 bytes long,
Must modify c2compiler/ast_utils/constants.c2:line 21: MaxIdentifierLen = 96 when you want build sokol-c2-example

...uh, that's really not an option :)

Do you know which symbols in the sokol-c2 bindings, examples or sokol-shdc-generated code violate this restriction? Maybe it's just one or two places that are easy to fix.

This restriction really isn't great though, would be good if that's relaxed on the C2 side (without having to patch the compiler).

Angluca:
I also think that the variable length limit of 31 bytes is too little, This makes it inconvenient to use third-party C libraries, 
and recently found a good language, maybe I will make a new language sokol-xxx of sokol ;)

Angluca:
I was wondering if I should add a sokol-nature
sokol-c2 is not used for the time being

PR #1350, opened 2025-10-12
Add Vulkan backend (wip)

floooh:
...this will take a while, don't hold your breath.

Reading material:
- https://www.khronos.org/assets/uploads/developers/presentations/Ensure_Vulkan_Synchronization_Vulkanised_Oct_2021.pdf
- https://github.com/KhronosGroup/Vulkan-Docs/wiki/Synchronization-Examples
- https://themaister.net/blog/2019/08/14/yet-another-blog-explaining-vulkan-synchronization/
- https://www.gfxstrand.net/faith/blog/2022/08/descriptors-are-hard/
- https://medium.com/@williscool/vulkan-descriptor-buffers-redux-603ea2be3979
- https://www.khronos.org/blog/vk-ext-descriptor-buffer

Todo:

- [x] blogpost
- [x] changelog
- [x] merge bindings
    - [x] https://github.com/floooh/sokol-zig/pull/143
    - [x] https://github.com/floooh/sokol-odin/pull/35
    - [x] nim (no changes needed)
    - [x] https://github.com/floooh/sokol-rust/pull/37
    - [x] https://github.com/floooh/sokol-d/pull/87
    - [x] https://github.com/floooh/sokol-c3/pull/10
    - [x] https://github.com/colinbellino/sokol-jai/pull/9
- [x] rename `swapchain_next()` back to `swapchain()`
- [x] merge sokol-shdc PR: https://github.com/floooh/sokol-tools/pull/196
- [x] test sokol-app wgpu changes
- [x] fix sync-validation errors:
    - [x] swapchain images:
        - [x] no-msaa
        - [x] msaa
    - [x] offscreen rendering images:
        - [x] no msaa
        - [x] msaa
    - [x] other...
        - [x] write-storageimage
        - [x] instancing: errors
        - [x] instancing-pull: errors
        - [x] instancing-compute
        - [x] dyntex-sapp
- [x] sg_desc: move backend-specific options into nested structs
- [x] ~~in sokol_app.h/_sapp_vk_create_swapchain(): delete depth-stencil and msaa surface!!~~ (actually already happened, added a comment for clarification)
- [x] work through and fix deprecations: https://docs.vulkan.org/guide/latest/deprecated.html
- [x] ~~figure out why RenderDoc doesn't work~~ when running in RenderDoc 1.41, device creation fails because of EXT_descriptor_buffer (even though RenderDoc claims the extension is supported)
- [x] track vulkan specific frame stats:
- [x] util headers:
     - [x] sokol_debugtext.h
     - [x] sokol_gl.h
     - [x] sokol_spine.h
     - [x] sokol_imgui.h
     - [x] sokol_fontstash.h
     - [x] sokol_nuklear.h
     - [x] sokol_gfx_imgui.h (only after adding Vulkan-specific frame-stats)
         - [x] shader panel (spirv bindslots)
         - [x] vulkan specific frame stats
         - [x] backend string
- [x] test image staging in chunks with very small staging buffer (tested in cgltf-sapp with 16 KByte staging buffer)
- [x] test buffer staging in chunks with very small staging buffer (tested in cgltf-sapp with 16 KByte staging buffer)
- [x] cgltf-sapp problems:
    - [x] uniform descriptor size 0 (fixed by accumulating sg_apply_uniform calls until draw/dispatch call)
    - [x] compressed texture update problems
- [x] sapp_desc: move backend-specific options into nested structs
- [x] deletion queue! similar to the Metal backend, it's illegal to destroy resource objects while in flight. Must implement a deletion queue which deletes resources after 'inflight-frames'
- [x] figure out why triangle-bufferless is actually upside down...
    - see: https://www.saschawillems.de/blog/2019/03/29/flipping-the-vulkan-viewport/
- [x] swapchain
    - [x] color buffer
    - [x] depth buffer
    - [x] msaa buffer
- [x] begin-pass/end-pass
    - [x] swapchain pass - no depth buffer
    - [x] swapchain pass - with depth buffer
    - [x] swapchain pass - with msaa buffer
    - [x] offscreen pass
    - [x] compute pass
- [x] basic features
    - [x] clear loop (no depth buffer)
    - [x] bufferless triangle
    - [x] triangle-sapp sample (immutable vertex buffers, sg_apply_bindings for vertex buffers)
    - [x] quad-sapp (+ immutable index buffers, + sg_apply_bindings for index buffers)
    - [x] bufferoffsets-sapp (+ immutable buffers with offset, +  multiple draws)
    - [x] cube-sapp sample (+ depth buffer, + uniform data)
    - [x] texcube-sapp sample (+ immutable textures and texture views)
- [x] potential descriptor solutions:
    - [x] ~~`VK_KHR_push_descriptor` core since Vulkan 1.4~~ can't be in the same descriptor layout as descriptor buffer descriptors
    - [x] `VK_EXT_descriptor_buffer` ... availability?
- [x] frame-command-buffer strategy:
    - [x] double-buffered command buffer
    - [x] current command buffer is acquired in first call to:
        - [x] sg_update/append_buffer()
        - [x] sg_update_image()
        - [x] sg_begin_pass()
        - ...this also means that update-staging-transfers go through the regular frame command buffer
    - [x] current command buffer is submitted and set to null in sg_commit()
    - [x] it's ok to record resources into the delete queue in the gap between acquire-command-buffer and sg_commit(), since these resources cannot be recorded into the next command buffer, so it's ok to destroy them in the context of the previous command buffer (e.g. as if those resources would be used in the last command buffer)
- [x] copy-staging-strategy (for new resources), simple and straightforward
    - [x] use for newly created immutable and dynamic-update resources
    - [x] a single small-ish staging buffer (4..16 MB?)
    - [x] a single dedicated init command buffer
    - [x] copy data to init-staging-buffer, record data copy, submit and wait, repeat if data.size > init-staging-buffer
    - [x] ~~alignment for transfer operations?~~
    - [x] ~~also round up staging buffer size for transfer alignment~~
- [x] stream-staging-strategy:
    - [x] used for stream-update resources
    - [x] ~~generally fall back from 'stream-update' to 'copy-update' if per-frame-stream-staging buffer would overflow~~ (no, just log an error)
    - [x] `usage.dynamic` resources are updated like newly created resoruces and cause a pipeline flush, there are no size update-restrictions
    - [x] `usage.streaming` resources are updated through the per-frame staging buffers and are recorded into a separate command buffer which is acquired at the same time as the frame command buffer but is enqueued in front of the frame command buffer (so that streaming updates can be performed "inside" begin/end pass, but are actually executed before the regular frame commands.
    - [x] ~~alignment for transfer operations?~~ not needed, but may be non-optimal for image staging
    - [x] ~~also round up staging buffer size for transfer alignment~~
- [x] uniform-update-strategy
    - [x] only update uniform descriptor buffer and set descriptor buffer offset once in draw/dispatch function if dirty, otherwise the same descriptor is set multiple times when there's more than one uniform buffer in a drawcall!
    - [x] uniform updates shouldn't go through a staging buffer but instead the double-buffered per-frame uniform buffer should live in host-accessible memory
    - [x] ~~use push-descriptor-set for recording uniform updates in sg_apply_uniforms() (unfortunately only one push-descriptor-set allowed, so regular resource bindings needs to go through regular descriptor-sets~~ (switched to descriptor buffer approach)
- [x] resource binding:
    - [x] expect VK_EXT_descriptor_buffer
- [x] barrier strategy:
    - [x] note: can not simply skip barriers if `cur_access == new_access`, if cur_access is a write access (e.g. writes to storage images and storage buffers in compute passes ('write-after-write')
    - [x] begin-pass
    - [x] end-pass:
    - [x] sg_apply_bindings:
    - [x] staging:
        - [x] init/copy:
            - [x] buffers: no barriers needed
            - [x] images: before copy => transfer_dst, after copy => texture
        - [x] streaming:
            - [x] ...?
- [x] sg_update_buffer:
    - [x] streaming
    - [x] dynamic
    - [x] streaming buffer overflow
- [x] sg_append_buffer:
    - [x] streaming
    - [x] dynamic
- [x] sg_update_image:
    - [x] streaming
    - [x] dynamic
- [x] sg_apply_bindings:
    - [x] apply vertex buffers
    - [x] apply index buffer
    - [x] apply views and samplers
- [x] swapchain resize problem:
    - when calling `vkAcquireNextImageKHR()` the surface extent reported by `vkGetPhysicalDeviceSurfaceCapabilitiesKHR()` changes mid-call during window resizes, e.g. calling `vkGetPhysicalDeviceSurfaceCapabilitiesKHR()` reports a different surface extent than after calling `vkAcquireNextImageKHR()` while `vkAcquireNextImageKHR()` returning VK_SUCCESS, this size difference cannot be caught to recreate the swapchain, but will then cause a validation layer error in begin-render because the swapchain image view sizes don't match... 
    - waiting for the fence before acquiring the next swapchain image doesn't help
    - vkAcquireNextImageKHR() and vkQueuePresentKHR() never return with ERROR_OUT_OF_DATE or other error
    - also happens on an X11 session
    - port to Windows and also check there!
    - also: https://docs.vulkan.org/spec/latest/chapters/VK_KHR_surface/wsi.html
    - [x] new attempt (yay that actually works great)
        - _sapp_x11_update_dimensions never updates framebuffer size on vk or emits resized event
        - only _sapp_vk_create_swapchain updates framebuffer size
        - _sapp_vk_recreate_swapchain() emits resized event if framebuffer size had been changed 
          by _sapp_vk_create_swapchain()
        - _sapp_vk_recreate_swapchain() is only called on vkQueuePresentKHR() and vkAcquireNextImageKHR() results
        - once that is validation-layer-clean figure out what to do about zero-sized framebuffer (need to communicate invalid swapchain to sokol-gfx and skip rendering in pass)
- [x] working samples:
    - [x] clear
    - [x] triangle
    - [x] triangle-bufferless
    - [x] quad
    - [x] bufferoffsets
    - [x] cube
    - [x] noninterleaved
    - [x] texcube
    - [x] vertexpull
    - [x] vertexindexbuffer
    - [x] vertextexture
    - [x] sbuftex
    - [x] drawex
    - [x] shapes
    - [x] shapes-transform
    - [x] offscreen
    - [x] offscreen-msaa
    - [x] instancing
    - [x] instancing-pull
    - [x] mrt
    - [x] mrt-pixelformats
    - [x] arraytex
    - [x] tex3d
    - [x] dyntex3d
    - [x] dyntex
    - [x] basisu
    - [x] cubemap-jpeg
    - [x] cubemaprt
    - [x] miprender
    - [x] layerrender
    - [x] primtypes
    - [x] uvwrap
    - [x] mipmap
    - [x] shared-bindings
    - [x] customresolve
    - [x] uniformtypes
    - [x] blend
    - [x] blend-op
    - [x] sdf
    - [x] shadows
    - [x] shadows-depthtex
    - [x] imgui
    - [x] imgui-dock
    - [x] imgui-highdpi
    - [x] cimgui
    - [x] imgui-images
    - [x] imgui-usercallback **doesn't take dpi_scale into account, but that's unrelated**
    - [x] events
    - [x] cursor
    - [x] icon
    - [x] droptest
    - [x] drawcallperf:
         - [x] increase descriptor-buffer-size
         - [x] disable validation layer in release mode!
         - [x] test in release mode
    - [x] texview
    - [x] nuklear 
    - [x] nuklear-images
    - [x] fontstash
    - [x] fontstash-layers
    - [x] debugtext
    - [x] debugtext-printf
    - [x] debugtext-userfont
    - [x] debugtext-context
    - [x] debgtext-layers
    - [x] texview
    - [x] restart
    - [x] sgl
    - [x] sgl-lines
    - [x] sgl-points
    - [x] sgl-context
    - [x] sgl-microui
    - [x] loadpng
    - [x] plmpeg
    - [x] cgltf
    - [x] ozz-anim
    - [x] ozz-skin
    - [x] ozz-storagebuffer
    - [x] shdfeatures
    - [x] spine-simple
    - [x] spine-inspector
    - [x] spine-layers
    - [x] spine-skinsets
    - [x] spine-switch-skinsets
    - [x] spine-contexts
    - [x] sbufoffset
    - [x] instancing-compute
    - [x] computeboids
    - [x] write-storageimage
    - [x] imageblur

### Update 2 notes

- add tests for _sg_track_t funcs
- Windows support
- pretty cool idea inspired by Josh SImmons: record 'pre-bind-barriers' inside passes into a separate command buffer to move them out of vkBegin/EndRendering (e.g. exactly same idea as with the stream-staging-commands. This will need per-pass command-buffers though instead of a single command buffer for the whole frame.
- related: consider using separate command buffers for compute commands and offscreen passes? Those wouldn't need to wait for the present-complete-semaphore!
- support multiple swapchain passes (need to handle multiple present-complete-semaphores)
- implement object labels (only makes sense once RenderDoc actually works)
- attempt to allocate uniform and descriptor buffers in host-visible + device-local memory (and fallback to host-visible if such a heap isn't available)
- native resource injection
- `sg_vulkan_*_info` objects and functions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,38 @@
 - sokol_gfx.h gl: unused framebuffer attachment slots are now explicitly cleared
   in `sg_begin_pass()`. See PR https://github.com/floooh/sokol/pull/1390 for details! Many thanks to @etherbound-dev
   for catching and fixing the issue!
+- sokol_app.h: some minor breaking changes in the public API (most code won't be affected):
+  - the platform/backend specific config items in `sapp_desc` have been moved
+    into nested structs to allow a more consistent designated-init coding style
+  - a new enum `sapp_pixel_format` has been added and the functions `sapp_color_format()` and
+    `sapp_depth_format()` now return `sapp_pixel_format` instead of an `int`. Note
+    that the `sapp_pixel_format` values are no longer directly convertible to `sg_pixel_format`
+    values, you need some mapping code between the two enum types
+  - a new function `sapp_get_environment()` has been added which returns a new struct
+    `sapp_environment()`, this conceptually plugs into the sokol-gfx struct `sg_environement`
+    but requires mapping code to translate `sapp_environment` into `sg_environment` (an example
+    of such mapping code is for instance in sokol_glue.h)
+  - likewise, a new function `sapp_get_swapchain()` has been added which returns a new
+    struct `sapp_swapchain`, which conceptually plugs into the sokol-gfx struct `sg_swapchain`
+  - a ton of platform/backend specific functions have been removed which returned type-erased
+    pointers to various 3D backend objects, those pointers have moved into the new
+    structs `sapp_environemnt` and `sapp_swapchain`
+- sokol_gfx.h: a similar minor breaking change in `sg_desc` as was implemented in `sapp_desc`:
+  the individual backend-specific config items have been moved into nested structs
+  to allow a more consistent designated-init style
+- An *experimental Vulkan backend* has been added to sokol_app.h and sokol_gfx.h,
+  [please read this blog post for an overview](https://floooh.github.io/2025/12/01/sokol-vulkan-backend-1.html).
+  Currently the backend only supports Linux and has only been tested on an Intel
+  Meteor Lake embedded GPU. If you want to test this with the sokol-samples:
+  - make sure the following packages are installed (this is for Ubuntu + apt):
+    - libvulkan-dev
+    - vulkan-tools
+    - vulkan-validationlayers
+  - clone https://github.com/floooh/sokol-samples
+  - from within sokol-samples:
+    - ./fips set config sapp-vk-linux-ninja-debug
+    - ./fips build
+    - ./fips run cube-sapp-ui
 
 ### 13-Nov-2025
 


slendidev:
GG!

PR #1354, opened 2025-10-16
sokol_app.h: fix disabling vsync not working

jorbakk:
Disabling vsync by setting `sapp_desc.swap_interval = 0` has no effect. The default value for `swap_interval` is 1, if not explicitely set (or if set to 0, which is the same).

Possible solutions from my point of view:
1. Set the default value to 0, then vsync is disabled by default. This might be confusing and inconvenient having to activate it explicitly for most of the use cases.
2. Introduce -1 as the value for disabling vsync, which looks like the preferred option. A patch is attached.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1842,7 +1842,7 @@ typedef struct sapp_desc {
     int width;                          // the preferred width of the window / canvas
     int height;                         // the preferred height of the window / canvas
     int sample_count;                   // MSAA sample count
-    int swap_interval;                  // the preferred swap interval (ignored on some platforms)
+    int swap_interval;                  // the preferred swap interval (ignored on some platforms), default: 1, disable vsync: -1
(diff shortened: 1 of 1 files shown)


floooh:
Sorry but I'll need to close the PR.

Disabling vsync currently isn't supported in sokol_app.h, even if it might accidentially work on some backends. I'm not sure yet what the proper solution will look like because the details are very backend-specific.

One thing that's safe to assume is that the `swap_interval` will go away in favour of a target-frame-rate (because swap interval doesn't make much sense when some systems swap at 60Hz and others at 120Hz, disabling vsync will probably be via some sort of 'swap strategy' enum. No ETA yet though.

jorbakk:
Okay. Thanks for your explanation!

Looking forward to the final solution. At least for me, it's quite useful to make quick comparisons of frame times with vsync disabled.

Thanks for your excellent and tireless work on this great project!

Issue #1353, opened 2025-10-15
Suggestion: Don't escape the key strings in sokol_args


cloudwu:
The document says 
> The 'key' string must be a simple string without escape sequences or whitespace.

https://github.com/floooh/sokol/blob/master/sokol_args.h#L89

I found when I pass the windows style pathname from the command line , sokol_args still escape the `\` .

For example, 

```
myapp.exe  .\main.lua
```

`sokol_args(1)` returns `.main.lua` rather than `.\main.lua` .

floooh:
Hmm you're right, the documentation seems to contradict the behaviour.

I'm a bit hesitant changing the code though because the arg-parser state machine isn't trivial and easy to mess up:

https://github.com/floooh/sokol/blob/8c06b5e266f415c1c48b2e668c03df691950c015/sokol_args.h#L597-L668

...if you don't need easy support for parsing args on a web version (e.g. you only care about native scenarios) I would actually recommend using a different arg parser, for instance in sokol-shdc I'm using this getopt library:

https://github.com/wc-duck/getopt

...easy to integrate, and much more declarative, e.g. you describe your cmdline options like this:

https://github.com/floooh/sokol-tools/blob/701f157ed99162b6f296501dce51bf29a893b9bc/src/shdc/args.cc#L35-L55

...and then loop over them like this:

https://github.com/floooh/sokol-tools/blob/701f157ed99162b6f296501dce51bf29a893b9bc/src/shdc/args.cc#L180-L190

cloudwu:
> Hmm you're right, the documentation seems to contradict the behaviour.
> 
> I'm a bit hesitant changing the code though because the arg-parser state machine isn't trivial and easy to mess up:
> 

~~I think we don't need change the state machine, just restore the key string when the value string is empty. I can create a PR if you wish.~~

~~https://github.com/cloudwu/sokol/commit/02a55ae28f602218f6920905912f928f4b3b0bdd~~ 

I can also use `argv[index]` directly if I find `sargs_value_at(index)` returns an empty string.

https://github.com/cloudwu/sokol/commit/f457f546d994cb62a787ed675508634bbb9348a0  Is it okay to just add `_sargs_parsing_val()` check ?

It appears to pass all tests, and I add a new test for escape char in key :  https://github.com/cloudwu/sokol/commit/1e8707ae94f0b5e4f079d1834ae67f4435c92fb7

floooh:
[...]

PR #1355, opened 2025-10-18
Don't escape on the key

cloudwu:
See #1353


floooh:
Testing this now...

floooh:
...and merged. I did a bit of testing via https://floooh.github.io/tiny8bit/ which uses sokol_args.h extensively.

Many thanks!

PR #1358, opened 2025-10-19
[audio] added psvita backend

CrackedPixel:
I've been using sokol for audio on desktop+web with great success and wanted to keep using it for vita.

For your consideration, here is an implementation

### DIFF
--- a/sokol_audio.h
+++ b/sokol_audio.h
@@ -42,6 +42,7 @@
     - on Android: aaudio
     - on Windows with MSVC or Clang toolchain: no action needed, libs are defined in-source via pragma-comment-lib
     - on Windows with MINGW/MSYS2 gcc: compile with '-mwin32' and link with -lole32
+    - on Vita: SceAudio
 
     FEATURE OVERVIEW
     ================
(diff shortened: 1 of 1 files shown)


sokol_audio.h
```diff
@@ -982,6 +1011,18 @@ typedef struct {
     uint8_t* buffer;
 } _saudio_web_backend_t;
 
+#elif defined(_SAUDIO_VITA)
+
+typedef struct {
+    int* device;
+    float* buffer;
+    int16_t* buffer_vita;
```
CrackedPixel: this is the only thing i dont really like. to keep the same API and such it needs a secondary buffer as the vita uses shorts instead of floats

floooh:
Nice :) I don't have a way to test this so I can't promise to properly maintain the backend in the long term, oth it's not a lot of code.

I'll need to think about it a bit, and will also need to take of some other pending PRs first.

CrackedPixel:
> Nice :) I don't have a way to test this so I can't promise to properly maintain the backend in the long term, oth it's not a lot of code.
> 
> I'll need to think about it a bit, and will also need to take of some other pending PRs first.

Yeah, it was pretty easy to setup with your api :sunglasses:

floooh:
Ok I just YOLO'd it and merged the PR :)

Normally I wouldn't merge PRs which I cannot really maintain, but I think in this case an exception is justified since it's very little code and it's unlikely that the internal design of sokol_audio.h changes so radically that the code would become stale (in that case I might depend on your help or the backend might have to be removed again).

Thanks for the PR!

Issue #1360, opened 2025-10-21
[sokol_spine]: Slots that use multiply blend mode are not rendered correct


bryanjeal:
As noted in the title, any slots that use the multiply blend mode do not render correctly. I have been able to replicate this with the Spine Chibi Example project.

<img width="1860" height="1124" alt="Image" src="https://github.com/user-attachments/assets/dd56ca3b-8add-474b-ad15-330cc7037eb9" />

<img width="341" height="527" alt="Image" src="https://github.com/user-attachments/assets/aaf5188f-9d49-4a9b-bb9a-02720a95c9e9" />

[chibi-stickers.zip](https://github.com/user-attachments/files/23012626/chibi-stickers.zip)

floooh:
Thanks for the issue and PR!

bryanjeal:
Thank you for creating sokol and sokol_spine! <3

floooh:
Fixed via https://github.com/floooh/sokol/pull/1361

PR #1361, opened 2025-10-21
[sokol_spine] Fixes issue with rendering slots that use multiply blend mode.

bryanjeal:
[sokol_spine] Fixes issue with rendering slots that use multiply blend mode.
#1360 

Pre-fix:
image

With fix applied:
image

### DIFF
--- a/util/sokol_spine.h
+++ b/util/sokol_spine.h
@@ -5024,6 +5024,7 @@ static void _sspine_draw_layer(_sspine_context_t* ctx, int layer, const sspine_l
                     sg_apply_pipeline(cmd->pip);
                     cur_pip_id = cmd->pip.id;
                     sg_apply_uniforms(0, &vsparams_range);
+                    sg_apply_uniforms(1, &fsparams_range);
                     cur_view_id = SG_INVALID_ID;
                 }
                 if ((cur_view_id != cmd->view.id) || (cur_smp_id != cmd->smp.id)) {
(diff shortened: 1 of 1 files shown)


floooh:
Oh, good catch... `ZERO/ONE` was definitely wrong. I guess the examples all use premultiplied alpha, so I didn't catch this.

I'll try to take care of the PR later today.

bryanjeal:
@floooh there was a sokol_gfx validation issue that I had previously fixed, but forgot about. It is still required and is part of my second commit. The PR should be good to review now. :)

floooh:
Hmm, that way the `sg_apply_uniforms()` is running for each draw call, which shouldn't be necessary... what was the validation layer message? Something about sg_appy_uniforms not being called?

Can you try this instead? (move the additional `sg_apply_uniform` up into the block which calls `sg_apply_pipeline`):

```c
                if (cur_pip_id != cmd->pip.id) {
                    sg_apply_pipeline(cmd->pip);
                    cur_pip_id = cmd->pip.id;
                    sg_apply_uniforms(0, &vsparams_range);
                    sg_apply_uniforms(1, &fsparams_range);
                    cur_view_id = SG_INVALID_ID;
                }
                if ((cur_view_id != cmd->view.id) || (cur_smp_id != cmd->smp.id)) {
                    ctx->bind.views[0] = cmd->view;
                    ctx->bind.samplers[0] = cmd->smp;
                    sg_apply_bindings(&ctx->bind);
                    cur_view_id = cmd->view.id;
                    cur_smp_id = cmd->smp.id;
                }
                if (cur_pma != cmd->pma) {
                    fsparams.pma = cmd->pma;
                    sg_apply_uniforms(1, &fsparams_range);
                    cur_pma = cmd->pma;
                }
```

PS: this may be a side effect of more strict validation that might have been added after sokol_spine.h was created, and in my samples this isn't triggered because everything uses `pma`.

The TL;DR is basically that an `sg_apply_pipeline` requires that bindings and uniforms are re-applied before the next draw call.

bryanjeal:
@floooh your suggestion does indeed fix the issue. I cannot remember what the error was and at the moment, I do not have time to replicate it - Sorry!

floooh:
Ok, I gave it a quick whirl with the sokol-sample Spine samples and merged.

I also added a blurb to the changelog.

Thanks!

PR #1372, opened 2025-11-09
[audio] Nintendo 3DS backend

CrackedPixel:
Similar to the psvita port, i was talked into doing a 3ds port and wanted to use sokol (last one, i promise :sweat_smile:)

~~It isnt quite perfect, and with the \~300ms delay it isnt very ideal... but for the game i ported it was good enough. I'm hoping maybe someone more familiar with the 3ds/ndsp would be able to offer suggestions on how to improve it etc.~~
update: the delay is only on emulators, not the real device

I will leave some comments on specific lines for discussion, too

Tested on both an emulator and real hardware. No issues in terms of stability/performance/artifacts/glitches/etc

### DIFF
--- a/sokol_audio.h
+++ b/sokol_audio.h
@@ -422,10 +422,12 @@
     nothing is provided.
 
     There is a known issue of a noticeable delay when starting a new
-    sound due to the NDSP. I was not able to improve this to my liking
+    sound on emulators. I was not able to improve this to my liking
     and ~300ms can be expected. This can be improved by using a lower
     buffer size than the 2048 default but I would not suggest under
     1536. It may crash under 1408, and they must be in multiples of 128.
+    Note: I was NOT able to reproduce this issue on a real device and
+    the audio worked perfectly.
 
 
     MEMORY ALLOCATION OVERRIDE


sokol_audio.h
```diff
@@ -400,6 +402,31 @@
...
+
+    The queue size and other NDSP specific parameters can be chosen by
+    the provided 'user_data_3ds_t' user_data. Defaults will be used if
+    nothing is provided.
+
+    There is a known issue of a noticeable delay when starting a new
+    sound due to the NDSP. I was not able to improve this to my liking
+    and ~300ms can be expected. This can be improved by using a lower
+    buffer size than the 2048 default but I would not suggest under
+    1536. It may crash under 1408, and they must be in multiples of 128.
```
CrackedPixel: This delay could make it annoying for some games that rely on sfx to be played from user input, but due to the queue system of buffers that the NDSP uses i'm really not sure how to improve it. Maybe it cant be

sokol_audio.h
```diff
@@ -2282,7 +2346,151 @@ _SOKOL_PRIVATE void _saudio_vita_backend_shutdown(void) {
...
+//
+// >>3ds
+#elif defined(_SAUDIO_3DS)
+
+/* NDSP triggers a callback for _saudio_3ds_cb on the main thread */
+_SOKOL_PRIVATE void _saudio_3ds_cb(void*) {
+    const user_data_3ds_t* user_data = (user_data_3ds_t*)_saudio.desc.user_data;
+    /* can probably move these 2 below to _saudio.backend? */
+    static const int totalSamples = _saudio.buffer_frames * _saudio.num_channels;
+    static const float scale = 32767.0f;
```
CrackedPixel: should i move these to `_saudio.backend`? they were put here because it took me awhile to get to the current version and i made a LOT of changes during that time. but now that its finalized maybe it makes more sense to move them

sokol_audio.h
```diff
@@ -2282,7 +2346,151 @@ _SOKOL_PRIVATE void _saudio_vita_backend_shutdown(void) {
...
+}
+
+_SOKOL_PRIVATE void _saudio_3ds_ndsptrigger_cb(void*) {
+    if(_saudio.backend.thread_stop) {
+        return;
+    }
+
+    /* ndsp requested more data. trigger cb */
+    _saudio_3ds_cb(NULL);
+}
```
CrackedPixel: more info on this below (`ndspSetCallback`)

sokol_audio.h
```diff
@@ -2282,7 +2346,151 @@ _SOKOL_PRIVATE void _saudio_vita_backend_shutdown(void) {
...
+
+        _saudio.desc.user_data = defaultUserData;
+    }
+
+    user_data_3ds_t* user_data = (user_data_3ds_t*)_saudio.desc.user_data;
+
+    /* clamp to 2 channels max */
+    if (_saudio.num_channels > 2) {
+        _saudio.num_channels = 2;
+    }
```
CrackedPixel: this is a sanity check but i could probably remove it and just expect the user to be reasonable and know what theyre doing

sokol_audio.h
```diff
@@ -2282,7 +2346,151 @@ _SOKOL_PRIVATE void _saudio_vita_backend_shutdown(void) {
...
+
+    /* allocate the streaming buffer */
+    _saudio.backend.buffer_byte_size = _saudio.buffer_frames * _saudio.bytes_per_frame;
+    _saudio.backend.buffer = (float*) _saudio_malloc_clear((size_t)_saudio.backend.buffer_byte_size);
+    _saudio.backend.buffer_3ds = (int16_t*)linearAlloc(user_data->queue_count * _saudio.buffer_frames * _saudio.num_channels * sizeof(int16_t));
+    _saudio.backend.queue_3ds = (ndspWaveBuf*)_saudio_malloc(user_data->queue_count * sizeof(ndspWaveBuf));
+
+    /* prepare the 3ds audio queues */
+    int16_t* bufferPtrCopy = _saudio.backend.buffer_3ds;
+    for (int i = 0; i < user_data->queue_count; ++i) {
```
CrackedPixel: the 3DS uses multiple queues, each (can) point to the same linear buffer at different points

makes it a little trickier than other platforms that manage this stuff internally and just let you pump them with data

sokol_audio.h
```diff
@@ -560,6 +587,8 @@ extern "C" {
     _SAUDIO_LOGITEM_XMACRO(BACKEND_BUFFER_SIZE_ISNT_MULTIPLE_OF_PACKET_SIZE, "backend buffer size isn't multiple of packet size") \
     _SAUDIO_LOGITEM_XMACRO(VITA_SCEAUDIO_OPEN_FAILED, "sceAudioOutOpenPort() failed") \
     _SAUDIO_LOGITEM_XMACRO(VITA_PTHREAD_CREATE_FAILED, "pthread_create() failed") \
+    _SAUDIO_LOGITEM_XMACRO(3DS_NDSP_OPEN_FAILED, "ndspInit() failed") \
+    _SAUDIO_LOGITEM_XMACRO(3DS_USERDATA_DEFAULT_MALLOC_FAILED, "user_data not provided, default malloc failed") \
```
CrackedPixel: the build is failing with these names starting with a digit :sweat_smile: ill fix it

sokol_audio.h
```diff
@@ -2282,7 +2346,151 @@ _SOKOL_PRIVATE void _saudio_vita_backend_shutdown(void) {
...
+    }
+    else {
+        if (0 == _saudio_fifo_read(&_saudio.fifo, (uint8_t*)_saudio.backend.buffer, _saudio.backend.buffer_byte_size)) {
+            /* not enough read data available, fill the entire buffer with silence */
+            _saudio_clear(_saudio.backend.buffer, (size_t)_saudio.backend.buffer_byte_size);
+        }
+    }
+}
+
+_SOKOL_PRIVATE void _saudio_3ds_ndsptrigger_cb(void*) {
```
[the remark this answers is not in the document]
floooh: I agree, the function should be merged into `_saudio_3ds_cb`. Also, I know it's not standard, but I usually don't use the `NULL` macro, instead just `0` for null pointers.

sokol_audio.h
```diff
@@ -1023,6 +1055,36 @@ typedef struct {
...
+    /* NDSP_INTERP_NONE      = 2 (low quality, fast)    */
+    ndspInterpType interpolation_type; /* default value = 0 */
+
+    /* 3DS supports different audio channels. they can be used */
+    /* in a variety of ways as independent streams etc.        */
+    /* this implementation in sokol does NOT allow multiple    */
+    /* due to calling the global ndspInit/ndspExit functions.  */
+    /* valid range 0-23                                        */
+    int channel_id; /* default value = 0 */
+} user_data_3ds_t;
```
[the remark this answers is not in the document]
floooh: This is the one place I would do differently, since these look like 3DS-specific setup options.

The `saudio_desc.user_data` pointer should be reserved for pointing to actual caller-specific userdata (like a pointer to the caller's application state).

I suggest the following changes:

- rename `user_data_3ds_t` to `saudio_3ds_desc` and move it up into the public API in front of the `saudio_desc` declaration
- embed that struct into `saudio_desc` (not ifdef-ed, because the sokol header public APIs are supposed to be the same no matter which backend is selected):

    ```c
    typedef struct saudio_desc {
        int sample_rate;
        // ...
        saudio_3ds_desc _3ds;    // (or maybe '.three_ds' is nicer?
        saudio_allocator allocator;
        saudio_logger logger;
    } saudio_desc;     
    ```

Then add those config values also to the `_saudio_3ds_backend_t` struct and in the `_saudio_3ds_backend_init()` function add a block to fill in the defaults similar to here:

https://github.com/floooh/sokol/blob/ce5f7c96010a76876d4c15f3e618ffbe583991e1/sokol_audio.h#L2348-L2352

...and in the rest of the code simply access those values from the backend struct instead of the user_data pointer.

sokol_audio.h
```diff
@@ -2282,7 +2346,151 @@ _SOKOL_PRIVATE void _saudio_vita_backend_shutdown(void) {
...
+            _SAUDIO_ERROR(N3DS_USERDATA_DEFAULT_MALLOC_FAILED);
+            return false;
+        }
+
+        defaultUserData->queue_count = 2;
+        defaultUserData->interpolation_type = NDSP_INTERP_POLYPHASE;
+        defaultUserData->channel_id = 0;
+
+        _saudio.desc.user_data = defaultUserData;
+    }
```
floooh: As I wrote above, I would do this entire block differently and not use a heap allocation or the user-data pointer.

sokol_audio.h
```diff
@@ -2282,7 +2346,151 @@ _SOKOL_PRIVATE void _saudio_vita_backend_shutdown(void) {
...
+
+    if (_saudio.backend.channel_id >= 0) {
+        ndspChnWaveBufClear(_saudio.backend.channel_id);
+        ndspExit();
+        _saudio.backend.channel_id = -1;
+    }
+
+    _saudio_free(_saudio.backend.queue_3ds);
+    _saudio_free(_saudio.backend.buffer_3ds);
+    _saudio_free(_saudio.backend.buffer);
```
floooh: I think you'd need to also free the optionally allocated user-data block here, but as I said above, those config params should just go into nested structs instead :)

floooh:
I added some comments (should have done a proper code-review in hindsight heh).

I would suggest one important change in how the 3DS specific config options are passed into sokol-audio (not via the user_data pointer, but a nested struct in `saudio_desc`.

sokol_audio.h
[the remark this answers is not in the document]
CrackedPixel: function was merged :heavy_check_mark:

sokol_audio.h
CrackedPixel: heap allocation removed :heavy_check_mark: 

user_data pointer no longer used/needed :heavy_check_mark:

sokol_audio.h
```diff
@@ -608,6 +655,7 @@ typedef struct saudio_desc {
     void (*stream_cb)(float* buffer, int num_frames, int num_channels);  // optional streaming callback (no user data)
     void (*stream_userdata_cb)(float* buffer, int num_frames, int num_channels, void* user_data); //... and with user data
     void* user_data;        // optional user data argument for stream_userdata_cb
+    saudio_n3ds_desc n3ds;       // optional data for use on n3ds
```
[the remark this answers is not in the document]
floooh: No that's fine, really :) When not using the N3DS backend the item can simply be ignored in the saudio_setup() call, and when using it, you don't need to deal with platform-specific #ifdefs in the saudio_setup() call.

E.g. if I'm not using the N3DS backend at all I can continue ignoring the `.n3ds` item:

```c
saudio_setup(&(saudio_desc){
    .logger.func = slog_func
});
```

...and if I'm using the N3DS backend in some multi-platform project, I can just always pass the N3DS options into sokol-audio without having to ifdef the code:

```c
saudio_setup(&(saudio_desc){
    .n3ds = {
        .interpolation_type = ...,
    },
    .logger.func = slog_func
});
```

...and if for instance the PSVita backend would also require config params it could look like this:

```c
saudio_setup(&(saudio_desc){
    .n3ds = {
        .interpolation_type = ...,
    },
    .psvita = {
        // ...
    },
    .logger.func = slog_func
});
```

I'll try to take care of the PR later today :)

floooh:
And merged. Thanks! I'll add a blurb to the changelog next.

Btw I'm starting to wonder whether direct support for 16-bit integer samples would make sense, maybe initially only via the stream callback method (e.g. not the 'push' method via saudio_push()). Basically extending `saudio_desc` by

```c
    void (*streami16_cb)(int16_t* buffer, int num_frames, int num_channels);
    void (*streami16_userdata_cb)(int16_t* buffer, int num_frames, int num_channels, void* user_data);
```

...but maybe int16_t support for `saudio_push()` also wouldn't be a big deal (e.g. a new function `saudio_pushi16(const int16_t* frames, int num_frames)` since it looks like the whole FIFO buffer stuff is byte-based anyway.

In the beginning, backends could simply reject the 'non-native' sample format, later there could be a transparent conversion step.

PR #1374, opened 2025-11-15
win32: Prioritize attaching console, otherwise create

danielchasehooper:
This change makes it possible to get good console behavior regardless of if the app was launched via double-click or from the console. To get the new improved behavior, set both `sapp_desc.win32_console_attach` and `sapp_desc.win32_console_create` to `true`

The previous behavior had these issues:

- Console launch +`sapp_desc.win32_console_create=true` + `sapp_desc.win32_console_attach=false`
-> app-created console would disappear on crash and you can't see error logs.

- Double-click launch + `sapp_desc.win32_console_attach=true` 
-> you can't see console output

New behavior when both `sapp_desc.win32_console_create` and `sapp_desc.win32_console_attach` are true:
- Console launch: outputs to console.
- Double-click launch: creates new console


floooh:
Looks good! I'll try to give it a whirl on Monday.

floooh:
Working nicely. Thanks!

PR #1375, opened 2025-11-15
[audio] fixed issue in 3ds shutdown

CrackedPixel:
The original code worked 99% of the time, but I did have 1 strange test case where it would crash. Changing it to always exit the ndsp solved the problem and 100% of my tests are fine now

Sorry about that :sweat_smile: i only recently got the hardware and have been testing more thoroughly


floooh:
Thanks! :)

PR #1381, opened 2025-11-19
update binding to add sokol_gfx_imgui.h

robboito:
Only tested with Zig as that is what I have running at the minute.


floooh:
...linking my comment https://github.com/floooh/sokol-zig/issues/129#issuecomment-3552007282

floooh:
FYI I have merged the sokol_gfx_imgui.h cleanup which makes the changes in the gen-bindings script unnecessary. I'll close this PR and will do the other needed changes (just adding the sokol_gfx_imgui.h header when taking care of https://github.com/floooh/sokol-zig/pull/142

PR #1392, opened 2025-12-03
sokol_app add vulkan support on windows platform

THISISAGOODNAME:


floooh:
Apologies for not merging the PR, I prefer to do the remaing Vulkan work myself on my own pace ;)

PS: I'm actually surprised that it is so little code though ;)

PR #1396, opened 2025-12-06
fix typo in _sg_gl_create_shader, leading to wrong bind slots being used

Julianiolo:
There is a typo in `_sg_gl_create_shader` which causes the shader storage image binding number (`shd->gl.simg_binding[i]`) to be assigned the storage buffer number instead (which is always zero in this situation, since the shader view is actually a storage image view).


floooh:
Whoops, good catch. Thanks!

PR #1410, opened 2026-01-12
D3D12: implement backend

oviano:
Hello @floooh

Long-time sokol user here. Thanks for the great library!

I know you've mentioned D3D12 is better suited for sokol_gfx2. That said, I had a need for it and have a complete implementation that passes all samples. Offering it in case it's useful, with no expectation to merge.

Even if it's not right for mainline, others may find it helpful?

Summary

  - Full D3D12 backend implementation for sokol_gfx.h
  - Supports all existing sokol features including compute shaders
  - Passes all unit tests and samples (d3d12-specific samples in separate sokol-samples PR)
  - Shader compilation via d3dcompiler_47.dll (same approach as D3D11)
  - Performance: ~5-25% faster than D3D11 depending on workload

Design

The implementation follows D3D12 best practices rather than being a thin wrapper over D3D11 patterns:

  - Double-buffered frames - per-frame command allocators for CPU/GPU overlap
  - Fence-based synchronization - application provides fence via sg_environment
  - Root signatures - per-pipeline, using descriptor tables and root CBVs
  - Pipeline state objects - full PSO with caching and deferred release
  - Descriptor heaps - CPU-side for creation, GPU-visible for binding (configurable sizes)
  - Uniform buffer ring - per-frame with 256-byte alignment
  - Upload ring buffer - per-frame staging for image updates
  - Resource barriers - automatic state tracking and transitions
  - Deferred release queue - fence-based safe resource cleanup

Written in C99-compatible style consistent with the other backends.

Implementation size: ~4,600 lines (compared to D3D11's ~2,300 lines). The larger size reflects D3D12's explicit resource management model.

Changes

  - sokol_gfx.h - D3D12 backend implementation
  - sokol_app.h - D3D12 swapchain/window integration
  - sokol_glue.h - D3D12 environment/swapchain helpers
  - util/sokol_*.h - D3D12 support in utility headers
  - tests/ - D3D12 test configurations and presets

Testing

  - All 322 unit tests pass (Debug)
  - Static analysis clean (clang --analyze)
  - Tested on Windows with NVIDIA RTX 3060

Note

The language binding examples (sokol-odin, sokol-rust, etc.) will need their switch/match statements updated to handle the new `D3D12` backend case.

Related PRs

https://github.com/floooh/sokol-samples/pull/188
https://github.com/floooh/sokol-tools/pull/206

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1893,7 +1893,7 @@ typedef struct sapp_d3d12_environment {
     const void* command_queue;
     const void* fence;
     void* fence_event;
-    uint64_t* fence_value;
+    void* fence_value;              // uint64_t*, passed as void* for language binding compatibility
 } sapp_d3d12_environment;
 
 typedef struct sapp_wgpu_environment {
@@ -14461,7 +14461,7 @@ SOKOL_API_IMPL sapp_environment sapp_get_environment(void) {
         res.d3d12.command_queue = (const void*) _sapp.d3d12.command_queue;
         res.d3d12.fence = (const void*) _sapp.d3d12.fence;
         res.d3d12.fence_event = (void*) _sapp.d3d12.fence_event;
-        res.d3d12.fence_value = &_sapp.d3d12.fence_value;
+        res.d3d12.fence_value = (void*)&_sapp.d3d12.fence_value;
     #endif
     #if defined(SOKOL_WGPU)
         res.wgpu.device = (const void*) _sapp.wgpu.device;
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -5087,7 +5087,7 @@ typedef struct sg_d3d12_environment {
     const void* command_queue;
     const void* fence;
     void* fence_event;
-    uint64_t* fence_value;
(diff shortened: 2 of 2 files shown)


floooh:
Impressive work, but this is too big to review/merge (as you expected though ;)

I'll keep the PR open though for other people to reference, and maybe also as reference for a time when an official D3D12 backend might be implemented.

PS: when I got some more time I'll try to play around with the PR a bit. E.g. when the performance is comparable to the D3D11 backend, that might be a good reason to start thinking about replacing the D3D11 backend with D3D12 - mainly because D3D12 has a more modern shader compiler infrastructure...

floooh:
PS: the failing CI is about missing switch-prongs for a switch on the backend enum in some of the language samples... nothing critical.

oviano:
Yup, all understood and that makes sense.

Essentially...use it as you see fit (or not!).

oviano:
I added my stress test to a new branch, which might be useful if/when you come to experiment.

https://github.com/oviano/sokol/tree/d3d12-backend-with-stress-test

PR #1412, opened 2026-01-14
fix: GLES3 FBO depth attachment switching

luigi-rosso:
On some GLES3 drivers (tested on Pixel 8 Pro with Mali GPU), switching between GL_DEPTH_ATTACHMENT and GL_DEPTH_STENCIL_ATTACHMENT on the same FBO causes GL_FRAMEBUFFER_UNSUPPORTED, even when attaching valid textures.

When sokol attaches a new depth/depth-stencil texture, it doesn't first detach the other attachment type. Some GLES3 drivers fail if both attachment points have ever been used on the same FBO.

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -11365,16 +11365,14 @@ _SOKOL_PRIVATE void _sg_gl_begin_pass(const sg_pass* pass, const _sg_attachments
             glFramebufferTexture2D(GL_FRAMEBUFFER, gl_att_type, GL_TEXTURE_2D, 0, 0);
         }
         if (atts->ds_view) {
-            // GLES3 FBO fix: When switching between depth-only and depth-stencil attachments,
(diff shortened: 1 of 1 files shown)


floooh:
Thanks! The change makes sense. I'll try to give it a whirl and merge in the next few days (mainly checking if it works as expected on WebGL2).

luigi-rosso:
I just repro-ed this same issue on the SteamDeck with SOKOL_GLCORE. So it might be necessary to completely remove the if def...

floooh:
> I just repro-ed this same issue on the SteamDeck with SOKOL_GLCORE. So it might be necessary to completely remove the if def...

Good to know, can you remove the ifdefs in the PR?

floooh:
Thanks for the fix! It'll probably be until Monday before I get around to test and merge though.

floooh:
Tested on macOS and WebGL2 and merged. Thanks!

PR #1414, opened 2026-01-15
fix: enable glColorMaski on GLES 3.2+

luigi-rosso:
I noticed this wasn't working on Android, since so many modern devices now support this I thought it'd be nice to contribute it.

I check for API level 24 (Android 7.0 in 2016 added gl32.h to the NDK) just in case since I think you're supporting 21+.

Could do this for base vertex too!

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -5579,7 +5579,11 @@ inline int sg_append_buffer(sg_buffer buf_id, const sg_range& data) { return sg_
                 #include <GLES3/gl3.h>
             #endif
         #elif defined(__ANDROID__)
-            #include <GLES3/gl31.h>
+            #if __ANDROID_API__ >= 24
+                #include <GLES3/gl32.h>
+            #else
+                #include <GLES3/gl31.h>
+            #endif
         #elif defined(__linux__) || defined(__unix__)
             #if defined(SOKOL_GLCORE)
                 #define GL_GLEXT_PROTOTYPES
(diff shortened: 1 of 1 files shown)


floooh:
I wonder if we can reduce the number of code paths a bit more... I'll try to tinker with that a bit when I get around to properly review and test the code (I'll try to do that on Monday).

floooh:
Ok, merged but with some changes (I didn't actually test on Android since I can't currently debug on my phone for some mysterious reason):

https://github.com/floooh/sokol/commit/d147bc3b1b3f2b6a10cf43df4ddc5ceb79597399

...what I changed:

- also define `_SOKOL_GL_HAS_COLORMASKI` for all desktop GL platforms which allows simplifications further down
- on Android, use the `GL_ES_VERSION_3_2` macro to decide whether to define _SOKOL_GL_COLOR_MASKI (https://github.com/floooh/sokol/commit/d147bc3b1b3f2b6a10cf43df4ddc5ceb79597399#diff-2e1da1beda27f21ea55e93ee52191efb90c1a1794a65aa34dce1db773a65332bR5624-R5626)
- move the unrelated `mrt_independent_blend_state` out of the conditional code block and always set it to false: https://github.com/floooh/sokol/commit/d147bc3b1b3f2b6a10cf43df4ddc5ceb79597399#diff-2e1da1beda27f21ea55e93ee52191efb90c1a1794a65aa34dce1db773a65332bR10002-R10005
- simplify the actual `glColorMask` vs `glColorMaski` selection: https://github.com/floooh/sokol/commit/d147bc3b1b3f2b6a10cf43df4ddc5ceb79597399#diff-2e1da1beda27f21ea55e93ee52191efb90c1a1794a65aa34dce1db773a65332bR11726-R11742

...tested via the basisu-sapp sample on macOS and WebGL2.

floooh:
...oops looks like I broke the iOS build... one sec.

PR #1416, opened 2026-01-18
sapp: Add window position and screen size APIs

squk:
EDIT: Checkout https://github.com/squk/sokol_utils
- Added sapp_set_window_position() to set the window top-left position.
- Added sapp_screen_width() and sapp_screen_height() to query screen dimensions.
- Win32 implementation uses MonitorFromWindow to return dimensions of the current monitor.

:)

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2176,6 +2176,12 @@ SOKOL_APP_API_DECL void sapp_set_clipboard_string(const char* str);
 SOKOL_APP_API_DECL const char* sapp_get_clipboard_string(void);
 /* set the window title (only on desktop platforms) */
 SOKOL_APP_API_DECL void sapp_set_window_title(const char* str);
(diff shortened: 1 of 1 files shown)


floooh:
Thanks for the effort, I'll keep the PR open for other to use, but will not merge it into the main version. There will eventually be more control over window positioning, but it will likely be implemented alongside multiwindow support (which is more straightforward now with the re-designed sokol begin/end-pass functions.

Another way to add missing functionality like this is to either create an 'extension header' which is included after the sokol_app.h implementation (and thus has access to private sokol_app.h functions and state), or create a separate library or helper functions which use the platform specific window handles (e.g. `sapp_win32_get_hwnd()` etc...).

squk:
Thanks for taking a look and the advice!

squk:
I created a repo to house my extensions : https://github.com/squk/sokol_utils

PR #1418, opened 2026-01-20
feature: expose Metal CommandQueue

luigi-rosso:
Adds sg_mtl_command_queue similar to other Metal specific API calls.

I've been playing around with integrating the Rive Renderer with Sokol, and one of the things I currently have to do on Metal is use two CommandQueues and sync with `[commandBuffer waitUntilCompleted];` to not get flickering. If I could grab Sokol's CommandQueue I could avoid the cpu stall.


floooh:
Looks good, thanks!

PR #1423, opened 2026-01-24
Add support for dual source blending

Tremus:
Comment by @floooh: Not closed but actually merged via https://github.com/floooh/sokol/pull/1426! See https://github.com/floooh/sokol/pull/1423#issuecomment-3800032783 (TL;DR: I messed up)

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -2726,6 +2726,10 @@ typedef enum sg_blend_factor {
     SG_BLENDFACTOR_ONE_MINUS_BLEND_COLOR,
     SG_BLENDFACTOR_BLEND_ALPHA,
     SG_BLENDFACTOR_ONE_MINUS_BLEND_ALPHA,
+    SG_BLENDFACTOR_SRC1_COLOR,
+    SG_BLENDFACTOR_ONE_MINUS_SRC1_COLOR,
+    SG_BLENDFACTOR_SRC1_ALPHA,
+    SG_BLENDFACTOR_ONE_MINUS_SRC1_ALPHA,
     _SG_BLENDFACTOR_NUM,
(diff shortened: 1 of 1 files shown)


floooh:
Ok, CI fails on the GLES3 platforms, but that's expected. I'll take care of that during the merge (the failing Linux ci is because it's possible to build against the GLES3 headers on Linux).

floooh:
...starting to look into this now...

floooh:
Btw: dual-source-blending on Vulkan is optional, and either requires an extension or needs to be enabled via a feature flag (haven't checked yet, but see this WebGPU discussion: https://github.com/gpuweb/gpuweb/issues/391).

I'm adding changes on top of your PR in a branch I created from it, when this is merged this PR here should also be automatically be marked as merged (I think).

My general plan is:

- add `sg_feature.dual_source_blending` feature flag
- add a validation layer check against this flag
- fix the GLES3 code to not depend on the dual source constants
- figure out what's needed on Vulkan

...at some later point I will replace the `blend-sapp` and `blend-op-sapp` samples with a general 'blending sample' which has a Dear ImGui UI to play around with all blending features.

PS: wip PR which builds on top of your PR: https://github.com/floooh/sokol/pull/1426

floooh:
Ah dang... 

So the feature has been merged via https://github.com/floooh/sokol/pull/1426, unfortunately I clicked `Squash and merge` which was preselected from another PR, and this caused your commits to not be included in the git history and Github also didn't see the connection to this original PR and didn't mark this PR as merged. Instead I'll need to manually close it without merging.

Sorry about that, I should have been more careful. I mentioned you at least in the changelog though.

Tremus:
No worries mate, I'm just glad the feature is there. Thanks for all your hard work :)

Issue #1370, opened 2025-11-03
HDR/EDR frame buffer best practices?


framkant:
Hi (and thanks for a great toolkit!),  I am trying to get EDR/HDR rendering working on a small app but can't  get my head around how it is supposed to work when using sokol.   I dev on Mac(but the goal is cross platform) and when using Metal + GLFW I have old code where I achieved HDR/EDR rendering like this:

`
NSWindow *nswin = glfwGetCocoaWindow(window);

CAMetalLayer *layer = [CAMetalLayer layer];
layer.device = device;
layer.pixelFormat = MTLPixelFormatRGBA16Float;//  std is MTLPixelFormatBGRA8Unorm;
const CFStringRef name = kCGColorSpaceExtendedLinearDisplayP3;
CGColorSpaceRef colorspace = CGColorSpaceCreateWithName(name);
layer.colorspace = colorspace;
nswin.contentView.layer = layer;
nswin.contentView.wantsLayer = YES;
layer.wantsExtendedDynamicRangeContent = YES;
`

But no code similar to  this is available anywhere in sokol as far as I can tell so I assume I have to add/hack it myself? If so, is there any recommended ways to approach it?

floooh:
Can't be much of a help with this there unfortunately, I haven't tinkered with HDR framebuffers yet. 

Currently sokol_app.h is using MTKVIew instead of managing a CAMetalLayer directly (however that's going to change at some point), that's probably the reason why you didn't find the above CAMetalLayer configuration, I'm not sure if MTKView exposes settings to configure a HDR framebuffer (from quickly looking at the docs it doesn't look like it).

framkant:
Ok, thanks for a super quick answer. I had a look around and it seems you might be able to get the layer from the MTKView . However to have it play nice there is also some other places to change code in sokol_glue and sokol_app that is hardcoded to return BGR8 . (like "sapp_color_format()" etc )

Anyway: adding it as a quick hack like this:
`

        _sapp.macos.view.colorPixelFormat = MTLPixelFormatBGRA8Unorm;
        _sapp.macos.view.depthStencilPixelFormat = MTLPixelFormatDepth32Float_Stencil8;
        _sapp.macos.view.sampleCount = (NSUInteger) _sapp.sample_count;
        _sapp.macos.view.autoResizeDrawable = false;
        _sapp.macos.window.contentView = _sapp.macos.view;
        [_sapp.macos.window makeFirstResponder:_sapp.macos.view];
[...]

PR #1449, opened 2026-02-25
sapp: add HDR support and runtime pixel format switching API

squk:
At first I implemented basic HDR support through a bool flag on sapp_desc in REMINDER_TO_INSERT_PR_LINK. Then I saw your comment in https://github.com/floooh/sokol/issues/1370 stating sRGB support should be implemented first so I decided to just have the API use that type and handle the Metal HDR flag in the platform specific impl. 

In addition to the field on sapp_desc, I decided to add an API for switching the swapchain format at runtime, since most games/applications provide this option without needing to restart the application to see its effects(which also allows the user to confirm its working on their hardware configuration before persisting to the apps on-disk settings, potentially breaking the app at next start). 

Draft since it's only working on Mac currently.


floooh:
Thanks for the PR, but I'd like to tackle that myself in an update which generally adds more config options for the swapchain and does a general cleanup round for 'display parameters' in sapp_desc (like getting rid of some things that didn't work out - like `swapInterval`, this should instead be something like `targetFrameRate`).

I'll keep the PR open, but won't merge it (I'll most like use it for 'inspiration' though).

I haven't thought much yet about being able to reconfigure the swapchain after initialization, but that's definitely needed for games that want to avoid the "restart the game for setting changes to take effect", so I'm very much in favour - but I would keep these things in separate updates.

floooh:
Ok closing this PR now. HDR support is now implemented for Metal and WebGPU via https://github.com/floooh/sokol/pull/1520.

There is no support yet for reconfiguring the swapchain at runtime, but this will happen in a followup update: https://github.com/floooh/sokol/issues/1542

squk:
Awesome!

PR #1453, opened 2026-03-02
Pause the display link rather than invalidating it.

### LINKED ISSUE #1448: Instant crash on macOS when running from tmux and crash on close
Running any of the examples from inside a tmux session on macOS causes the app to crash instantly (log attached).

Also, when pressing the close button on the window of any of the examples when not running from inside tmux will produce the same sort of crash report.

N.B. I’ve not run the C examples directly, only through the Rust and Zig bindings, but the behaviour is the same for both so assume the issue is with the C library.

[crash.log](https://github.com/user-attachments/files/25526517/crash.log)

rizerco:
Seems to fix #1448. Is it okay to pause the display link rather than invalidating and recreating it?

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -5145,7 +5145,7 @@ _SOKOL_PRIVATE id<CAMetalDrawable> _sapp_macos_mtl_swapchain_next(void) {
 }
 
 _SOKOL_PRIVATE bool _sapp_macos_mtl_display_link_active(void) {
-    return _sapp.macos.mtl.display_link != nil;
+    return !_sapp.macos.mtl.display_link.paused;
 }
 
 _SOKOL_PRIVATE void _sapp_macos_mtl_timing_init(void) {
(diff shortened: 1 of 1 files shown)


floooh:
This is actually one thing I wanted try and is also a more lightweight solution than invalidating and recreating the display link each time the window is obscured or minified.

Does this also fix the crash on shutdown you were seeing?

PS: it probably does because the display link is now technically not released on application shutdown, I guess that's fine though since on shutdown everything is destroyed anyway...

floooh:
Don't worry about the failed ci, looks like a (hopefully temporary) Github fuckup.

rizerco:
> Does this also fix the crash on shutdown you were seeing?

Yep, all the previous crashes I was seeing are gone; minimise, occlude, switching spaces, and close :)

floooh:
Ok merged! I did one minor change (shouldn't affect the behaviour though, just a bit of 'defensive coding'):

https://github.com/floooh/sokol/commit/941a1923508ab5a643e857c9a51e84db0ae84ee3

...also added a changelog item.

Thanks for the PR!

rizerco:
> Thanks for the PR!

You’re welcome! It’s a pleasure to contributing to sokol :)

PR #1454, opened 2026-03-02
center window on windows to match macOS and linux

nickav:
If you like, I'm also happy to implement centering as an option instead. As in something like this:

```c
    sapp_run(&(sapp_desc){
        .init_cb = init,
        .frame_cb = frame,
        .cleanup_cb = cleanup,
        .event_cb = event,
        .width = 1280,
        .height = 720,
        .high_dpi = true,
        .center = true,
        .sample_count = 4,
        .logger.func = slog_func,
    });

```

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -9863,11 +9863,9 @@ _SOKOL_PRIVATE void _sapp_win32_create_window(void) {
         int monitor_width = info.rcMonitor.right - info.rcMonitor.left;
         int monitor_height = info.rcMonitor.bottom - info.rcMonitor.top;
 
-        int width = _sapp.framebuffer_width;
-        int height = _sapp.framebuffer_height;
-        int center_x = (monitor_width - width) / 2;
-        int center_y = (monitor_height - height) / 2;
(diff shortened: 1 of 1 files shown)


floooh:
Thanks for the PR, but I'd like defer changes to the window placement behaviour until I get around to add an overhaul of that entire area via explicit placement/sizing functions (using GLFW as 'inspiration').

I also seem to remember that I played around with similar code like on macOS and Linux for the default behaviour on Windows but had unintended side effect in multi-monitor setups when the multimonitor config was setup in a certain way (e.g. stretching the desktop over both monitors) - like the window being centered and sized as if both monitors would be a single monitor, don't quite remember though if this was on Windows or only on Linux, and I might have come to the conclusion that `CW_USEDEFAULT` is the lesser evil. 

I also seem to remember though that `CW_USEDEFAULT` did a pretty good job at least on Win10 (IIRC I got a centered window which filled most of the screen - and on Linux and macOS I tried to replicate this behaviour with explicit placement. I'll need to check if this behaviour has changed on Win11.

For now I'd recommend that you keep your change in a fork.

nickav:
On my Windows 11 machine without this change the window appears in the top left corner of the screen. But no worries, thanks for taking a look!

Issue #1224, opened 2025-03-16
IFRAME interrupts kb/mouse inputs


mikesmullin:
Hi Andre,

Thanks again for creating Sokol! ❤

FYI: The game is not receiving inputs when its wrapped in a Browser IFRAME.

Compare:
- with IFRAME: [example 1](https://games.mikesmullin.com/meow.html) (not working)
- without IFRAME: [example 2](https://cdn.smullindesign.com/meow/build/main.html) (working)

My experience (in IFRAME):
- in **Chrome**: Mouse works, but keyboard does not. (Tested on two computers)
- in **FIrefox Developer Edition**: Mouse works, but keyboard does not.

My current workaround is just not to use IFRAMEs.

I may look into it myself later on, as well. 🤞

floooh:
One difference between mouse and keyboard input is that the mouse event handler is attached to the canvas object:

https://github.com/floooh/sokol/blob/46d3c53d2485ce2eee67287931ff4a5c1d4c3563/sokol_app.h#L5923-L5928

...while keyboard input is attached to the window object (I don't quite remember the reason *why* I'm getting the keyboard input from the window object though, maybe `EMSCRIPTEN_EVENT_TARGET_DOCUMENT` would make more sense?):

https://github.com/floooh/sokol/blob/46d3c53d2485ce2eee67287931ff4a5c1d4c3563/sokol_app.h#L5929-L5931

Also (while I don't think that this is the issue) you could try playing with event bubbling options in sapp_desc:

https://github.com/floooh/sokol/blob/46d3c53d2485ce2eee67287931ff4a5c1d4c3563/sokol_app.h#L1814-L1818

floooh:
PS: just linking this issue which describes a similar problem without having read through all of it so far:

https://github.com/emscripten-core/emscripten/issues/5796

mikesmullin:
`.html5_bubble_mouse_events = true` fixed it, Thanks! 😻 
I tried various combinations of the others, but this is the only one that seemed to matter.

mikesmullin:
I also found this type of approach useful to prevent `Tab` key from escaping the pointer capture.
```js
  <script>
    // Intercept Tab keydown event
    document.addEventListener('keydown', (event) => {
      if (event.key === 'Tab') {
        // Prevent default behavior (e.g., focus next element in the tab order)
        event.preventDefault();

        // Prevent the event from bubbling up to the browser (e.g., releasing pointer capture)
        event.stopPropagation();
      }
[...]

PR #1456, opened 2026-03-06
sapp: emsc: fix permanent loss of focus in iframe

nkorth:
As described in issue #1224, a sokol game running in an iframe (eg. on itch.io) will not receive keyboard events unless the iframe is focused. Since by default mouse events do not bubble up to the page, it's not possible to re-focus the iframe when the canvas takes up the entire iframe. Enabling html5_bubble_mouse_events does fix this, but I think it makes more sense to maintain the window focus behavior even when you don't want all mouse events bubbled up.

I verified that this patch fixes keyboard events on itch.io, but my project is way back on version 85ec51b13f72d211be3fcf5604623a539041987a so it wasn't an ideal test. [Before patch](https://nkorth.itch.io/project-sable-20260305?secret=dMiNe1YjYJ5kJcYN9il5kDfnDNk), [after patch](https://nkorth.itch.io/project-sable-20260306?secret=kbihjpeg71pPMJwVrHWNUnaoXs). I looked briefly at commit history and couldn't find any changes to this code path since then.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -7208,6 +7208,14 @@ EM_JS(void, sapp_js_init, (const char* c_str_target_selector, const char* c_str_
     }
 })
 
+EM_JS(void, sapp_js_focus, (void), {
+    if (window) {
+        if (window.focus) {
+            window.focus();
+        }
+    }
+})
+
 _SOKOL_PRIVATE EM_BOOL _sapp_emsc_pointerlockchange_cb(int emsc_type, const EmscriptenPointerlockChangeEvent* emsc_event, void* user_data) {
     _SOKOL_UNUSED(emsc_type);
     _SOKOL_UNUSED(user_data);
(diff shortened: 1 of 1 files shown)


floooh:
Hmm, not a fan tbh, because the `html5_bubble_mouse_events` (and the other bubble flags) were exactly created for  situations like this. Do you have an example situation where `html5_bubble_mouse_events` has undesired side effects?

If you need a more 'focused' bubbling behaviour (e.g. only bubble mouse clicks but not other mouse events) the recommended way is to set `html5_bubble_mouse_events` to true, but then call `sapp_consume_event()` in your event handler for all mouse events that are not clicks.

If you still need to explicitly call `window.focus()` I would recommend doing that in your own code in a small emscripten specific code block:

```c
#if defined(__EMSCRIPTEN__)
#include <emscripten/em_js.h>
EM_JS(void, web_focus_window, (void), {
    window.focus();
})
#else
void web_focus_window(void) { }
#endif
```

At a later point, sokol_app.h might get more window control functions (e.g. something like `sapp_focus()`), but no ETA.

PR #1477, opened 2026-04-12
sokol_app.h: fix empty struct warning on Emscripten

atimics:
The `_sapp_timestamp_t` struct has no members in the `_SAPP_EMSCRIPTEN` branch, which triggers `-Wgnu-empty-struct` when building with `-Wpedantic`.

Fix: add a dummy `int` member.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2565,7 +2565,7 @@ typedef struct {
             uint64_t start;
         } mach;
     #elif defined(_SAPP_EMSCRIPTEN)
-        // empty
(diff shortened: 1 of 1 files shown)


floooh:
I'm surprised that -Wpedantic triggers a lot more things tbh. Technically the sokol-headers are written in the common C/C++ subset, e.g. not "pedantic" C.

In any case, good fix, thanks!

atimics:
got my build warnings down to zero anyway, thanks for accepting :)

PR #1478, opened 2026-04-12
sokol_app.h: suppress dollar-in-identifier warning in EM_JS cursor code

atimics:
The `sapp_js_make_custom_mouse_cursor` EM_JS block uses JS template literals with `${url}` which triggers `-Wdollar-in-identifier-extension` on Clang with `-Wpedantic`.

Fix: wrap the EM_JS block in `#pragma GCC diagnostic push/pop`.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -7205,6 +7205,8 @@ _SOKOL_PRIVATE void _sapp_emsc_update_cursor(sapp_mouse_cursor cursor, bool show
     sapp_js_set_cursor((int)cursor, shown ? 1 : 0, custom_cursor ? 1 : 0);
 }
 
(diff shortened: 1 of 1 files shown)


floooh:
~~Is this warning part of a broader warning set or does it need to be enabled individually? (if yes I'm tending towards not merging tbh)~~ nvm, -Wpedantic...

I'm not sure if it's worth opening up that particular can of worms tbh (`-Wpedantic`), since as I wrote in the other MR, the sokol headers are not written in standard C, so it's very likely that pedantic warnings will be triggered in new code anyway.

floooh:
Still merged though, thanks!

Issue #1484, opened 2026-04-19
sokol_app_imgui.h to language bindings and sokol-samples


floooh:
- [x] add sokol_app_imgui.h to ~~language bindings~~ sokol-zig
- [x] add sokol-app menu item to sokol-zig-imgui-sample
- [x] add sokol-app menu item to sokol-samples which directly render the debug ui

floooh:
Fixed.

PR #1486, opened 2026-04-19
gen_d: add `sokol_app_imgui`

kassane:
### References
- https://github.com/floooh/sokol/issues/1484
- https://github.com/floooh/sokol/pull/1485


floooh:
Nice, thanks! I'll just go ahead and merge the two PRs...

I also have an idea how to make adding new headers a bit less awkward in the future (for instance avoiding that chicken egg problem): https://github.com/floooh/sokol/issues/1489

PR #1498, opened 2026-04-27
sokol_gfx.h: fix Mac Catalyst OpenGL include headers

purpln:
`sokol_gfx.h` can't be compiled for Mac Catalyst because it is `TARGET_OS_IPHONE` environment but has no OpenGLES and it's headers.

Also added `GLES_SILENCE_DEPRECATION` to silence warnings when building for iOS.

Checked:
- macOS (OpenGL)
- macOS Mac Catalyst (OpenGL)
- iOS (OpenGLES)


floooh:
> because it is TARGET_OS_IPHONE environment but has no OpenGLES and it's headers.

...wow what a mess :D 

Just waiting for the CI to finish and then I'll merge. Thanks!

Issue #1502, opened 2026-05-02
Android timestamps currently jitter


learnopengles:
Android doesn't currently use Choreographer, so timestamps between frames have some jitter to them. For example:

Emulator:
```
frame=1 dt=5.149ms fps=194.2 raw=5.149ms raw_fps=194.2
frame=2 dt=5.082ms fps=196.8 raw=2.452ms raw_fps=407.8
frame=3 dt=24.289ms fps=41.2 raw=24.289ms raw_fps=41.2
frame=4 dt=15.449ms fps=64.7 raw=15.449ms raw_fps=64.7
frame=5 dt=15.536ms fps=64.4 raw=18.939ms raw_fps=52.8

...

frame=60 dt=16.849ms fps=59.4 raw=17.417ms raw_fps=57.4 
```

Samsung Galaxy Z Flip 6:
```
frame=5 dt=14.870ms fps=67.2 raw=14.870ms raw_fps=67.2
frame=6 dt=9.397ms fps=106.4 raw=9.397ms raw_fps=106.4
frame=7 dt=9.348ms fps=107.0 raw=7.437ms raw_fps=134.5
frame=8 dt=9.317ms fps=107.3 raw=8.102ms raw_fps=123.4
frame=9 dt=9.297ms fps=107.6 raw=8.514ms raw_fps=117.5

...

frame=60 dt=8.578ms fps=116.6 raw=8.546ms raw_fps=117.0
```

Even the smoothed timestamps still have some jitter.

With Choreographer (similar to CADisplayLink on iOS), we can get frame callbacks with precise timestamps:

```
frame=5 dt=8.323ms fps=120.1 raw=8.324ms raw_fps=120.1
frame=6 dt=8.323ms fps=120.1 raw=8.324ms raw_fps=120.1
frame=7 dt=8.323ms fps=120.1 raw=8.324ms raw_fps=120.1
frame=8 dt=8.323ms fps=120.1 raw=8.323ms raw_fps=120.1
frame=9 dt=8.323ms fps=120.1 raw=8.324ms raw_fps=120.1

...

frame=60 dt=8.324ms fps=120.1 raw=8.325ms raw_fps=120.1
```

Where frame count Is sapp_frame_count(), dt is sapp_frame_duration(), and raw is sapp_frame_duration_unfiltered().

floooh:
fixed via #1503

PR #1503, opened 2026-05-02
Android choreographer

learnopengles:
This improves frame timing by adding support for Android Choreographer.

If the Choreographer is unavailable or unsupported due to minsdk, it will fall back on the existing implementation which calls _sapp_android_frame in a loop. This is similar to what CADisplayLink does for iOS.

It was a big file to look over, please let me know how the PR looks! BTW, I've been following your work for a while and I appreciate your minimalist approach to C dev. I hope that this PR is helpful. 

Related issue: https://github.com/floooh/sokol/issues/1502

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -1800,6 +1800,8 @@ typedef struct sapp_allocator {
     _SAPP_LOGITEM_XMACRO(ANDROID_NATIVE_ACTIVITY_ONCREATE, "NativeActivity onCreate") \
     _SAPP_LOGITEM_XMACRO(ANDROID_CREATE_THREAD_PIPE_FAILED, "failed to create thread pipe") \
     _SAPP_LOGITEM_XMACRO(ANDROID_NATIVE_ACTIVITY_CREATE_SUCCESS, "NativeActivity successfully created") \
+    _SAPP_LOGITEM_XMACRO(ANDROID_CHOREOGRAPHER_ENABLED, "Choreographer frame loop enabled") \
+    _SAPP_LOGITEM_XMACRO(ANDROID_CHOREOGRAPHER_UNAVAILABLE, "Choreographer unavailable, using poll loop") \
     _SAPP_LOGITEM_XMACRO(WGPU_DEVICE_LOST, "wgpu: device lost") \
     _SAPP_LOGITEM_XMACRO(WGPU_DEVICE_LOG, "wgpu: device log") \
     _SAPP_LOGITEM_XMACRO(WGPU_DEVICE_UNCAPTURED_ERROR, "wgpu: uncaptured error") \
@@ -2515,6 +2517,9 @@ inline void sapp_run(const sapp_desc& desc) { return sapp_run(&desc); }
     #include <time.h>
     #include <android/native_activity.h>
     #include <android/looper.h>
+    #if __ANDROID_API__ >= 29
+        #include <android/choreographer.h>
+    #endif
     #include <EGL/egl.h>
     #include <GLES3/gl3.h>
 #elif defined(_SAPP_LINUX)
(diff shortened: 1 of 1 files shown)


floooh:
Thanks! The change looks simple enough to be included I think.

I've become a bit careful with alternative frame timing methods (I recently removed the DXGI presentation timestamp method for instance, because it had too many edge cases to take care of, while on macOS and iOS, CADisplayLink is about the right amount of "bang for the buck" - in general I think that the new EMA-filter is good enough to not really depend on platform specific frame timing methods).

I'll try to give the PR a whirl tomorrow.

floooh:
Starting to look into this now...

sokol_app.h
```diff
@@ -10660,9 +10696,22 @@ _SOKOL_PRIVATE void* _sapp_android_loop(void* arg) {
...
     /* main loop */
     while (!_sapp.android.is_thread_stopping) {
-        /* sokol frame */
+        #if __ANDROID_API__ >= 29
+            if (_sapp.android.choreographer != NULL) {
+                /* When we have a choreographer, we get frame callbacks via _sapp_android_frame_callback. */
+                if (!_sapp.android.frame_callback_in_flight && _sapp_android_should_update()) {
+                    AChoreographer_postFrameCallback64(_sapp.android.choreographer, _sapp_android_frame_callback, NULL);
+                    _sapp.android.frame_callback_in_flight = true;
+                }
```
floooh: Is this `AChoreographer_postFrameCallback64` needed here since the next frame is already chained via `AChoreographer_postFrameCallback64` up in the actual frame callback?

sokol_app.h
```diff
@@ -10660,9 +10696,22 @@ _SOKOL_PRIVATE void* _sapp_android_loop(void* arg) {
...
+                /* Process events. We don't need a while loop here because we're already being driven by
+                   the outer while loop. */
+                ALooper_pollOnce(-1, NULL, NULL, NULL);
```
floooh: This is blocking until the next event right? Would be nice to mention that in the comment, since at first I was thinking this loop would run unthrottled now.

floooh (commented):
Added a couple of comments, please respond :)

I also wonder if we actually need to care about the pre-android-api 29 fallback?

PS: you can ignore the below comments (became clear during when stepping through the code), but can you add a bit more info to the comments (btw you can use `//` comment style for new comments)

E.g.:

- in the event while-loop, why the extra postFrameCallback is needed (as far as I understand it: to initially "ignite" the frame callback, and to "restart" the frame callback after the app was inactive in the background.
- in the ALooper_pollOnce call, add a note that this is blocking.

floooh:
Despite the review comments I'll continue with some testing, since the comments are not really critical. Will also need to check if I can get AStudio debugging to work again :/ Somehow this stopped working with my Google Pixel 4a a while ago.

floooh:
Ok, got device debugging working again whoop whoop.

sokol_app.h
floooh: Ok, after stepping through the code, I guess this is needed to "restart" the frame callback in situations like the app coming back from background-idling (e.g. situations where the `AChoreographer_postFrameCallback64` in the frame callback is skipped.

learnopengles:
Hi @floooh!

> I also wonder if we actually need to care about the pre-android-api 29 fallback?

I wasn't sure. If you don't need the support then certainly you could remove the fallback. I'll update the PR per your comments.

For debugging, I did it the old-school way with logs. 😇 I often have trouble with the debugger as well.

floooh:
Ok, let's keep pre-Android-29 support in for now.

Let me know when the PR is ready for merge from your side :)

learnopengles:
@floooh Please check now.

floooh:
Looking good. Going to merge...

I'll also add a changelog blurb after the merge.


Thanks for the PR!

floooh:
...and merged.

PR #1527, opened 2026-05-29
Expose Android's native window ANativeWindow (fixes #1526)

### LINKED ISSUE #1526: Expose Android's ANativeWindow
Hi, I finally have a case to use sokol, and it's awesome! 

I need Android's native window (ANativeWindow) for some stuff, like settings frame rate: https://developer.android.com/ndk/reference/group/a-native-window#anativewindow_setframerate

There are technically ways to fetch it from JNI through some very convoluted cursed Activity subclassing, but since sokol already has it internally, I figured it would be more convenient to expose it. I did this in my private fork, and seems like it works well.

There is an old PR that does this: https://github.com/floooh/sokol/pull/938, but it also adds some other, unrelated stuff. I only want the window :)

The implementation would be just this:

```c
SOKOL_API_IMPL const void* sapp_android_get_native_window(void) {
    #if defined(_SAPP_ANDROID)
        return (void*)_sapp.android.current.window;
    #else
        return 0;
    #endif
}
```
Or we can add some validation/asserts in it to make sure it doesn't get called before activity is set (which happens in sokol_main).

What do you think? Would you be opposed to the idea? If you're fine with it, I can make a PR. Or, if you think there could be a better way to do this that isn't straight up hijacking sokol's `_sapp_android_on_native_window_created` and everything it does downstream - we can do that as well.

Thanks a lot!

valignatev:
As discussed in #1526, here's the PR that exposes Android's window. Thanks!

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -377,8 +377,9 @@
             whether the GL context is a GLES context
 
         const void* sapp_android_get_native_activity(void);
-            On Android, get the native activity ANativeActivity pointer, otherwise
-            a null pointer.
+        const void* sapp_android_get_native_window(void);
+            On Android, get the native activity ANativeActivity pointer,
+            or native window ANativeWindow pointer, otherwise a null pointer.
 
     --- Implement the frame-callback function, this function will be called
         on the same thread as the init callback, but might be on a different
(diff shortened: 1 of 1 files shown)


floooh:
That was quick :) I'll let the CI finish and then will merge. Thanks!

floooh:
And merged.

Issue #1529, opened 2026-06-02
sapp_frame_time() filtering goes crazy when the framerate is low


Seb-degraff:
I just upgraded sokol_app.h for my projects, and noticed that scrolling had a slightly jerky feeling. To investigate, I plotted the value of sapp_frame_duration(), and the new sapp_frame_duration_unfiltered(). It turns out that sapp_frame_duration() is quite smooth normally, but when the framerate drops (here because a somewhat poorly optimized editor is opened), it starts producing oscillating values.

In this video, the cyan blue line is the filtered frame duration and the green one is the unfiltered frame duration.

https://github.com/user-attachments/assets/8c58279c-8b90-4ed3-9c6f-2783d18226ee

I'm seeing this on macos 14 (sonoma). I'll check how the filter behaves on other platforms soon. I'll also double check that I'm not doing anything weird on my side. The fact that the unfiltered values are smooth make me think this is likely a bug in sokol_app though.

floooh:
Woah, definitely unexpected, seems like the new EMA filter is wildly overshooting and doesn't recover on its own (the expected behaviour would a smooth curve which goes back to an average).

Maybe the smoothing parameters are too 'tight'.

I won't be able to look into the problem for the rest of the week. If you need a quick fix it's probably best to use the unfiltered value via `sapp_frame_duration_unfiltered()`.

The other options is to play around with the EMA filter `alpha` parameters here (try changing the value in both direction and check in which direction disappears, e.g. try changing the 0.025 to 0.25 and 0.0025)

https://github.com/floooh/sokol/blob/84a23f40d1cacdec979e7a37df38a8998beb15d2/sokol_app.h#L2653-L2662

NOTE: this advice only makes sense when `sapp_frame_duration()` actually returns the EMA filtered value, e.g. only on the GL backend, but not the Metal backend. The Metal backend directly uses timestamps from CADisplatLink for `sapp_frame_duration()` without going through any filtering, see comment below.

floooh:
...wait a sec though...

On macOS (with Metal) it's a bit more complicated: the EMA filter won't even be used when the display link is active (e.g. when the window is shown)...

~~E.g. even though the frame time smoothing goes wild (which is definitely a bug) on macOS with Metal backend the filtered value isn't actually returned in `sapp_frame_duration()`~~ (not sure if this actually happens since the smoothed value isn't really accessible), instead the returned value is computed directly from a timestamp provided by CADisplayLink without additional smoothing.

[...]

Issue #1344, opened 2025-10-08
Sporadic input delay on MacOS Tahoe 26.0.1 (1000Hz mouse)


terrybrash:
After updating from Sequoia to Tahoe, input events are sometimes throttled and delayed for a period of time. This is really obvious with mouse input:

https://github.com/user-attachments/assets/3b7306d0-838f-40b7-8a67-1791f76a690f

Every line in the console output contains a "=" per event. My mouse is a Logitech G305 set to 1000hz polling rate which is why I'm getting so many mouse events per frame initially.

This is the expected behavior:

https://github.com/user-attachments/assets/98da4187-bdbf-41f8-82e0-b28d12992ade

I didn't think to create a repro test case before downgrading back to Sequoia but it should be as simple as plugging in a high rate mouse and moving it around for a while. The throttling happens pretty frequently.

floooh:
Ouch, tbh I don't know what to do about that, the mouse events are directly forwarded from the Cocoa event callback without any complex computations (nothing that would explain the slowdown when running 1000x per seconds):

https://github.com/floooh/sokol/blob/a99d4f344bf314a9ec3968fd22307d003579c745/sokol_app.h#L5081-L5089

https://github.com/floooh/sokol/blob/a99d4f344bf314a9ec3968fd22307d003579c745/sokol_app.h#L4467-L4474

floooh:
...wait where is that `_sapp_gl_make_current()` coming from...

Are you on the GL backend?

I'll need to do some git archeology, because making the GL context current in each event definitely doesn't look right...

PS: ok... I remember where that 'make GL context current' thingie is coming from. The problem was that outside the NSOpenGLView frame callback the context isn't active and any attempts to call sokol-gfx functions from inside the sokol-app event callback failed...

But, question remains: is this the GL or Metal backend?

terrybrash:
> But, question remains: is this the GL or Metal backend?

This is the Metal backend.

It's not obvious from the video but I don't think the frame time was impacted, just rate at which input events were being sent to the application. Every frame I should be receiving ~8 frames of mouse move events but on Tahoe sometimes it dips down to 1 mouse move event per frame.

floooh:
> This is the Metal backend.

Ok, then I'm out of ideas unfortunately... I'll see if I can reproduce the issue... probably need to get a high-frequency mouse first (I have a 'gaming mouse' on my PC but not sure how many ticks-per-second this does)...

[...]

PR #1536, opened 2026-06-08
macos+metal: remove displaylink timing

floooh:
See #1529 and #1344.

Entirely changes the macos+metal frame pacing (again) by removing CADisplayLink and instead relying on CAMetalLayer alone for vsync throttling. Instead of letting CADisplayLink injecting a selector into the event loop at vsync time (which causes all sorts of weird problems when the application cannot meet the frame budget), sokol_app.h immediately injects the frame callback selector into the runloop at the end of a frame (pretty much like requestAnimationFrame works in web browsers). This will start the next frame at the next possible moment, and the waiting for vsync will always happen in the CAMetalLayer nextDrawable call.

There is still a suprisingly high CPU cost when moving the mouse cursor, but this has a less catastrophic effect when the app cannot meet the frame budget as it has with CADisplayLink (and ultimately this high CPU cost for moving the mouse must be fixed by macOS).

TODO:

- [x] NOTE: draining events breaks event handling in macOS 27 Beta!
- [x] remove CADisplayLink
- [x] rename fallback timer to occluded timer
- [x] trigger next frame in non-occlued state via `performSelector`
- [ ] Q: should we drain the event queue at the start of a frame? there's hardly any events to drain at that point (at most one in very long frames)
- [x] don't call nextDrawable in occluded state, instead return an 'invalid' swapchain-struct in `sapp_get_swapchain()`
- [ ] merge into WIP swapchain-config branch
- [ ] extensively test in debug vs release vs app-bundle vs 'raw' exe

No changes on iOS or WGPU (those will continue to use CADisplayLink for now)

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -5128,6 +5128,7 @@ _SOKOL_PRIVATE void _sapp_macos_mtl_request_next_frame(void) {
                       withObject:nil
                       waitUntilDone:NO
                       modes:@[NSDefaultRunLoopMode, NSEventTrackingRunLoopMode, NSModalPanelRunLoopMode]];
+    CFRunLoopWakeUp(CFRunLoopGetMain());
 }
 
 _SOKOL_PRIVATE bool _sapp_macos_mtl_occluded(void) {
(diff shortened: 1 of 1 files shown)


floooh:
Abandoning this PR for now because it doesn't consistently improve things (e.g. the unfiltered frame rate is actually more jittery than with CADisplayLink).

PR #1537, opened 2026-06-09
Updating gen_c3.py to define enums the C3 0.8.x way.

NilOmniscient:
From: https://github.com/floooh/sokol-c3/pull/12#issuecomment-4651670168

Fixing the enum generation so the bindings work with C3 0.8.x

### DIFF
--- a/bindgen/gen_c3.py
+++ b/bindgen/gen_c3.py
@@ -325,7 +325,7 @@ def gen_enum(decl, prefix):
     tpe = "int"
     if any(as_enum_item_name(check_override(item['name'])) == 'FORCE_U32' for item in decl['items']):
         tpe = "uint"
-    l(f'enum {as_struct_or_enum_type(enum_name, prefix)} : const {tpe}')
+    l(f'constdef {as_struct_or_enum_type(enum_name, prefix)} : {tpe}')
     l('{')
     value = "-1"
     for item in decl['items']:


floooh:
Thanks! I'll test and merge later today (late afternoon/early evening).

We should also update the C3 version used in CI:

https://github.com/floooh/sokol/blob/0eab0e92731f997312b139292b327b34e1c9e5a9/.github/workflows/gen_bindings.yml#L383

...but if you don't get around to that fix until the merge I can simply do that during merging (but looks like the new code is already accepted by 0.7.x).

floooh:
Ok, looking into this now.

floooh:
Ok merged. Looking into the CI problem now...

floooh:
The directory structure of the C3 release zips has changed causing the radekm/setup-c3 action to fail. I'll see if I can fix that in a fork, temporarily use that fork in the CIs and create a PR for @radekm (if I don't manage I'll just remove the macos and windows builds from the buid matrix for now)

floooh:
Ok, my setup-c3 fork seems to work:

https://github.com/floooh/sokol-c3/actions/runs/27225773993

I'll create a PR for https://github.com/radekm/setup-c3

floooh:
CI jobs updated to 0.8.0 and all working now (with my temporary fork of the setup-c3 action):

- https://github.com/floooh/sokol/actions/runs/27226063055
- https://github.com/floooh/sokol-c3/actions/runs/27225773993

PR #1551, opened 2026-07-14
sapp: Add generic native_event_cb hook to sapp_desc

jasonfrowe:
I am creating an emulator for the Picocomputer 6502 that targets Android with gamepads.  

This is a potential solution that provides a platform-agnostic "escape hatch" allowing applications to intercept raw, OS-level window and input events (e.g.,  AInputEvent*  on Android) before Sokol processes or translates them by adding a generic native event callback ( native_event_cb ) to the  sapp_desc  configuration struct.  If the callback returns true, the event is marked as handled/consumed, and Sokol bypasses its own touch or keyboard translation handlers.

### DIFF
--- a/bindgen/gen_c3.py
+++ b/bindgen/gen_c3.py
@@ -112,6 +112,8 @@ aliases = {
         ("StreamCb", "fn void(float*, CInt, CInt)"),
     "void (*)(float *, int, int, void *)":
         ("StreamDataCb", "fn void(float*, CInt, CInt, void*)"),
+    "bool (*)(const void *)":
+        ("NativeEventCb", "fn bool(void*)"),
 }
 
 struct_types = []
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -2063,6 +2063,7 @@ typedef struct sapp_desc {
     void (*frame_cb)(void);
     void (*cleanup_cb)(void);
     void (*event_cb)(const sapp_event*);
+    bool (*native_event_cb)(const void* native_event);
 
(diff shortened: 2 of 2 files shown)


floooh:
Hmm, turning this into a platform-agnostic thing is too much hassle tbh since event handling across platforms is very different (most don't have a common event struct).

Let's turn it into an Android-speciifc thing:

E.g. `sapp_desc` already has the following platform specific options:

```c
    sapp_gl_desc gl;
    sapp_metal_desc metal;
    sapp_win32_desc win32;
    sapp_html5_desc html5;
    sapp_ios_desc ios;
```

Please add a struct `sapp_android_desc` above `sapp_desc:

```c
typedef struct sapp_android_desc {
    bool (*native_event_cb)(const void* native_event);
} sapp_android_desc;
```
...and put that at the bottom of `sapp_desc`:

```c
typedef struct sapp_desc {
    // ...
    sapp_gl_desc gl;
    sapp_metal_desc metal;
    sapp_win32_desc win32;
    sapp_html5_desc html5;
    sapp_ios_desc ios;
    sapp_android_desc android;
```

...thanks for also fixing that C3 problem! I just wanted to write about that and then saw your other commit.

PS: I'll won't repsond until tomorrow, it's past midnight here :)

floooh:
...and merged. Thanks!

(I also added a changelog blurb)

prime31:
Just wanted to say thanks for this! I just hacked in this exact thing and decided to check the PRs and here it was! Thanks crew!

PR #1567, opened 2026-08-06
Update gen_nim so the code generated works also in nimony (to be nim3)

leiserfg:


leiserfg:
I can remove the first commit if you don't like it, it's mostly for my own use. 

The changes were:
- Replaced `passc` by `passC`, the first one should not be used.
- Fixed the case of the vars in the wrapper as nimony is case sensitive (not like in nim2). 
- Added `{.feature: "lenientconverters".}` to allow the custom converters as nimony does not like converters for base types.
- Replaced the `static: asserts N<n` by `{.requires: N<n.}`, as static was removed in nimony and `requires` can be implemented (for this cases) as a macro.

leiserfg:
I checked that the modules compile, maybe I need to update the examples too, but before doing that I will like you to merge my changes.

leiserfg:
I also noticed that the code generated by the shader codegen has some case missmatches that make the compilation fail (til now the ones of the enum `ShaderAttrBaseType`). I will also like to fix them (that would be on the shader's project). 

Edit:
I ended doing the change here, 'cause the name generated by sdhc looks better to me.

floooh:
I'll do a proper code review later, but please remove the `.envrc` and `shell.nix` files that somehow slipped into the PR.

floooh:
Ignore the emscripten CI error, that's some unrelated new thing.

leiserfg:
> I'll do a proper code review later, but please remove the `.envrc` and `shell.nix` files that somehow slipped into the PR.

They are in a separated commit, I will remove it once you give me the :+1:  for the rest if that's fine for you. I added it just to have clang in that folder cause I don't have it globally.

floooh:
> I checked that the modules compile, maybe I need to update the examples too, but before doing that I will like you to merge my changes.

There may be a chicken egg problem (we'll see once CI has finished). In that case we may need to prepare a 'manual' PR for the sokol-nim project which would contain updated examples *and* the bindings output (e.g. such a manual PR would look like this: https://github.com/floooh/sokol-nim/pull/41)

I think that would only be an issue when the new bindings are not backward compatible.

leiserfg:
> There may be a chicken egg problem (we'll see once CI has finished).

The changes are supposed to be backward compatible cause the ones in casing don't affect nim2 (it does not care about casing) and the other ones are guarded by checking the nim version.

floooh:
Oki, I'll try to do a proper code review later today or latest tomorrow.

For shader code-generation changes we'd need a separate PR in https://github.com/floooh/sokol-tools, in this source file: https://github.com/floooh/sokol-tools/blob/master/src/shdc/generators/sokolnim.cc

leiserfg:
> Oki, I'll try to do a proper code review later today or latest tomorrow.
> 
> For shader code-generation changes we'd need a separate PR in https://github.com/floooh/sokol-tools, in this source file: https://github.com/floooh/sokol-tools/blob/master/src/shdc/generators/sokolnim.cc

See my last commit, with that is enough so we don't need to change shdc.

leiserfg:
Besides that, I would like to replace the chunks in the code with a gazillion `l` calls by `ll` with multi-lines strings as that's easier to read, but I can do that in another PR (if you like me to do so).

bindgen/gen_nim.py
```diff
@@ -209,8 +212,16 @@ def as_camel_case(s, prefix, wrap=True):
         outp = wrap_keywords(outp)
     return outp
 
+
+enum_item_replacements= {
+    "SHADERATTRBASETYPE": "SHADER_ATTR_BASE_TYPE" # This override is done in order to match the output of shdc
+}
+
```
floooh: I think this can be better solved by adding a new item to the existing `enumPrefixOverrides` map (near the top of the source file).

bindgen/gen_nim.py
```diff
@@ -209,8 +212,16 @@ def as_camel_case(s, prefix, wrap=True):
...
 # PREFIX_ENUM_BLA_BLO => blaBlo
 def as_enum_item_name(s, wrap=True):
+    for og, replacement in enum_item_replacements.items():
+        s = s.replace(og, replacement)
+
```
floooh: ...and when `SHADERATTRBASETYPE` is handled via `enumPrefixOverrides` this special case can be removed.

bindgen/gen_nim.py
```diff
@@ -570,15 +578,29 @@ def gen_extra(inp):
...
+    when not defined(nimony):
+      import std/macros
+      macro requires(condition: untyped, body: untyped): untyped =
+        result = body
+        let assertStmt = quote do:
+          static:
+            doAssert `condition`, "Precondition failed: " + astToStr(`condition`)
+        result.body.insert(0, assertStmt)
+      """)
```
floooh: Please use the regular `l()` function like I'm doing here: https://github.com/floooh/sokol/blob/6c283ecc698d583f1e8b6f85388c6b55b4bbe0cc/bindgen/gen_zig.py#L467-L505, and remove the special `ll()` function.

bindgen/gen_nim.py
```diff
@@ -159,6 +159,9 @@ def l(s):
     global out_lines
     out_lines += s + '\n'
 
+def ll(s):
+    l(dedent(s))
+
```
floooh: Remove this helper function (see other comment below where it is used).

floooh (changes requested):
First review round for things that immediately caught my eye, will do another more careful round later.

floooh:
> Besides that, I would like to replace the chunks in the code with a gazillion `l` calls by `ll` with multi-lines strings as that's easier to read, but I can do that in another PR (if you like me to do so).

I don't agree ;) See my code review comments. I prefer the existing single-line solution with one `l()` call per line.

bindgen/gen_nim.py
leiserfg: Your repo, your rules :smile: 
I just prefer using """, its easier to type.
I will do as you asked.

bindgen/gen_nim.py
leiserfg: Crap sorry didn't notice that something like this was already in place.

floooh (approved):
Ok, the gen_nim.py changes look good now, can you remove the two nix files? Then I can merge.

floooh:
Ok merged. Thanks! 

Once this pipeline goes green, the bindings in the sokol-nim repo will be uptodate:

https://github.com/floooh/sokol/actions/runs/31199934337

PR #1570, opened 2026-08-09
bindgen-nim: Make pointer and cstring types explicitly nil-able

leiserfg:


leiserfg:
This change is also backward compatible, as nil-able is the default in nim2.

floooh:
And merged. Thanks!

PR #1571, opened 2026-08-09
Add missing override 

leiserfg:
I found these missing overrides during the examples rewrite. I will later make another PR to sokol-nim to update the generated shader files and the first working one.


floooh:
Oki doki, I'm wondering though... this only affects the capitalization style and wouldn't break compilation right? Might be a bit too much hassle in the future to keep those manual mapping updated.

leiserfg:
These changes are just so the output matches the code already in the examples. Btw, I'm almost done with those.

Issue #1315, opened 2025-08-17
sokol_gfx.h: maybe make sg_destroy calls with an invalid handle an error.


floooh:
With the view update there's the following 'loophole' when destroying a view and its associated image:

```c
sg_destroy_view(view);
sg_destroy_image(sg_query_view_image(view));
```

The `sg_destroy_image` call will be a silent no-op because `sg_query_view_image()` returns an invalid handle.

Maybe at least log an error when the `sg_destroy_*` funcs are called with an `SG_INVALID_ID` handle. This dials back some of the 'convenience' but might at least avoid a resource leak.

PR #1575, opened 2026-08-10
sokol_gfx.h: add validation errors for invalid handle destruction

nihiL7331:
Handles #1315.

This PR adds explicit validation layer checks to all sg_destroy_* functions to catch and report when a user attempts to destroy a stale or invalid handle. Previously, passing an invalid handle would silently fail and return.

Changes:
- Added 6 new `_SG_LOGITEM_XMACRO` entries for destruction validation.
- Added `_sg_validate_destroy_*` functions to the internal validation section.
- If the validation fails, the public function now returns early before attempting destruction.


floooh:
The "silent failure" is intended behaviour (also in other places, e.g. trying to render with resources bound that are not in valid resource state - although details slightly differ between `SG_INVALID_ID` versus non-valid resource state).

(at least so far... from time to time I'm considering making it an non-fatal error, e.g. see https://github.com/floooh/sokol/issues/1315, but I haven't made up my mind yet... maybe the best course of action is to add an option `destroy_warn_on_invalid_handles` debug-flag to sg_desc, but I haven't decided on that yet.

Did you run into a concrete situation where having the destroy functions fail on being called with an invalid handle would be useful?

Issue #1552, opened 2026-07-15
sg_make_pipeline: the automatic vertex attribute offset computation should be decided per layout buffer.


floooh:
About this code:

https://github.com/floooh/sokol/blob/b02162392486717c2c9b229c37166594d6d7a7ac/sokol_gfx.h#L24313-L24339

Currently *all* vertex attribute offsets must be zero for the auto-offset computation to kick in. This would deactive that offset-computation for the instance-data attributes in this case (since the sshape helper functions return an explicit offset, but that's for the buffer-layout at index 0:

```c
[ATTR_display_instanced_pos] = sshape_position_vertex_attr_state(&shp),
[ATTR_display_instanced_normal] = sshape_normal_vertex_attr_state(&shp),
[ATTR_display_instanced_inst_xxxx] = { .format = SG_VERTEXFORMAT_FLOAT4, .buffer_index = 1 },
[ATTR_display_instanced_inst_yyyy] = { .format = SG_VERTEXFORMAT_FLOAT4, .buffer_index = 1 },
[ATTR_display_instanced_inst_zzzz] = { .format = SG_VERTEXFORMAT_FLOAT4, .buffer_index = 1 },
[ATTR_display_instanced_inst_color] = { .format = SG_VERTEXFORMAT_FLOAT4, .buffer_index = 1 },
```
...this is confusing and should be fixed. E.g. since all offsets for `.buffer_index` are zero, their attribute offsets should be auto-computed.

floooh:
Fixed via https://github.com/floooh/sokol/pull/1576

PR #1576, opened 2026-08-10
sokol_gfx.h: evaluate vertex attr auto-offsets per buffer index

nihiL7331:
Handles #1552.

Changes:
- Replaced the `bool use_auto_offset` with `bool use_auto_offset[SG_MAX_VERTEXBUFFER_BINDSLOTS]`.
- An explicit offset now only disables auto-compute for its specific `buffer_index`.


floooh:
Thanks! This looks like a good change. I'll try to do a proper code review and some testing 'soon-ish' (later today or tomorrow)

floooh:
Ok, merged! Thanks!

I did two minor changes (changing the array declaration to the _SG_STRUCT helper macro, ok confusingly named for arrays), and added a missing SOKOL_ASSERT which range-checks `a_state->buffer_Index`, also added a changelog blurb.

E.g. the merged result ended up like this:

https://github.com/floooh/sokol/commit/3bfdce6c09885dc52e5726d764b7ba8dbedb4b4b

PR #1577, opened 2026-08-11
sokol_framebuffer.h: minor cleanups - remove duplicate macro and fix typo

nihiL7331:
This PR provides cleanup for `sokol_framebuffer.h`:

- Fixes typo in `sfb_resource_state` description (`sfg_query_framebuffer_state`->`sfb_query_framebuffer_state`).
- Removes duplicate `_SFB_INVALID_SLOT_INDEX` macro (kept one closer to `_sfb_slot_t`)
- Adds space in `//>#shdgen` to match sectioning style (`//>#shdgen`->`// >#shdgen`)

### DIFF
--- a/util/sokol_framebuffer.h
+++ b/util/sokol_framebuffer.h
@@ -583,7 +583,7 @@ inline void sfb_render_ex(sfb_framebuffer fb, const sfb_render_desc& desc) { ret
     #define SOKOL_ASSERT(c) assert(c)
 #endif
 
-// >#shdgen
+//>#shdgen
 #if defined(SOKOL_GLCORE)
 static const uint8_t _sfb_rgba8_vs_source_glsl410[407] = {
     0x23,0x76,0x65,0x72,0x73,0x69,0x6f,0x6e,0x20,0x34,0x31,0x30,0x0a,0x0a,0x75,0x6e,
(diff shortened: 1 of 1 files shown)


util/sokol_framebuffer.h
```diff
@@ -583,7 +583,7 @@ inline void sfb_render_ex(sfb_framebuffer fb, const sfb_render_desc& desc) { ret
     #define SOKOL_ASSERT(c) assert(c)
 #endif
 
-//>#shdgen
+// >#shdgen
```
floooh: This change needs to be reverted because it would break the automatic shader generation script which specifically scans for the text patterns `//>#shdgen` and `//<#shdgen`.

(e.g. see: https://github.com/floooh/sokol/blob/1896c59828340fc9b8d95df218bca3c0c47e07d4/shdgen/shdgen.ts#L149-L152)

floooh (changes requested):
Looks good except that one `//>#shdgen` change, it's not for 'sections' but a marker for the automatic shader injection script.

util/sokol_framebuffer.h
nihiL7331: That's on me, sorry for not confirming its exact use case. Will handle that in a minute.

floooh (approved):

floooh:
Ok thanks! That D bindings CI error looks completely unrelated. I'll merge the PR.

Issue #1556, opened 2026-07-25
Feature Request: Add support for Indirect Draw Calls (sg_draw_indirect)


WOLFxxxxxx:
Hi @floooh,

Is there any plan to support Indirect Draw calls (such as sg_draw_indirect) in sokol_gfx.h for modern backends that support it (Vulkan, WebGPU, Metal, D3D11)?

Having drawIndirect support would be extremely beneficial for GPU-driven pipelines, specifically for GPU Frustum Culling and LOD management. It would allow us to render large numbers of instances directly on the GPU without requiring CPU readbacks or creating CPU bottlenecks.

Thanks for maintaining such an amazing library! 🚀

floooh:
Eventually there will be a draw-indirect feature as implemented by WebGPU. It will have to be an optional feature, most likely supported by the same backends that currently support compute shaders. The question is just how useful draw-indirect actually is with the limited number of resource bindpoints (e.g. without also having some sort of bindless-support - which would be much harder to abstract since bindless 'leaks' into shader authoring, e.g. SPIRVCross and Tint would need to support a common bindless model too).

Don't hold your breath though, currently I'm focusing on a resource update API rewrite.

floooh:
Relevant:

https://issues.chromium.org/issues/475719023

PR #1578, opened 2026-08-11
sokol_gfx.h: add indirect draw and dispatch support

squk:
https://github.com/floooh/sokol/issues/1556

I've had a hacky variant of this working in a private branch of my https://github.com/squk/sokol_utils repo that I got working for a virtual geometry renderer PoC for Metal only. Here's a broader attempt.


floooh:
I won't merge the PR, but I'll keep it open in case people are interested before an 'official' solution arrives.

floooh:
(I should probably start thinking about setting up a CONTRIBUTING.md...)

floooh:
PS: having looked through the PR a bit, I will very likely base the 'official' thing directly on it, so one way or another the PR will most likely end up in master but with a couple of things renamed, moved around and added. I just want to push through at least the first half of the 'new resource update API' first. Also for new features I want to take enough time to get deeply familiar with the code (e.g. "as if I wrote it myself").

In any case, thanks for the effort! :)

Issue #1557, opened 2026-07-27
Planning: new sokol util header sokol_cmdbuf.h


floooh:
This would make up for the loss of `sg_append_buffer` (which has lousy performance anyway on some backends) in the new sokol-gfx resource update API:

A header for recording pass commands into a command buffer (regular memory, not a backend 3D API command buffer), and submit the recorded commands inside a sokol pass.

This would allow to 'interleave' appending buffer updates and draw calls (e.g. append a new chunk of data to a buffer and then issue a draw command which renders from this new chunk of data), and generally record draw commands ahead of sokol passes (e.g. "anywhere in a frame").

...but it would also drastically simplify other sokol util headers which implement their own internal command buffers (sokol_debugtext.h, sokol_gl.h, sokol_spine.h, ...?).

API:
```c
void scb_setup(&(scb_desc){ .buffer_pool_size = ... });
void scb_shutdown();
sbc_buffer sbc_make_buffer(&(sbc_buffer_desc){ .size = ... });  // size in bytes
void scb_destroy_buffer(cmd_buf);

// sokol-gfx shims (all functions that can be called in a pass), args are direct sokol-gfx types
void scb_apply_viewport(cmd_buf, x, y, width, height, origin_top_left);
void scb_apply_viewportf(cmd_buf, x, y, width, height, origin_top_left);
void scb_apply_scissor_rect(cmd_buf, x, y, width, height,origin_top_left);
void scb_apply_scissor_rectf(cmd_buf, x, y, width, height,origin_top_left);
void scb_apply_pipeline(cmd_buf, pip);
void scb_apply_bindings(cmd_buf, bindings);
void scb_apply_uniforms(cmd_buf, ub_slot, data);
void scb_draw(cmd_buf, base_element, inum_elements, num_instances);
void scb_draw_ex(cmd_buf, base_element, num_elements, num_instances, base_vertex, base_instance);
void scb_dispatch(cmd_buf, num_groups_x, num_groups_y, num_groups_z);

// submit recorded commands to sokol-gfx and rewind command buffer, call within a sokol-gfx pass
void sbc_submit(cmd_buf);
```
...and that's it :)

- Q: should rewind be optionally explicit? That way command buffers could be recorded which could be submitted multiple times, could be a creation-flag, and an additional function `sbc_rewind(cmd_buf)`

[...]

PR #1580, opened 2026-08-11
add sokol_cmdbuf.h: command buffers for deferred execution

nihiL7331:
Handles #1557.

This PR introduces sokol_cmdbuf.h, a new utility header for recording sokol-gfx pass commands into memory and executing them deferred.

The logic heavily resembles sokol_framebuffer.h, so it shares its pool logic, documentation and code styling.

A couple of architectural questions i stumbled upon while working on it:

1. Should bindings/uniforms be copied in their entirety to the buffer? The benefit of copying them is that the user doesn't need to keep the structs valid until execution, on the other hand the memory footprint for these two calls is visibly larger. If we just store the ID, it might be counterintuitive and hard to track down for users when issues occur. My implementation copies the full struct with its data.

2. How should buffer overflow be handled? I decided to warn on overflow but let it execute in the frame, similarly to how rendering buffers are handled.

Regarding questions from the issue:

Q: should rewind be optionally explicit? That way command buffers could be recorded which could be submitted multiple times, could be a creation-flag, and an additional function sbc_rewind(cmd_buf)

I suppose this is approach allows for great flexibility without becoming a big hurdle for normal per-frame generation usage. I decided to not implicitly clear the buffer on scb_submit and go with this option.

Q: how far to go with special support for multihtreading? Command buffers could be populated in threads, but resource update calls would still need to happen on the main-thread... extend with resource update calls (and separate 'data buffers' in a followup update? Only makes sense once the new resource update API is finished.

As I wanted to just provide a baseline for this issue, I didn't touch multithreading. I personally see this as a long-term update to this header.

Posting this PR as a WIP. It still needs docs polish, also more info should be logged.

### DIFF
--- a/util/sokol_cmdbuf.h
+++ b/util/sokol_cmdbuf.h
@@ -1,0 +1,1034 @@
+#if defined(SOKOL_IMPL) && !defined(SOKOL_CMDBUF_IMPL)
+#define SOKOL_CMDBUF_IMPL
+#endif
+#ifndef SOKOL_CMDBUF_INCLUDED
+/*
+    sokol_cmdbuf.h -- command buffers for deferred execution
+
+    Project URL: https://github.com/floooh/sokol
+
+    Optionally provide the following defines with your own implementations:
+
+    SOKOL_ASSERT(c)       - your own assert macro (default: assert(c))
+    SOKOL_CMDBUF_API_DECL - public function declaration prefix (default: extern)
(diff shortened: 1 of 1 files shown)


floooh:
Aaah stop it ;)

I'll do that myself.

floooh:
E.g. TL:DR: smaller fixes like the PRs I accepted so far are welcome, but don't bother with anything that's more than a handful lines of code, I'll not merge those :)

PR #1590, opened 2026-08-28
EM_JS UTF8ToString WASM64 compat: BigInt to number

schaten:
Hi,
thanks for making sokol!

We had some small difficulties using sokol with 64 bit webassembly,
these changes solved the problem for us.

Best
Philip

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -7345,7 +7345,7 @@ EM_JS(void, sapp_js_remove_clipboard_listener, (void), {
 })
 
 EM_JS(void, sapp_js_write_clipboard, (const char* c_str), {
-    const str = UTF8ToString(c_str);
+    const str = UTF8ToString(Number(c_str));
     const ta = document.createElement('textarea');
     ta.setAttribute('autocomplete', 'off');
     ta.setAttribute('autocorrect', 'off');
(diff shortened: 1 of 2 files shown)


floooh:
Hmm, while I'm sure it works, the fix looks a bit 'hacky'. I'll need to look and ask around a bit to figure out if this is the recommended approach or a bug/oversight in Emscripten.

schaten:
There is this an open issue at emscripten-core: https://github.com/emscripten-core/emscripten/issues/16975 not sure if you have seen it - I mentioned it only in the commit not in the comment of the pull request - sorry for that.

Still, maybe if there is a better way to work around this, that would be nice. :)

floooh:
...ignore the CI error, that's just Github acting up again :/

floooh:
Ah thanks for the Emscripten issue link that clears things up...

...I think we can just generally do `UTF8toString(Number(ptr))` also on wasm32 and get rid of the ifdefs? I don't think that adds any noticeable overhead...

Can you try this change in the PR?

After that I can take over and fix the few other places (e.g. sokol_fetch.h also has 2 calls)... and I probably also need to fix a couple of other places where WASM heap pointers are passed to the JS side that are not strings.

floooh:
Just out of interest, what do you want to do with a web app that needs more than 4 GB of memory? :)

Because memory load/store operations on wasm64 are actually slower than on wasm32, so the only reason to switch is heap size, but not performance (e.g. see https://spidermonkey.dev/blog/2025/01/15/is-memory64-actually-worth-using.html).

schaten:
I have updated the pull request!
As to why 64 bit: Magnetic Resonance Imaging (MRI) reconstructions in the browser with [BART](https://codeberg.org/mrirecon/bart/). :)

Thanks for the heads up I wasn't aware of the performance impact..

floooh:
Ok, I added the same fix to sokol_fetch.h, also added a comment above each call, but didn't fix other places where pointers are passed into EM_JS functions (let's just fix those when needed... I suspect that the Emscripten SDK will also come up with a solution, so any fix we do now might then be redundant or obsolete anyway).

Also did a couple of quick tests in the sokol-samples, but only in wasm32 mode (just to make sure nothing broke).

Thanks for the PR!

(e.g. this is what ended up in master: https://github.com/floooh/sokol/commit/6b70854a8030f48105d410652127d359da8e0f5f)

PR #1600, opened 2026-09-09
support optional passthrough of tvOS menu button press events.

tomasandrle:
this allowed me to comply with Apple's rule for tvOS: the remote's menu button should always navigate one level up inside the app, until it reaches the app's navigation root. then it should suspend the app and the user should get to the tv's home screen.

when I process sokol events, I check the game's navigation: during gameplay, the menu button will pause the game. in the level selection screen, it will go back to the main menu. then on the main menu I will skip the `sapp_consume_event()` call, thus passing it through to the OS, which will suspend the app.

tested on physical device with tvOS 26.


floooh:
Merged. Thanks!

PR #1601, opened 2026-09-10
(not so great way to..) partially support Apple TV HD (4th gen)

tomasandrle:
Apple TV HD runs tvOS 26. Which is surprising, given that it has a iPhone 6 chip. The fillrate is appalling but my simple sprite-based game manages to run OK.

I'm not sure about this patch. It works but it's messy. I wouldn't merge it if I were you, honestly, unless you see a way to simplify this?

### DIFF
--- a/sokol_gfx.h
+++ b/sokol_gfx.h
@@ -15971,6 +15971,13 @@ _SOKOL_PRIVATE void _sg_mtl_init_caps(void) {
     _sg.features.msaa_texture_bindings = true;
     _sg.features.draw_base_vertex = true;
     _sg.features.draw_base_instance = true;
+    if (@available(macOS 10.15, iOS 13.0, *)) {
+        // A8-class GPUs (Apple TV HD) are Apple2: no baseVertex/baseInstance draws
+        if ([_sg.mtl.device supportsFamily:MTLGPUFamilyApple1] && ![_sg.mtl.device supportsFamily:MTLGPUFamilyApple3]) {
+            _sg.features.draw_base_vertex = false;
+            _sg.features.draw_base_instance = false;
+        }
+    }
     _sg.features.dual_source_blending = true;
     _sg.features.vertexformat_int10_n2 = true;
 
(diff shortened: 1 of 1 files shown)


floooh:
Hmm I'm a bit confused about this one. Does tvOS not support the drawPrimitiveIndex version with baseVertex and baseInstance at all?

E.g. this is not allowed (baseVertex:0 and baseInstance:0)?

```objc
            [_sg.mtl.render_cmd_encoder drawIndexedPrimitives:pip->mtl.prim_type
                indexCount:(NSUInteger)num_elements
                indexType:pip->mtl.index_type
                indexBuffer:_sg_mtl_id(ib->mtl.buf[ib->cmn.active_slot])
                indexBufferOffset:index_buffer_offset
                instanceCount:(NSUInteger)num_instances
                baseVertex:0
                baseInstance:0];
```

E.g. the docs say it's supported since tvOS 9.0:
image

...or is there a performance penalty when calling the baseVertex/baseInstance version with baseVertex:0 and baseInstance:0?

tomasandrle:
Even if I call it with both base parameters set to 0, I get this:
```
MTLValidateFeatureSupport:8091: failed assertion `Base Vertex Instance Drawing is only supported on MTLGPUFamilyApple3 and later.'
```

tomasandrle:
One other way I saw was to split at _sg_draw vs _sg_draw_ex but that would propagate into all the other backends..

floooh:
Hmm that sucks :D 

I currently work on https://github.com/floooh/sokol/pull/1599, will come back to this PR afterwards.

tomasandrle:
Maybe it could be split into _sg_mtl_draw_ex and _sg_mtl_draw which would not accept the params and avoid the extra conditions inside...

floooh:
> Maybe it could be split into _sg_mtl_draw_ex and _sg_mtl_draw which would not accept the params and avoid the extra conditions inside...

My thought too... the common internal _sg_mtl_draw() only made sense when it really only was a single and simple code path. Can you change the PR like that in the meantime? But I can also simply do that during merge, either way is fine with me :)

tomasandrle:
Will do, and re-test on the device.

tomasandrle:
branch updated with *a* solution; another may be to have `_sg_mtl_draw` and `_sg_mtl_draw_ex` as mentioned and branch in `_sg_draw`, idk.

tested on iOS (iPhone 13), macOS (M2), tvOS with apple2-level A8, but only the new path; old path in `_sg_mtl_draw_ex` should be unchanged except for the added asserts.

It seems like there was never a device with tvOS 9 that would support mtl level apple3; Apple TV HD released with tvOS 9 but doesn't support it; the following Apple TV 4K (gen 1) released with tvOS 11 already...

floooh:
Ah, of course. I hadn't thought of that there's only a single internal `_sg_draw()` func. But that makes sense, because at some later time I probably want to turn `sg_draw_ex()` into the regular `sg_draw()`.

I'll first need to finish https://github.com/floooh/sokol/pull/1599, then I'll take care of your PR.

floooh:
Hmm, the code block which disables the base_vertex and base_instance features also triggers in the iOS simulator for an iPhone 17 Pro... investigating...

PS: looks like the capability querying via supportsFamily is generally useless on the simulator, since baseVertex/baseInstance works just fine in the simulator. Since this is such an esoteric edge case (tvOS support for older devices) I'll remove that feature detection block and pretend that baseVertex/baseInstance is always supported (and instead leave it to the Metal validation layer to complain), I think that's an ok compromise.

tomasandrle:
OK, it won't affect my usecase. Thanks!

Apple docs:

> Simulator provides a device instance with capabilities similar to an Apple family 2 GPU

"similar" indeed.

floooh:
Ok, merged with the feature-detection block removed and a couple minor code formatting changes, this is what ended up in master:

https://github.com/floooh/sokol/commit/477122b1b6662dbf0f1a36d3c797e8b41bff66cb

PR #1602, opened 2026-09-11
sokol_app.h: updated X11 mouse lock code to be similar to win32

xXAbieGamingXx:
Changed the X11 code for mouse lock to be more similar to win32 (request instead of on the spot). This change fixes a difference between X11 and win32 where locking the mouse during before the first frame works on win32, but fails on X11, while setting the mouse lock variable to true when it wasn't.

### DIFF
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3184,6 +3184,7 @@ typedef struct {
     Window root;
     Colormap colormap;
     Window window;
+    bool window_focused;
     Cursor hidden_cursor;
     Cursor standard_cursors[_SAPP_MOUSECURSOR_NUM];
     Cursor custom_cursors[_SAPP_MOUSECURSOR_NUM];
@@ -13413,15 +13414,15 @@ _SOKOL_PRIVATE int _sapp_x11_get_window_state(void) {
     return result;
 }
 
+_SOKOL_PRIVATE bool _sapp_x11_window_focused(void) {
+    return _sapp.x11.window_focused;
+}
+
 _SOKOL_PRIVATE void _sapp_x11_lock_mouse(bool lock) {
     _sapp.x11.requested_mouse_lock = lock;
 }
 
 _SOKOL_PRIVATE void _sapp_x11_do_lock_mouse(void) {
-    if (!_sapp_x11_window_visible()) {
-        return;
-    }
-
     int result = XGrabPointer(
         _sapp.x11.display,          // display
         _sapp.x11.window,           // grab_window
@@ -13449,10 +13450,6 @@ _SOKOL_PRIVATE void _sapp_x11_do_lock_mouse(void) {
 }
 
 _SOKOL_PRIVATE void _sapp_x11_do_unlock_mouse(void) {
-    if (!_sapp_x11_window_visible()) {
-        return;
-    }
-
     if (_sapp.x11.xi.available) {
         XIEventMask em;
         unsigned char mask[] = { 0 };
@@ -13471,6 +13468,15 @@ _SOKOL_PRIVATE void _sapp_x11_do_unlock_mouse(void) {
 }
 
 _SOKOL_PRIVATE void _sapp_x11_update_mouse_lock(void) {
+    // mouse lock can only be active when we're the active window
+    if (!_sapp_x11_window_focused()) {
+        // unlock mouse if currently locked
+        if (_sapp.mouse.locked) {
+            _sapp_x11_do_unlock_mouse();
+        }
+        return;
+    }
+
     // nothing to do if requested lock state matches current lock state
     const bool lock = _sapp.x11.requested_mouse_lock;
     if (lock == _sapp.mouse.locked) {
@@ -13769,17 +13775,15 @@ _SOKOL_PRIVATE void _sapp_x11_on_genericevent(XEvent* event) {
 _SOKOL_PRIVATE void _sapp_x11_on_focusin(XEvent* event) {
     // NOTE: ignoring NotifyGrab and NotifyUngrab is same behaviour as GLFW
     if ((event->xfocus.mode != NotifyGrab) && (event->xfocus.mode != NotifyUngrab)) {
+        _sapp.x11.window_focused = true;
         _sapp_x11_app_event(SAPP_EVENTTYPE_FOCUSED);
     }
 }
 
 _SOKOL_PRIVATE void _sapp_x11_on_focusout(XEvent* event) {
-    // if focus is lost for any reason, and we're in mouse locked mode, disable mouse lock
-    if (_sapp.mouse.locked) {
-        _sapp_x11_do_unlock_mouse();
-    }
     // NOTE: ignoring NotifyGrab and NotifyUngrab is same behaviour as GLFW
     if ((event->xfocus.mode != NotifyGrab) && (event->xfocus.mode != NotifyUngrab)) {
+        _sapp.x11.window_focused = false;
         _sapp_x11_app_event(SAPP_EVENTTYPE_UNFOCUSED);
     }
 }
--- a/sokol_app.h
+++ b/sokol_app.h
@@ -3189,6 +3189,7 @@ typedef struct {
     Cursor custom_cursors[_SAPP_MOUSECURSOR_NUM];
     int window_state;
     float dpi;
+    bool requested_mouse_lock;
     unsigned char error_code;
     Atom UTF8_STRING;
     Atom CLIPBOARD;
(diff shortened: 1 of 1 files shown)


sokol_app.h
```diff
@@ -3806,13 +3807,13 @@ _SOKOL_PRIVATE void _sapp_setup_default_icon(void) {
...
+        for (int ty = 0; ty < 8; ty++) {
             const uint32_t color = colors[ty];
-            for (int sy = 0; sy < scale; sy++, y++) {
+            for (int sy = 0; sy < scale; sy++) {
                 uint8_t bits = tile[ty];
-                for (int tx = 0, x = 0; tx < 8; tx++, bits<<=1) {
+                for (int tx = 0; tx < 8; tx++, bits<<=1) {
                     uint32_t pixel = (0 == (bits & 0x80)) ? blank : color;
-                    for (int sx = 0; sx < scale; sx++, x++) {
+                    for (int sx = 0; sx < scale; sx++) {
```
floooh: Can you undo this change since it's unrelated? I'll deal with that via a separate ticket: https://github.com/floooh/sokol/issues/1603

sokol_app.h
```diff
@@ -13453,6 +13413,78 @@ _SOKOL_PRIVATE int _sapp_x11_get_window_state(void) {
...
+        _sapp.x11.display,          // display
+        _sapp.x11.window,           // grab_window
+        True,                       // owner_events
+        ButtonPressMask | ButtonReleaseMask | PointerMotionMask,    // event_mask
+        GrabModeAsync,              // pointer_mode
+        GrabModeAsync,              // keyboard_mode
+        _sapp.x11.window,           // confine_to
+        _sapp.x11.hidden_cursor,    // cursor
+        CurrentTime);               // time
+    if(result == GrabSuccess) {
```
floooh: Nitpick: whitespace, change to `if (`

sokol_app.h
```diff
@@ -13453,6 +13413,78 @@ _SOKOL_PRIVATE int _sapp_x11_get_window_state(void) {
     return result;
 }
 
+_SOKOL_PRIVATE void _sapp_x11_lock_mouse(bool lock) {
+    _sapp.x11.requested_mouse_lock = lock;
+}
+
+_SOKOL_PRIVATE void _sapp_x11_do_lock_mouse(void) {
+    if(!_sapp_x11_window_visible()) {
```
floooh: Nitpick: whitespace, change to if (

sokol_app.h
```diff
@@ -13453,6 +13413,78 @@ _SOKOL_PRIVATE int _sapp_x11_get_window_state(void) {
...
+            em.mask_len = sizeof(mask);
+            em.mask = mask;
+            XISetMask(mask, XI_RawMotion);
+            XISelectEvents(_sapp.x11.display, _sapp.x11.root, &em, 1);
+         }
+    }
+}
+
+_SOKOL_PRIVATE void _sapp_x11_do_unlock_mouse(void) {
+    if(!_sapp_x11_window_visible()) {
```
floooh: Nitpick: whitespace, change to if (

floooh (changes requested):
Only some quick code style issues for now, will need to do a more thorough review later.

The tricky thing about touching the mouse lock code is basically to make sure the mouse can never get stuck in locked mode... that's why I want to take my time to do a proper review and test session.

sokol_app.h
```diff
@@ -13161,47 +13162,6 @@ _SOKOL_PRIVATE void _sapp_x11_update_cursor(sapp_mouse_cursor cursor, bool shown
...
-            unsigned char mask[] = { 0 };
-            em.deviceid = XIAllMasterDevices;
-            em.mask_len = sizeof(mask);
-            em.mask = mask;
-            XISelectEvents(_sapp.x11.display, _sapp.x11.root, &em, 1);
-        }
-        XWarpPointer(_sapp.x11.display, None, _sapp.x11.window, 0, 0, 0, 0, (int) _sapp.mouse.x, _sapp.mouse.y);
-        XUngrabPointer(_sapp.x11.display, CurrentTime);
-    }
-    XFlush(_sapp.x11.display);
```
floooh, line 13202: PS: this 'defensive flush' seems to be missing in your new code.

xXAbieGamingXx: I moved the mouse_lock_update call to be right before the XFlush call in the main loop to not have a repeat function call

floooh (approved):

floooh:
Thanks for the fixes. I'll look more closely at the PR over the weekend (next in queue is: https://github.com/floooh/sokol/pull/1601).

sokol_app.h
```diff
@@ -13453,6 +13413,79 @@ _SOKOL_PRIVATE int _sapp_x11_get_window_state(void) {
...
 }
 
+_SOKOL_PRIVATE void _sapp_x11_lock_mouse(bool lock) {
+    _sapp.x11.requested_mouse_lock = lock;
+}
+
+_SOKOL_PRIVATE void _sapp_x11_do_lock_mouse(void) {
+    if (!_sapp_x11_window_visible()) {
+        return;
+    }
```
floooh: This 'window visible' check both at the start of the _sapp_x11_do_lock_mouse() and _sapp_x11_do_unlock_mouse() functions don't look right here (they are not equivalents for the 'window is focused' check on Windows)

Was there a specific problem so that you had to add those checks?

sokol_app.h
```diff
@@ -13743,7 +13776,7 @@ _SOKOL_PRIVATE void _sapp_x11_on_focusin(XEvent* event) {
 _SOKOL_PRIVATE void _sapp_x11_on_focusout(XEvent* event) {
     // if focus is lost for any reason, and we're in mouse locked mode, disable mouse lock
     if (_sapp.mouse.locked) {
-        _sapp_x11_lock_mouse(false);
+        _sapp_x11_do_unlock_mouse();
```
floooh: I think this direct call in the focusout handler might now cause a potential problem, since `lock_requested` and `locked` are now separate flags: 

- `requested_mouse_lock` might be true while `_sapp.mouse.locked` is false.
- focusout event is triggered, window becomes unfocused, the mouse isn't locked yet (only the `requested_mouse_lock` flag is set, so `_sapp_x11_do_unlock_mouse()` isn't called
- next call to `_sapp_x11_update_mouse_lock()` would attempt to lock the mouse while the window is unfocused

IIRC I was running into exactly this issue in the Windows backend too, and the solution was to completely move from event-driven to polling instead of a mix of event-driven and polling.

E.g. for a similar solution we'd need:

- a new flag `bool window_focused` in `_sapp_x11_t`
- this flag would be set in `_sapp_x11_on_focusin` and unset in `_sapp_x11_on_focusout` (inside the if-block that calls the sokol-app event callback)
- `_sapp_x11_do_unlock_mouse()` is removed from `_sapp_x11_on_focusout`
- a new helper function `bool _sapp_x11_window_focused(void)` (just returns `_sapp.x11.window_focused`

And finally the `_sapp_x11_update_mouse_lock` should have the same 'unlock mouse when window is unfocused' like the Windows version, e.g. smth like this:

```c
_SOKOL_PRIVATE void _sapp_x11_update_mouse_lock(void) {
    // mouse lock can only be active when we're the active window
    if (!_sapp_x11_window_focused()) {
        // unlock mouse if currently locked
        if (_sapp.mouse.locked) {
            _sapp_x11_do_unlock_mouse();
        }
        return;
    }
    // ...rest of the function
}
```

floooh (changes requested):
A more thorough code review. IMHO we should more closely replicate the Windows behaviour.

floooh (approved):

floooh:
Ok thanks for the fixes! I'll take over from here.

floooh:
Btw, the new code is strictly better, but regarding:

> This change fixes a difference between X11 and win32 where locking the mouse during before the first frame works on win32, but fails on X11

I cannot reproduce this problem, e.g. in cgltf-sapp.c, when I'm calling `sapp_lock_mouse(true)` at the end of the init() function, the mouse is properly locked (and technically, the init() callback is "part" of the first frame.

Can you describe what you did and your setup (Linux distro, window manager, ...)? (I'm on Kubuntu 26.04 LTS with KDE Plasma version 6.6.6).

...in any case: going to merge your PR now...

floooh:
Lol, btw, I just noticed that the 'Disable touchpad while typing' KDE setting (GNOME probably has something similar) also affects sokol_app.h (no touchpad events are delivered when a key is pressed down)... nothing sokol_app.h can do about that, and hopefully people don't get the idea to play an FPS with the touchpad ;)

floooh:
Ok tested and merged (with some minor code cosmetic code cleanup changes: https://github.com/floooh/sokol/commit/e5959048803d616230112c6296ebf9197947a10f).

Also added a changelog blurb.

Thanks!

xXAbieGamingXx:
Regarding the problem during init(), I was able to reproduce it in events-sapp using a previous version of sokol_app.h (before these changes), and when `sapp_lock_mouse(true)` was called during init(), it failed to capture the mouse, but succeeded when called in frame(). I confirmed that calling `sapp_lock_mouse(true)` during init() succeeded after adding this change.

floooh:
Hmm still strange since that scenario (calling `sapp_lock_mouse(true)` in init() worked for me, also with the old code. What's your Linux setup? (distro and window manager)

xXAbieGamingXx:
I was using Sway for previous tests, and the same test works on KDE Plasma for me. If it is relevant, I am running Arch.

